PR #9, opened 2017-08-04
Make sure, that we always publish logs

rmohr:
Use JobDSL configure block, to configure the not natively supported
Publish-Over-Ssh build wrapper, which publishes independent of the job
status.


rmohr:
@stu-gott thx for pointing me to the errors.

PR #10, opened 2017-08-07
Add entries to gitignore

stu-gott:


.gitignore
```diff
@@ -6,3 +6,5 @@ hosts.*
 .vagrant
 *~
 *pyc
+venv
+github/ci/ci.retry
```
rmohr: could you make `github/ci/*.retry` out of it?

stu-gott: Thanks! Wasn't aware there might be other retry files.

PR #18, opened 2017-12-04
Prepare matrix build for release testing

rmohr:
Add matrix build support. The automation script can check the $TARGET environment variable to determine what it should deploy and test.


davidvossel:
so, is there a follow up patch for automation/test.sh?

rmohr:
yes, on the way.

rmohr:
Released and works.

PR #23, opened 2018-03-20
Prepare for docker-only usage

rmohr:
* Assume Centos 7.4 as slave base, now that we only need docker
* Make sure that nested virtualization is enabled
* Make sure that the docker store is located at /home/docker where most
  free space is available

### DIFF
--- a/github/ci/jenkins-master/tasks/main.yml
+++ b/github/ci/jenkins-master/tasks/main.yml
@@ -8,7 +8,7 @@
     state: present
     key: http://pkg.jenkins-ci.org/redhat/jenkins-ci.org.key
 - name: Install jenkins
-  dnf:
+  yum:
     name: '{{ item }}'
     state: latest
   with_items:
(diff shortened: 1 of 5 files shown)


github/ci/jenkins-master/tasks/main.yml
```diff
@@ -72,6 +72,11 @@
     port: 8080/tcp
     permanent: true
     state: enabled
+- firewalld:
+    zone: public
+    port: 37947/tcp
```
cynepco3hahue: What service does use this port?
And some cosmetic comment(not really need to implement)
you can use `with` statement
```
- firewalld:
     zone: public
     port: "{{ item }}"
     permanent: true
     state: enabled
  with_items:
    - 8080/tcp
    - 37947/tcp
```

rmohr: The swarm plugin started using it for rpc calls.

rmohr: Updated.

PR #24, opened 2018-03-21
Windows job

rmohr:
This includes various changes to make defining the windows jobs possible:

* Set master executor count to 0, and register the master node itself optionally also via swarm
* Make swarmRootDir configurable, to allow setting the swarm base directory per node
* Add a dedicated job for windows. This job will only be executed on a node which contains the "windows" label
* Allow attaching node labels via the swarm plugin by adding a labels variable to selected nodes in the inventory


rmohr:
> When the target will equal to windows, we will put label windows on job, does it mean that it will use the slaves only with windows label?

yes. Every pull request would trigger a windows job. It should then be sufficient to just run the windows tests with the release manifests there and skip all other tests.

PR #27, opened 2018-03-23
Allow the same job to run in parallel on a node

### LINKED ISSUE #26: Allow the same job to run in parallel on the same node
We had to limit the same job type to run only once on a node at the same time. This limitation is not true anymore. Remove it.

rmohr:
Fixes #26 


slintes:
Is removing the throttle completely a good idea? Maybe just increase the allowed number of builds per node? Or is there some other mechanism preventing overloading?

rmohr:
Oh we have build slots defined separately. This simply removes the limitation that one specific job can only run once on a node. Our cluster tooling was created in mind with allowing that.

PR #33, opened 2018-06-06
Add post build plugin

cynepco3hahue:
- remove dangling containers and volumes

Signed-off-by: Artyom Lukianov <alukiano@redhat.com>

### DIFF
--- a/github/ci/jenkins-master/templates/04-jobs.groovy
+++ b/github/ci/jenkins-master/templates/04-jobs.groovy
@@ -1,6 +1,6 @@
 {% for target in targets %}
 job('kubevirt-functional-tests-{{ target }}') {
-    {% if target == "windows" %}
+    {% if target == "windows2016" %}
        label('windows')
     {% endif %}
     throttleConcurrentBuilds {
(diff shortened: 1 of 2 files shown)


github/ci/jenkins-master/templates/04-jobs.groovy
```diff
@@ -110,6 +110,21 @@ cd go/src/kubevirt.io/kubevirt && bash automation/test.sh 2>&1 | tee ${WORKSPACE
...
+        postBuildScripts {
+            steps {
+                shell('''#!/bin/bash
+set +e
+
+if [ -n "$(docker ps -q -f status=exited)" ]; then
+	docker rm $(docker ps -q -f status=exited)
+fi
+
+if [ -n "$(docker volume ls -qf dangling=true)" ]; then
```
rmohr: We are deliberately keeping some volumes, although they will not be used. I am not sure if this would not also remove these volumes.

cynepco3hahue:
Currently will close it, maybe trap on additional signals will solve the volume problem.

PR #43, opened 2018-06-21
Add script for sharing secrets

rmohr:
@cynepco3hahue @slintes if you know of a better way let me know.


slintes:
maybe add a link to https://bjornjohansen.no/encrypt-file-using-ssh-key, or even a decrypt script? At least I would have to look it up everytime ;)

rmohr:
@slintes added a basic decryption script.

PR #50, opened 2018-07-25
Use k8s 1.10.4 and 1.11.0 targets

slintes:


slintes:
> just do not forget to run ansible playbooks after the merge

done ;)

PR #56, opened 2018-09-28
Enable life cycle plugin.

awels:


rmohr:
@awels I was a little bit over-eager with merging. Now the cdi repo will react to the comments, but it will not periodically scrape the  repo and add the comments which will add the labels. You will have to change this file too: https://github.com/kubevirt/project-infra/blob/master/github/ci/prow/files/job-config.yaml

awels:
Okay, let me create another PR to include that.

PR #57, opened 2018-09-28
Update lifecycle job config to include CDI repo.

awels:


awels:
@rmohr I couldn't find any documentation on what arguments I can pass, but I saw multiple --labels, so I figured I can add a --query for another repo. If that doesn't work can you point me to some documentation?

rmohr:
@awels #58 is now merged. LIfecycle plugins should now also work for cdi.

/close

PR #58, opened 2018-10-02
Enable prow for the whole kubevirt org

rmohr:
This PR enables prow for all repos. Once we add an org-wide webhook prow will get operate on the whole org. As part of that it would also synchronize labels on all repos.

Labels which need to exist on all repos and labels which are specific to repos can be configured in `labels.yaml`. Plugins for all or specific repos can be configured in `plugins.yaml`.

If we really want to enable this, we should go through our existing repos and make sure that the labels they have are synchronized.


rmohr:
@fabiand let me know when you want to proceed with this.

PR #48, opened 2018-07-06
[WIP] Monitoring

rmohr:
Add the following things:
 - [x] Prometheus
 - [x] Prometheus Pushgateway
 - [x] Grafana
 - [x] Automatically add Prometheus as Grafana data source
 - [ ] Alertmanager
 - [x] Node Exporter to monitor CI slaves with prometheus
 - [x] Collect node exporter metrics with Prometheus
 - [x] Monitor Prow with the prometheus instance
 - [ ] Automatically add a reasonable dashboard to grafana
 - [ ] configure disk space alerts for alertmanager
 - [x] Secure with tls in combination with basic auth
 - [ ] Grafana config contains "secret" values. Either make it a secret or split it somehow.

PR #64, opened 2018-12-11
Skip new archived repos in label-sync

rmohr:
The label sync binary does not like archived repos. Use a forked build until a rebuild with kubernetes/test-infra#10407 included happens.

### DIFF
--- a/github/ci/prow/templates/label-sync.yaml
+++ b/github/ci/prow/templates/label-sync.yaml
@@ -29,12 +29,11 @@ spec:
         spec:
           containers:
             - name: label-sync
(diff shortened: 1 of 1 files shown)


rmohr:
Created a PR https://github.com/kubernetes/test-infra/pull/10407, which should  fix that permanently for us.

rmohr:
> just one question: how did you know that there is a problem?

I check the Jobs sometimes. We should finally continue with what was started in #48. It should now be much simpler because we only use the bots for now and don't have to monitor nodes and jenkins now.

slintes:
>  kubernetes/test-infra#10407 is merged

nice :)

PR #68, opened 2019-02-14
Add provisioner jobs for kubevirt/kubevirtci

rmohr:


cynepco3hahue:
We need to make people aware that if they add new provider they also need to add new job under this repo.

github/ci/prow/files/job-config.yaml
```diff
@@ -112,3 +148,268 @@ presubmits:
         resources:
           requests:
             memory: "500Mi"
+
+  kubevirt/kubevirtci:
+  - name: check-provision-os-3.11.0-multus
```
gbenhaim: Can we use yaml inheritance in order to make the config shorter?

github/ci/prow/files/job-config.yaml
```diff
@@ -112,3 +148,268 @@ presubmits:
...
+      affinity:
+        nodeAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            nodeSelectorTerms:
+            - matchExpressions:
+              - key: type
+                operator: In
+                values:
+                - vm
+              - key: kubernetes.io/hostname
```
gbenhaim: Why is this selector is different from the others?

github/ci/prow/files/job-config.yaml
rmohr: We can use one or multiple sub directories. Prow can collect all job configs there and combine them. I can create a PR after this one which rearranges that.

github/ci/prow/files/job-config.yaml
rmohr: Evgheni enable nested virt on that node for me, to test if the privisioning of the clusters, which does not involve starting nested vms, could be done in vms instead of bare-metal. For k8s clusters it looks like an option (have seen 30 minutes instead of 10 minutes), for openshift I think it is too slow (more than two hours compared to 50 minutes).

rmohr:
@cynepco3hahue @danielBelenky I think this should be save to merge. All tests are only triggered when explicitly asked and the placement should be good too.

cynepco3hahue:
@danielBelenky Can you please review?

PR #75, opened 2019-02-20
Pushgateway

rmohr:
Add the pushgateway and configure all prow components to push metrics to it. Greenhouse is the only component which wants to be scraped (under `/prometheus`) and does not want to push metrics.

This PR does not include prometheus itself or a graphana dashboard. It only prepares the ground.

### DIFF
--- a/github/ci/prow/files/config.yaml
+++ b/github/ci/prow/files/config.yaml
@@ -58,6 +58,9 @@ tide:
     # Treat unknown contexts as optional
     skip-unknown-contexts: true
 
+push_gateway:
+  endpoint: pushgateway
+
 presets:
 - labels:
     preset-dind-enabled: "true"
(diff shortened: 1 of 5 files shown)


rmohr:
Should be ready for merge. I see the ghproxy metrics in the pushproxy and greenhouse metrics are also accessible via the service.

cynepco3hahue:
@rmohr Can I see these metrics without entering the cluster?

rmohr:
Not yet. If you want to check if they work right now you  can do something like this:

```
oc port-forward -n kubevirt-prow svc/bazel-cache-metrics 8080:80 &
curl localhost:8080/prometheus
```

and

```
oc port-forward -n kubevirt-prow svc/pushgateway 8080:80 &
curl localhost:8080/metrics

```

PR #76, opened 2019-02-20
Add Prometheus and Grafana

rmohr:
This PR deploys
 * prometheus
 * grafana
 * github-cache and bazel-cache dashboard
 * token-counter (to monitor the github token usage)
 * bind all infra pods to vms, to avoid scheduling them on the bare-metal machines.

The entrypoint is https://grafana-kubevirt-prow.apps.ovirt.org.


github/ci/prometheus/templates/prometheus.yaml
```diff
@@ -0,0 +1,102 @@
...
+            - name: prometheus-data
+              mountPath: /data
+              subPath: ""
+      terminationGracePeriodSeconds: 300
+      volumes:
+        - name: config-volume
+          configMap:
+            name: prometheus-config
+        - name: prometheus-data
+          emptyDir: {} # todo, PVC template
```
rmohr: @jekader for prometheus I will then need another (small) PV. Will follow up on this. Right now we will loose all metrics if prometheus goes down.

rmohr:
@cynepco3hahue that PR should be ready. it still misses persistent storage for prometheus, but that can be done as a follow-up.

github/ci/prometheus/templates/token-counter.yaml
```diff
@@ -0,0 +1,44 @@
...
+    spec:
+      nodeSelector:
+        type: vm
+      containers:
+      - name: token-counter
+        args:
+        - --stderrthreshold=0
+        - --prom-only
+        - --token=/github-tokens/oauth
+        image: kubevirtci/token-counter:v20190222-092216
```
rmohr: We can use the prow token-counter again once https://github.com/kubernetes/test-infra/pull/11447 is merged.

PR #78, opened 2019-02-26
Add presubmit job for kubevirtci which can push to docker hub

rmohr:
Add a job which can push and build k8s-genie-1.11.1.


rmohr:
/hold

test

PR #81, opened 2019-03-12
Add presubmit jobs for the provider k8s-1.13.3

cynepco3hahue:
I think we should provide some test script under kubevirtci repository.

### DIFF
--- a/github/ci/prow/files/jobs/kubevirtci/kubevirtci-presubmits.yaml
+++ b/github/ci/prow/files/jobs/kubevirtci/kubevirtci-presubmits.yaml
@@ -130,6 +130,32 @@ presubmits:
           requests:
             memory: "8Gi"
 
+  - name: check-provision-k8s-1.13.3
+    always_run: false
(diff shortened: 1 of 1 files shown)


github/ci/prow/files/jobs/kubevirtci/kubevirtci-presubmits.yaml
```diff
@@ -130,6 +130,32 @@ presubmits:
           requests:
             memory: "8Gi"
 
+  - name: check-provision-k8s-1.13.3
+    always_run: false
+    optional: true
+    decorate: true
+    max_concurrency: 1
+    spec:
```
cynepco3hahue: @rmohr Say do we miss indentation? I can see double `spec` on the same indentation level.

rmohr: Yes you are right. The `spec:` line itself should be removed.

PR #82, opened 2019-03-14
Add memory limits to Prow deployment

danielBelenky:
To ensure that different components are not being evicted at runtime, we
should have memory requests and limits set on every deployed container.
Currently, added default req and lim for all the containers and set it
to 510Mi. For components that are expected to consume more memory
(currently greenhouse and deck), we overwrite the defaults.
Also, added pod anti-affinity rules for components with more than one
replica.

### DIFF
--- a/github/ci/prow/tasks/main.yml
+++ b/github/ci/prow/tasks/main.yml
@@ -89,6 +89,10 @@
       type: Opaque
       data:
         service-account.json: "{{ gcsServiceAccount | b64encode }}"
+- name: Deploy limit range
+  command: "oc -n {{prowNamespace}} apply -f -"
+  args:
+    stdin: "{{ lookup('file', '{{ role_path }}/templates/memory-defaults.yaml') }}"
 - name: Deploy the pushgateway
   command: "oc -n {{prowNamespace}} apply -f -"
   args:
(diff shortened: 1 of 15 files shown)


danielBelenky:
@rmohr hey, how do you suggest to verify this PR? I'm not sure that re-deploying to production is the right way to go :thinking:

rmohr:
I think that we need to decouple the configs and put them into a separate task. Then you can simply deploy prow in another namespace and see if all services come up. We can also add a base config for sanity checking.

danielBelenky:
> I think that we need to decouple the configs and put them into a separate task.

Sorry, which configs are we talking about? job configs?

rmohr:
> Sorry, which configs are we talking about? job configs?

Mostly about `config.yaml` and `plugins.yaml`. The jobs are only updated by the PR updater already. The other files are also updated on merges, but also via the ansible playbook (which should still be possible for those).

rmohr:
Here the files: https://github.com/kubevirt/project-infra/tree/master/github/ci/prow/files

danielBelenky:
Ok. Do you want this task to be a blocker to merge this PR? Since setting a CI flow for the deployment might take some time and we probably will want to plan it.

rmohr:
Just try it out. You can easily role back. We don't use it in kubevirt/kubevirt yet, so pretty risk free.

PR #84, opened 2019-03-21
Separate namespace

rmohr:
Move Jobs and Pods into a separate namespace, so that jobs are isolated from the prow components and that we can have for instance distinct memory limits.

### DIFF
--- a/github/ci/prow/files/config.yaml
+++ b/github/ci/prow/files/config.yaml
@@ -29,8 +29,8 @@ sinker:
   max_prowjob_age: 48h
   max_pod_age: 30m
 
-prowjob_namespace: kubevirt-prow
-pod_namespace: kubevirt-prow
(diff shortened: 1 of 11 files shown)


github/ci/prow/files/config.yaml
```diff
@@ -29,8 +29,8 @@ sinker:
   max_prowjob_age: 48h
   max_pod_age: 30m
 
-prowjob_namespace: kubevirt-prow
-pod_namespace: kubevirt-prow
+prowjob_namespace: kubevirt-prow-jobs
+pod_namespace: kubevirt-prow-jobs
```
slintes: wasn't the idea to use different namespaces? :thinking:

rmohr: Haha, for now different than the prow components.

slintes: got it ;) :+1:

PR #87, opened 2019-04-11
Add basic gocli CI jobs

rmohr:


rmohr:
Hm, the config updater has an issue. Something with missing name of the config map.

PR #89, opened 2019-04-18
Add demo jobs for kubevirt

rmohr:
* Add a docker mirror which can be referenced via `preset-docker-mirror: "true"`
 * Add a preset which mounts the shared images directory from stdci via `preset-shared-images: "true"`
 * Add e2e kubevirt jobs for all jobs which exist in stdci

### DIFF
--- a/github/ci/prow/files/docker-daemon-mirror.conf
+++ b/github/ci/prow/files/docker-daemon-mirror.conf
@@ -1,0 +1,2 @@
+DOCKER_OPTS="${DOCKER_OPTS} --data-root=/docker-graph"
+DOCKER_OPTS="${DOCKER_OPTS} --registry-mirror=http://docker-mirror.kubevirt-prow.svc:5000 --insecure-registry=http://docker-mirror.kubevirt-prow.svc:5000"
--- a/github/ci/prow/files/jobs/kubevirt/kubevirt-presubmits.yaml
+++ b/github/ci/prow/files/jobs/kubevirt/kubevirt-presubmits.yaml
@@ -8,6 +8,7 @@ presubmits:
     max_concurrency: 8
     labels:
       preset-dind-enabled: "true"
+      preset-docker-mirror: "true"
     spec:
       nodeSelector:
         model: r430
@@ -18,7 +19,7 @@ presubmits:
         - "/usr/local/bin/runner.sh"
         - "/bin/sh"
         - "-c"
-        - "export TARGET=k8s-multus-1.13.3 && automation/test.sh"
+        - "cat /etc/default/docker && ps aux | grep docker && export TARGET=k8s-multus-1.13.3 && automation/test.sh"
         # docker-in-docker needs privileged mode
         securityContext:
           privileged: true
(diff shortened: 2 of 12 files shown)


github/ci/prow/files/jobs/kubevirt/kubevirt-presubmits.yaml
```diff
@@ -0,0 +1,234 @@
+presubmits:
+  kubevirt/kubevirt:
+  - name: pull-kubevirt-e2e-k8s-1.10.11
```
cynepco3hahue: can we remove pull world, it just lacks sense her

github/ci/prow/files/docker-daemon-mirror.conf
```diff
@@ -0,0 +1,2 @@
+DOCKER_OPTS="${DOCKER_OPTS} --data-root=/docker-graph"
+DOCKER_OPTS="${DOCKER_OPTS} --registry-mirror=http://docker-mirror.kubevirt-prow.svc:5000 --insecure-registry=http://docker-mirror.kubevirt-prow.svc:5000"
```
cynepco3hahue: will the job use this registry instead of our internal `registry:5000`?

github/ci/prow/files/jobs/kubevirt/kubevirt-presubmits.yaml
rmohr: I took the convention from kubernetes. You will see postsubmit results also on the PR. It helps to differentiate them.

github/ci/prow/files/docker-daemon-mirror.conf
rmohr: That is just for docker itself. It pulls all kubevirtci images from the mirror in the datacenter instead of going to the internet. It should not interfere with bazel or our internal registry.

github/ci/prow/files/jobs/kubevirt/kubevirt-presubmits.yaml
```diff
@@ -0,0 +1,234 @@
+presubmits:
```
danielBelenky: Since most of the configurations here are the same except for `name` and `image`, maybe YAML anchors could make it more readable? WDYT?

github/ci/prow/templates/mirror-config.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v1
```
danielBelenky: Personally, I'd prefer to split the addition of the mirror service from the addition of the new prow jobs into different commits/PRs - but not that important in this stage.

github/ci/prow/files/jobs/kubevirt/kubevirt-presubmits.yaml
rmohr: Hm, it would always only save a minimal amount. To be honest I would right now prefer to simply see everything. We already use presets via the labels to avoid re-typing infra-related things. The rest is very similar but savings would be very similar and I find it this way more readable.

github/ci/prow/templates/mirror-config.yaml
rmohr: It's the infra for optional new jobs, so I don't see much of difference and it kind of belongs together. I agree that there are situations where it is better to split that into separte PRs. The mirror is in a separate commit.

rmohr:
@danielBelenky oh and I forgot that we do squashing here on merge. So now everything is one commit. Will split them next time as you suggested.

PR #92, opened 2019-04-21
Always run kubevirt e2e with prow instead of stdci

rmohr:
/hold

Can only be merged after the following preconditions:
 * #91 needs to be merged
 * https://github.com/kubevirt/kubevirt/pull/2199 needs to be merged
 * The stdci webhook needs to be disabled for kubevirt/kubevirt

In case something goes wrong, the tests can be disabled again and the stdci webhook re-enabled.
This does not enable `tide`. It only let's the prow jobs run.

### DIFF
--- a/github/ci/prow/files/jobs/kubevirt/kubevirt-presubmits.yaml
+++ b/github/ci/prow/files/jobs/kubevirt/kubevirt-presubmits.yaml
@@ -1,42 +1,13 @@
 presubmits:
   kubevirt/kubevirt:
-  - name: pull-kubevirt-e2e-k8s-1.10.11
-    always_run: false
-    optional: true
-    decorate: true
-    decoration_config:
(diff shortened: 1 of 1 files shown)


github/ci/prow/files/jobs/kubevirt/kubevirt-presubmits.yaml
```diff
@@ -1,13 +1,13 @@
...
   - name: pull-kubevirt-e2e-k8s-1.10.11
-    always_run: false
+    always_run: true
     optional: true
     decorate: true
     decoration_config:
       timeout: 14400000000000 # 4h
       grace_period: 500000000000 # 5m
-    max_concurrency: 8
+    max_concurrency: 5
```
rmohr: @danielBelenky does that sound save enough for a slow start?

PR #93, opened 2019-04-24
updated kubevirtci prow jobs

slintes:
- Reordered jobs to have check-provision and push jobs together for each cluster
- Added push jobs for os-3.11.0 / -crio / -multus


rmohr:
/lgtm

@slintes I opened #94 which explains on how you can also test your new jobs without having to merge them first. Give it a try if you like.

PR #94, opened 2019-04-25
Document how to create jobs from local configurations

rmohr:


slintes:
Didn't work, maybe the job manifest created by the tool isn't compatible with our prow version. Needs some investigation.
/hold

cynepco3hahue:
@slintes Can you show the error that you got?

cynepco3hahue:
I checked it locally and all worked fine, succeeded to create the prow job.
/lgtm

slintes:
@cynepco3hahue I was able to create jobs, but they weren't executed. Did they for you?

cynepco3hahue:
@slintes yes it worked for me and job started.

rmohr:
Hm, @cynepco3hahue could you add the tag of the mkpj which you used? I guess that in the version which @slintes used a bug might exist, or it is not compatible with our prow version.

cynepco3hahue:
@rmohr I used master one with the commit `ab25880be748bb3d20cfdcba86e4974fec36dbe6`

rmohr:
@slintes @cynepco3hahue this PR should work now too. The reason why it did not work before was that we were on an old prow version which did not support setting things like `5m` for grace periods. Works now with #104 merged.

PR #95, opened 2019-04-29
Give plank "get" access to prowjobs

rmohr:
"get" was missing and led to situations where plank can not properly
update prowjobs.


rmohr:
/approve

Works, removed the rbac issues.

PR #97, opened 2019-04-30
Run Prow services on the CI zone

gbenhaim:
The infra zone consist of VMs that can migrate, thus we use
it for running Jenkins instances in order to achieve high availability
(we can't run replicas of Jenkins because they can't share the same
storage).

We prefer not to have other services on the nodes that runs the Jenkins
instances, in order to make it more stable.

Signed-off-by: gbenhaim <galbh2@gmail.com>

### DIFF
--- a/github/ci/prometheus/templates/grafana.yaml
+++ b/github/ci/prometheus/templates/grafana.yaml
@@ -23,12 +23,13 @@ spec:
     spec:
       nodeSelector:
         type: vm
+        zone: ci
       containers:
         - name: grafana
           image: "grafana/grafana:5.2.1"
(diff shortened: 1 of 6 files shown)


gbenhaim:
@rmohr This should solve the issue where Jenkins fails to start threads. Can you please take a look?

gbenhaim:
@rmohr Do I need to run anything in order to apply those changes?

rmohr:
Yes, ansible needs to be run. On it.

rmohr:
Done.

gbenhaim:
Thanks

PR #108, opened 2019-06-03
Set MTU to 1450 for DIND

rmohr:
There seem to be pip issues regarding to refragment requests.


eedri:
Can you run a test on it on the two servers we added?
Packet (m1.xlargex86) - 139.178.69.73 
Hetzner (DELL R640 DX292) - 116.202.72.162 

LMK if you need root access to them.

PR #114, opened 2019-06-12
Removed rc0 postfix of okd4 job

slintes:
Renamed okd-4.1.0-rc.0 to okd-4.1.0, which aligns it with what we have in kubevirtci in the meanwhile

### DIFF
--- a/github/ci/prow/files/jobs/kubevirt/kubevirt-presubmits.yaml
+++ b/github/ci/prow/files/jobs/kubevirt/kubevirt-presubmits.yaml
@@ -203,7 +203,7 @@ presubmits:
         resources:
           requests:
             memory: "24Gi"
-  - name: pull-kubevirt-e2e-okd-4.1.0-rc.0
+  - name: pull-kubevirt-e2e-okd-4.1.0
     always_run: false
     optional: true
     decorate: true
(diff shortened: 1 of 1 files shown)


github/ci/prow/files/jobs/kubevirt/kubevirt-presubmits.yaml
```diff
@@ -203,7 +203,7 @@ presubmits:
         resources:
           requests:
             memory: "24Gi"
-  - name: pull-kubevirt-e2e-okd-4.1.0-rc.0
+  - name: pull-kubevirt-e2e-okd-4.1.0
     always_run: false
```
rmohr: Do we already want to enable it by default? If yes, `always_run` should be set to `true`.

cynepco3hahue: I think first we need to run all tests locally to see how many of the succeeds, otherwise the lane will always fail on the timeout(too many failures)
I started to run it locally but I think it will take me some time, so I do not want to delay this PR.

slintes: I did run it locally (well, on another remote server) and it it failed with the masquerade / iptables problem I posted in the chat. I created this PR with focused tests in order to check if that is a problem on my server: https://github.com/kubevirt/kubevirt/pull/2379

PR #118, opened 2019-06-18
Move kubevirt/kubevirt to new nodes

rmohr:
* migrate greenhouse and docker mirror to new nodes
 * trigger kubevirt/kubevirt jobs on new nodes only

/hold

### DIFF
--- a/github/ci/prow/files/jobs/kubevirtci/kubevirtci-presubmits.yaml
+++ b/github/ci/prow/files/jobs/kubevirtci/kubevirtci-presubmits.yaml
@@ -9,8 +9,7 @@ presubmits:
       preset-dind-enabled: "true"
     spec:
       nodeSelector:
-        model: r430
-        type: bare-metal
+        type: bare-metal-external
       containers:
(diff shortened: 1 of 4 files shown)


rmohr:
Because we don't want cross-zone cache traffic, the rollout order is like this:
 * temporary disable prow hook for kubevirt/kubevirt
 * stop all running jobs
 * move the cache by running ansible
 * merge this PR to update jobs
 * enable hook again

github/ci/prow/files/jobs/kubevirtci/kubevirtci-presubmits.yaml
```diff
@@ -548,7 +548,7 @@ presubmits:
     spec:
       nodeSelector:
         model: r430
```
eedri: should the model change as well? since its not r430 anymore

rmohr: Yes, good catch. Thanks.

rmohr: Fixed.

PR #119, opened 2019-06-18
Set right docker MTU also for kubevirtci jobs

rmohr:
Packet nodes use a MTU of 1450. Our DIND setup defaults to 1500 in the kubevirtci jobs. Inject the correct docker config.


rmohr:
@awels @mfranczy this should resolve the strange issues which started today on the kubevirtci jobs. Related to our switch to the new servers.

rmohr:
/cc @davidvossel 

can you `/lgtm` the PR?

PR #122, opened 2019-06-19
Feat: Added plugins on cloud-image-builder config

jparrill:
- Allowed automerge by squash method
- Added some plugins in order to start testing soon:
  - config-updater
  - skip
  - override
  - trigger

### DIFF
--- a/github/ci/prow/files/plugins.yaml
+++ b/github/ci/prow/files/plugins.yaml
@@ -22,6 +22,8 @@ plugins:
   - blunderbuss
   - lifecycle
   - verify-owners
+  - skip
+  - override
   - wip
   - pony
   - dog
@@ -57,7 +59,6 @@ plugins:
   - approve
   - lgtm
   - owners-label
-  - config-updater
   - skip
   - override
   - trigger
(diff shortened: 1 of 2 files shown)


github/ci/prow/files/plugins.yaml
```diff
@@ -66,6 +66,10 @@ plugins:
   - approve
   - lgtm
   - owners-label
+  - config-updater
```
rmohr: The config-update is enabled for the project-infra repo. It contains all configurations. You don't have to add this.

github/ci/prow/files/plugins.yaml
```diff
@@ -66,6 +66,10 @@ plugins:
...
+  - skip
```
rmohr: I think it makes sense to enable `skip` and `override` for the whole org (the `kubevirt:` section directly below `plugins:`.

jparrill: Make sense, yes. Almost all will end needing it. Adding them to the org section

rmohr:
@jparrill the validator wants that we remove `override` and `skip` for all subrepositories if we enable it for the whole org.

jparrill:
> @jparrill the validator wants that we remove `override` and `skip` for all subrepositories if we enable it for the whole org.

PR #124, opened 2019-06-24
Double allowed concurrent builds for all kubevirt prow jobs

rmohr:
The long running build lanes have a big backlog while we have resources
free.


rmohr:
We can probably reduce the memory requests and allow more in parallel, but let's increase slowly and monitor.

PR #129, opened 2019-07-03
Add automatic retest capability for kubevirt/kubevirt

rmohr:
Trigger /retest on one PR per hour at most which is approved.  This does
not enable tide. It should only help maintainers to not have to go back
to PRs and type retest there.


rmohr:
This PR should allow us to use `lgtm` and `approve` to trigger automatic retests. It will not enable prow and not merge anything automatically but might still help already a little.

rmohr:
@fabiand FYI

fabiand:
IOW This PR will pick an approved PR and trigger retest - and in general at max 1 PR per hour?

rmohr:
Yep. Should hopefully help to merge prs faster, since it eliminates the early morning retest routine.

slintes:
great idea!

/lgtm

PR #142, opened 2019-07-11
[wip] simple flakefinder implementation

rmohr:
flakefinder does the followingg:

 1. Fetching all merged PRs within a given time period
 2. Getting the last commit of each PR which got merged
 3. Correlate the PR with all prowjobs which were running against this last commit
 4. Extact skipped/failed/success from the junit results and create a html table

Selecting the right builds:

 * Filter out not merged PRs
 * Filter out identical prow jobs on multiple PRs (can be because of the merge pool)
 * Filter out jobs which don't have a junit result
 * Only shows test results for all lanes where a test at least failed once on one of the found lanes
 * Only take prow jobs into account which were run on the commit which got merged


rmohr:
flakefinder

rmohr:
joblist

robots/flakefinder/downloader.go
```diff
@@ -0,0 +1,197 @@
...
+	JUnit       []junit.Suite
+	BuildNumber int
+	PR int
+}
+
+func isLatestCommit(jsonText []byte, pr *github.PullRequest) bool {
+	var status startedStatus
+	err := json.Unmarshal(jsonText, &status)
+	for _, v := range status.Repos {
+		return err == nil && strings.Contains(v, fmt.Sprintf("%d:%s", pr.Number, pr.Head.SHA))
```
dhiller: A nit here:

IIUC you're leveraging that if ranging over an empty slice the `for` block will not get executed. if it is, the `err` and first element are checked. I find this puzzling and harder to read.

Is this a normal idiom of go? If not I'd rather use `if len(status.Repos) == 0` to guard against an empty slice here. Then afterwards I'd check the first element. This would make it more obvious what exactly is happening here.

robots/flakefinder/main.go
```diff
@@ -0,0 +1,109 @@
...
+type options struct {
+	ceiling         int
+	endpoint        flagutil.Strings
+	token           string
+	graphqlEndpoint string
+	merged          time.Duration
+}
+
+type client interface {
+	FindIssues(query, sort string, asc bool) ([]github.Issue, error)
```
dhiller: Formatting looks odd here.

robots/flakefinder/main.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+}
+
+func makeQuery(query string, minMerged time.Duration) (string, error) {
+	parts := []string{query}
+	if minMerged != 0 {
+		latest := time.Now().Add(-minMerged)
+		parts = append(parts, "merged:>="+latest.Format(time.RFC3339))
+	}
+	return strings.Join(parts, " "), nil
```
dhiller: There is no other `return` in the function, so we can remove the `error` returned here.

robots/flakefinder/downloader.go
rmohr: I should have iterated over all values and not just the first one, so this is a mistake on my side (would only have noticed it when working on jobs from the merge pool). In general iterating over `nil` arrays is normal in go.

davidvossel:
I didn't dive deep into the code or anything, but this looks pretty helpful. 

Whats time period this flake finder scans over and where do we go to see these results once this gets merged?

rmohr:
> Whats time period this flake finder scans over and where do we go to see these results once this gets merged?

I think that at the end we want different periods. The report shared recently covered 1000 hours. I think we want in general: sum of 4 weeks, sum of last week and in addition e.g. the last 3 days individually.

The plan is that we run it in a periodic prow jobs and that it uploads the results to GCS. @dhiller will work on that. We can then have a well-known http entry-page which the robot updates and which collects the urls to all reports.

slintes:
superseded by https://github.com/kubevirt/project-infra/pull/146

/close

PR #154, opened 2019-08-09
Use same container image for all Virt Blocks jobs

andreabolognani:
While technically the `pull-virtblocks-vet` and `pull-virtblocks-fmt` jobs don't need `libtool` to run, it makes sense that all jobs for the projects would use the same container image.


rmohr:
/approve
/lgtm

@andreabolognani the changes will be live within less than a minute, normally.

PR #172, opened 2019-08-26
Virt Blocks update

andreabolognani:
Build the Rust code along with the Go code; in addition to building, run the test suite and the example programs too.

Doing this requires using a custom container image that includes a Rust compiler.

### DIFF
--- a/github/ci/prow/files/jobs/virtblocks/virtblocks/virtblocks-periodics.yaml
+++ b/github/ci/prow/files/jobs/virtblocks/virtblocks/virtblocks-periodics.yaml
@@ -5,7 +5,7 @@ presubmits:
     always_run: true
     spec:
       containers:
-        - image: kubevirt/golang-libtool:1.12.5-r1
+        - image: quay.io/virtblocks/buildenv:1.0.0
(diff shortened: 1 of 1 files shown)


github/ci/prow/files/jobs/virtblocks/virtblocks/virtblocks-periodics.yaml
```diff
@@ -5,13 +5,39 @@ presubmits:
     always_run: true
     spec:
       containers:
-        - image: kubevirt/golang-libtool:1.12.5-r1
+        - image: quay.io/virblocks/buildenv:1.0.0
```
mfranczy: should not the name be `virtblocks`?

andreabolognani: > should not the name be `virtblocks`?

It definitely should, good catch! I'll fix it right away.

andreabolognani: Done!

PR #173, opened 2019-08-26
Virt Blocks update

andreabolognani:
Sorry for the noise :(

Unfortunately permission issues in the 1.0.0 container image caused builds to fail. I believe I've addressed the underlying problem in the 1.0.1 image.

Thanks for your patience.


andreabolognani:
@nertpinx pointed out that I'm listed in the [OWNERS file for Virt Blocks](https://github.com/kubevirt/project-infra/blob/master/github/ci/prow/files/jobs/virtblocks/virtblocks/OWNERS): does that mean I could `/approve` the PR myself? And if so, would it be considered appropriate, at least for a straightforward update such as this one? Thanks!

slintes:
I think you can approve but not lgtm your own PR :wink: 

/approve
/lgtm

andreabolognani:
> I think you can approve but not lgtm your own PR

I see! Thanks for merging, it seems to work now :)

PR #182, opened 2019-09-03
Added build-change label

slintes:
Will be used via OWNER file in kubevirt/kubevirt repo in order to raise awareness for build changes


slintes:
will just be a marker

PR #184, opened 2019-09-04
Added k8s 1.15 with ceph for release-0.20 branch, disabled k8s 1.11

slintes:


slintes:
> Just curious: Why `29Gi`?

good question. IIRC a high number for not letting too many jobs run in parallel on the CI nodes? @rmohr probably can tell more when he is back.

PR #190, opened 2019-09-13
Enable dco plugin for kubevirt and kubevirtci

rmohr:


rmohr:
@dhiller an alternative to the external DCO check.

dhiller:
^^ @rmohr

rmohr:
@dhiller You can enable it for the whole org by adding it to the plugins which are enabled for the whole org.

I would like to try it out first, I have never seen it in action.

dhiller:
> I would like to try it out first, I have never seen it in action.

@rmohr OK, then let's try with your PR and see how it goes.

PR #205, opened 2019-09-25
Add e2e job for virtblocks

rmohr:
Add an e2e job which runs on machines with /dev/kvm and update to the new builder image which contains golang and qemu.


rmohr:
@andreabolognani should be ready. If you are satisfied you can merge the PR with

```
/lgtm
/approve
```

PR #207, opened 2019-09-25
Enable DCO plugin for virtblocks org

rmohr:
Enable the DCO plugin for the virtblocks org.

### DIFF
--- a/github/ci/prow/files/plugins.yaml
+++ b/github/ci/prow/files/plugins.yaml
@@ -17,6 +17,7 @@ plugins:
 
   virtblocks:
   - trigger
+  - dco
 
   kubevirt:
   - size
@@ -105,6 +106,7 @@ triggers:
   trusted_org: kubevirt
 - repos:
   - virtblocks/virtblocks
+
 
 approve:
 - repos:


rmohr:
@andreabolognani let me know if you want this enabled. Just respond here and we can merge it.

rmohr:
/approve

/hold

holding until @andreabolognani agreed to this.

andreabolognani:
> @andreabolognani let me know if you want this enabled. Just respond here and we can merge it.

So IIUC this will cause the KubeVirt CI to verify `Signed-off-by` tags are present for each commit as required, and if so add the `dco-signoff: yes` label.

Will the submitter be notified automatically if the tags are not present, and told to add them?

Either way this definitely sounds like a good idea, so OK to merge from my side. Thanks!

rmohr:
> So IIUC this will cause the KubeVirt CI to verify `Signed-off-by` tags are present for each commit as required, and if so add the `dco-signoff: yes` label.

This and it will also add a `dco` check to the ci status on the PR.

> Will the submitter be notified automatically if the tags are not present, and told to add them?

Yes, prow writes a comment to the PR which explains what the user will have to do. Once the PR is signed-off, prow will delete the comment (a general pattern by prow, to first inform the user, but then remove the noise again from the PR)

andreabolognani:
> > So IIUC this will cause the KubeVirt CI to verify `Signed-off-by` tags are present for each commit as required, and if so add the `dco-signoff: yes` label.
> 
> This and it will also add a `dco` check to the ci status on the PR.
> 
> > Will the submitter be notified automatically if the tags are not present, and told to add them?
> 
> Yes, prow writes a comment to the PR which explains what the user will have to do. Once the PR is signed-off, prow will delete the comment (a general pattern by prow, to first inform the user, but then remove the noise again from the PR)

Sounds good, thanks for the explanation!

PR #218, opened 2019-10-07
Prepare prow for CDI

rmohr:


github/ci/prow/files/jobs/containerized-data-importer/OWNERS
```diff
@@ -0,0 +1,8 @@
+filters:
+  ".*":
+    reviewers:
+      - awels
+      - mhenriks
```
awels: can we add aglitke as well?

rmohr: Done

rmohr:
@awels I made all jobs optional, so that we can merge and test independently.

PR #224, opened 2019-10-14
Use latest image on KubeVirt tutorial tests

iranzo:
Signed-off-by: Pablo Iranzo Gómez <Pablo.Iranzo@gmail.com>


slintes:
Are you sure you want to use `latest`? Usually that is a bad practice, because you might pull in versions with breaking changes.

iranzo:
Hi @slintes our current image has issues with installing `bc` which requires updating 'read-only' system libraries and fails all the testing for the kubevirt-tutorial labs.

We're currently testing with the internal prow instance to decide which one we want to use (and that's why this is on /hold).

If you've any proposed image, we can move to that one instead in the meantime so that I can retest on it.

Thanks!

iranzo:
@slintes ready for you to review :-)

PR #228, opened 2019-10-15
Remove openshift-docker testlane for master

rmohr:
Docker is extensively tested on the k8s lanes. Now with release-0.22 we can stop testing the os-3.11 and docker combination on master.


slintes:
:+1: 

/lgtm
/approve

PR #251, opened 2019-11-13
Prepare PROW for hostpath provisioner

awels:

### DIFF
--- a/github/ci/prow/files/jobs/hostpath-provisioner/OWNERS
+++ b/github/ci/prow/files/jobs/hostpath-provisioner/OWNERS
@@ -1,0 +1,9 @@
+filters:
+  ".*":
+    reviewers:
+      - awels
+      - aglitke
+      - mhenriks
+      - j-griffith
+    approvers:
+      - awels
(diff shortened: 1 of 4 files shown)


awels:
@stu-gott @rmohr I am trying to add some CI and functional tests to the hostpath provisioner. I am following the example from CDI.

github/ci/prow/files/jobs/hostpath-provisioner/OWNERS
```diff
@@ -0,0 +1,6 @@
+filters:
+  ".*":
+    reviewers:
+      - awels
```
rmohr: Best to add someone else  from your team too. Otherweise you will either have to configure prow for your repo, that authors of their PR can also /lgtm their own code, or you will have to rely on external reviewers too.

awels: Sure, I can add some people.

awels:
@rmohr Since hostpath provisioner is already in kubevirt org, it should just work right?

rmohr:
Could be that we forgot to set something, but in general yes.

PR #252, opened 2019-11-14
Fix call to the cluster up script for hostpath provisioner.

awels:


awels:
@rmohr That should teach me to not use a script. I figured I could just call make cluster-up/sync/etc But then each time I need to fix something I have to bug you. So now just calling a script that I can manipulate from my repo.

PR #260, opened 2019-12-03
Add custom kubevirt-infra boostrap image build and publish

danielBelenky:
Some of our components need extra packages installed in
our test container. This PR adds a Dockerfile for this
container as well as Prow jobs to build and publish it.

### DIFF
--- a/github/ci/prow/files/jobs/project-infra/project-infra-presubmits.yaml
+++ b/github/ci/prow/files/jobs/project-infra/project-infra-presubmits.yaml
@@ -41,3 +41,29 @@ presubmits:
         args:
           - test
           - ./...
+  - name: build-kubevirt-infra-bootstrap-image
+    always_run: false
+    run_if_changed: "images/kubevirt-infra-bootstrap/.*"
+    decorate: true
+    labels:
+      preset-dind-enabled: "true"
+      preset-docker-mirror: "true"
+      preset-kubevirtci-docker-credential: "true"
+    spec:
+      nodeSelector:
+        type: vm
+      containers:
+        - image: gcr.io/k8s-testimages/bootstrap:v20190516-c6832d9
+          command:
+            - "/usr/local/bin/runner.sh"
+            - "/bin/bash"
+            - "-c"
+            - "cd images && ./publish_image.sh -b kubevirt-infra-bootstrap docker.io kubevirtci"
+          # docker-in-docker needs privileged mode
+          securityContext:
+            privileged: true
+          resources:
+            requests:
+              memory: "1Gi"
+            limits:
+              memory: "1Gi"
--- a/images/publish_image.sh
+++ b/images/publish_image.sh
@@ -1,0 +1,85 @@
+#!/bin/bash -xe
+
+
+main() {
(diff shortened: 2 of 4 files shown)


github/ci/prow/files/jobs/project-infra/project-infra-presubmits.yaml
```diff
@@ -40,4 +40,30 @@ presubmits:
...
+    spec:
+      nodeSelector:
+        type: vm
+      containers:
+        - image: gcr.io/k8s-testimages/bootstrap:v20190516-c6832d9
+          command:
+            - "/usr/local/bin/runner.sh"
+            - "/bin/bash"
+            - "-c"
+            - "cd images && publish_image -b ssp-operator-bootstrap docker.io kubevirtci"
```
rmohr: you pass `-b` here.

images/publish_image.sh
```diff
@@ -0,0 +1,71 @@
...
+
+# Usage:
+#    ./publish_image.sh [OPTIONS] BUILD_TARGET REGISTRY REGISTRY_ORG
+#
+#    Build and publish infra images.
+#    If the -d flag is specified, will only build.
+
+main() {
+    local build_only
+    while getopts "d" opt; do
```
rmohr: You talk about `-d` here.

images/publish_image.sh
```diff
@@ -0,0 +1,71 @@
...
+        build_image "$full_image_name"
+    )
+    [[ $build_only ]] && return
+    publish_image "$full_image_name"
+}
+
+get_image_tag() {
+    local current_commit
+    current_commit="$(git rev-parse HEAD)"
+    echo "${current_commit:0:6}"
```
rmohr: Can we also add the date the same way like it is done on the test-infra bootstrap image?

github/ci/prow/files/jobs/project-infra/project-infra-presubmits.yaml
danielBelenky: fixed, thanks

images/publish_image.sh
danielBelenky: The problem that I see with adding dates to tags is that you can build twice and get different builds from the same commit. Is this something that we want?

danielBelenky:
> LGTM, I would just suggest to not use name ssp-operator-bootstrap (the image will not be used for ssp operator but for common-templates), but something more generic, like kubevirt-bootstrap

I will rename the image :+1:

images/publish_image.sh
rmohr: Yes, because we don't have reproducible builds. It just reflects reality. It may or may not be different and therefore it should **not** look like the last build and should clearly be distinctable.

danielBelenky:
I've updated the PR with the requested fixes

images/publish_image.sh
```diff
@@ -0,0 +1,72 @@
+#!/bin/bash -xe
+
+# Usage:
```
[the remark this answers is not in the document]
danielBelenky: Good idea. Fixed.

PR #262, opened 2019-12-09
Add presets with kubevirtci secret

danielBelenky:
In order to publish the container to our org in dockerhub,
we need to have the credentials mounted.

### DIFF
--- a/github/ci/prow/files/jobs/project-infra/project-infra-postsubmits.yaml
+++ b/github/ci/prow/files/jobs/project-infra/project-infra-postsubmits.yaml
@@ -8,7 +8,7 @@ postsubmits:
       labels:
         preset-dind-enabled: "true"
         preset-docker-mirror: "true"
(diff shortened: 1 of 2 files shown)


github/ci/prow/files/jobs/project-infra/project-infra-postsubmits.yaml
```diff
@@ -8,7 +8,7 @@ postsubmits:
       labels:
         preset-dind-enabled: "true"
         preset-docker-mirror: "true"
-        preset-kubevirtci-docker-credential: "true"
+        preset-project-infra-kubevirtci-docker-credentia: "true"
```
ksimon1: I think it is missing 'l' at the end of the name?

danielBelenky: fixed

PR #263, opened 2019-12-09
update name of common-templates tests

ksimon1:
change of base image
tests will now always run

Signed-off-by: Karel Šimon <ksimon@redhat.com>

### DIFF
--- a/github/ci/prow/files/jobs/common-templates/common-templates-presubmits.yaml
+++ b/github/ci/prow/files/jobs/common-templates/common-templates-presubmits.yaml
@@ -1,10 +1,10 @@
 presubmits:
   kubevirt/common-templates:
-  - name: pull-rhel6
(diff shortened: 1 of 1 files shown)


github/ci/prow/files/jobs/common-templates/common-templates-presubmits.yaml
```diff
@@ -138,7 +138,7 @@ presubmits:
       nodeSelector:
         type: bare-metal-external
       containers:
-      - image: gcr.io/k8s-testimages/bootstrap:v20190516-c6832d9
+      - image: quay.io/ksimon/common-templates-e2e-env:v0.1.0
```
dhiller: Why is this one different?

ksimon1: Good catch, fixed

PR #264, opened 2019-12-10
enable windows 2019 test for common templates

ksimon1:
This PR enables win 2019 test. This job is optional, because i need to verify if win2019 can boot in prow.
//cc @dhiller 
Signed-off-by: Karel Šimon <ksimon@redhat.com>


rmohr:
@ksimon1 just in case you are not aware of it, since you own that config, you can approve it yourself when you have a lgtm.

ksimon1:
@dhiller can you please review?

PR #293, opened 2019-12-30
Fix ansible-kubevirt-module job

gbenhaim:
- The job of this projects requires pip, which was installed
  in the slave image that was added in this commit.

- Fix the job's name.

Signed-off-by: gbenhaim <galbh2@gmail.com>


gbenhaim:
verified https://github.com/kubevirt/ansible-kubevirt-modules/pull/278

PR #294, opened 2020-01-02
Make sriov lane optional

danielBelenky:
Until we can resolve the issues we see in this lane,
we should not block PRs from being merged.

### DIFF
--- a/github/ci/prow/files/jobs/kubevirt/kubevirt-presubmits.yaml
+++ b/github/ci/prow/files/jobs/kubevirt/kubevirt-presubmits.yaml
@@ -391,7 +391,7 @@ presubmits:
       fork-per-release: "true"
       k8s.v1.cni.cncf.io/networks: multus-cni-ns/sriov-passthrough-cni
     always_run: true
-    optional: false
+    optional: true
     decorate: true
     decoration_config:
       timeout: 5h


github/ci/prow/files/jobs/kubevirt/kubevirt-presubmits.yaml
```diff
@@ -391,7 +391,7 @@ presubmits:
       fork-per-release: "true"
       k8s.v1.cni.cncf.io/networks: multus-cni-ns/sriov-passthrough-cni
     always_run: true
```
rmohr: That should have been set to `false` too, I think.

danielBelenky: I thought so too, but Felix wanted it to run on every PR so we can track the progress of the fix.
I don't mind changing it...

rmohr: `required` is meant for the branch-protector to mark the job as required (if I can for instance just hit the merge button when jobs are failing). But tide will simply not merge anything when it sees a failing job.

danielBelenky: ah, I had the wrong assumption that it will allow us to merge with this job failing :/ anyway, the issue should be solved now when the faulty node is offline. Do you want me to still set this job to not always run?

PR #297, opened 2020-01-05
Disable auto-run for the sr-iov lane

danielBelenky:
Until we sort the issues with this lane, we don't want to block the
progress of unrelated PRs.

### DIFF
--- a/github/ci/prow/files/jobs/kubevirt/kubevirt-presubmits.yaml
+++ b/github/ci/prow/files/jobs/kubevirt/kubevirt-presubmits.yaml
@@ -391,6 +391,7 @@ presubmits:
       fork-per-release: "true"
       k8s.v1.cni.cncf.io/networks: multus-cni-ns/sriov-passthrough-cni
     always_run: true
+    skip_report: true
     optional: true
     decorate: true
     decoration_config:


phoracek:
Daniel, what is the effect of this. Does it mean SR-IOV lane will be executed only on demand?

danielBelenky:
> Daniel, what is the effect of this. Does it mean SR-IOV lane will be executed only on demand?

yes

phoracek:
Could we hold for a week, let's check if we just need to refresh those nodes and make it work for a while. Or even better, fix it if it is not a huge issue. I'd like to get into the CNI we use to see if we can make it more stable.

rmohr:
@phoracek for as long as it is broken it has no value. Why can't we enable it again when it is fixed?

phoracek:
To collect data about failed runs. There is some noise around the main issue we want to solve, having more runs recorded could help is find the cause. Also, once it is out of sight, it may become our of mind too. I'm only asking for 5 days hold, then feel free to disable it.

danielBelenky:
> To collect data about failed runs. There is some noise around the main issue we want to solve, having more runs recorded could help is find the cause. Also, once it is out of sight, it may become our of mind too. I'm only asking for 5 days hold, then feel free to disable it.

rmohr:
@phoracek we are not talking about flaky tests which sometimes fail and still have some value. You have plenty of getting-started data in https://prow.apps.ovirt.org/job-history/kubevirt-prow/pr-logs/directory/pull-kubevirt-e2e-kind-k8s-sriov-1.14.2. You can also still trigger on a WIP PR the jobs.
Red test lanes on pretty much all PRs are not the way to go when debugging infra issues. It has direct impact on PR authors and reviewers and breaks the automatic merge flows.

I am sure @danielBelenky can assist you with directly spawning as many prow jobs in parallel as you want, or with directly creating pods on nodes which you want to debug. PRs where people want to get something done and tested are not there to indirectly test code changes on e.g. a CNI plugin while debugging (if we would only talk about flakes this would of course be different but we talk about an almost completely broken job).

rmohr:
/reopen
/lgtm
/approve

not removing the hold yet.

danielBelenky:
Just want to state that I closed this PR by mistake.
I agree with Roman here...
@phoracek let's talk tomorrow and see what is required to debug this issue.

phoracek:
I'm on PTO tomorrow, let's discuss it on Tuesday, thanks.

BTW how does this affect automatic merges? I thought that's resolved by marking the job as not required.

rmohr:
/lgtm cancel

Actually, please leave the `always_run` set to `true` and instead add `skip_report: true`. Then it is not reported to the PR and @phoracek gets the test results.

rmohr:
> BTW how does this affect automatic merges? I thought that's resolved by marking the job as not required.

No, `required` is there for the branch protector, so that the job is configured as a required job on github. But I think that `skip_report` is what we want.

danielBelenky:
updated

rmohr:
@danielBelenky thx.

PR #298, opened 2020-01-08
Add knmstate job to publish after master merge

qinqon:


github/ci/prow/files/jobs/nmstate/kubernetes-nmstate-postsubmits.yaml
```diff
@@ -0,0 +1,28 @@
+postsubmits:
+  kubevirt/project-infra:
```
phoracek: Is this correct?

qinqon: Nope, fixed it

qinqon:
/hold

Looks like there is already a knmstate quay.io, we may need to use that instead of kubevirtbot.

qinqon:
/hold cancel

kubevirtbot has being added to nmstate org so we good.

PR #303, opened 2020-01-13
External hook plugin to trigger ProwJobs from PRs

danielBelenky:
To ease modifications of ProwJobs, we want the ability to trigger new
jobs upon new PRs with changes to some ProwJobs.

**This is still WIP, but will appreciate reviews**


robots/prowjob-experiment/handler/handler.go
```diff
@@ -0,0 +1,320 @@
...
+		}
+		if match {
+			out = append(out, s)
+		}
+	}
+	return out, nil
+}
+
+// Determine if we should perform on the given event.
+func shouldPerformOnEventAction(event *github.PullRequestEvent) bool {
```
dhiller: This method name does convey that it checks whether an action should be performed. However, to check in what case an action will be performed you need to read the method body. I'd recommend renaming the method to better convey that the following code should only run if a PR was opened, edited or synced. 

In short I mean, the method name talks about how it does things, but not what it does.

This is not an easy topic for me to explain, hopefully this makes sense to you.

robots/prowjob-experiment/handler/handler.go
```diff
@@ -0,0 +1,320 @@
...
+	if err := repo.Checkout(ref); err != nil {
+		return nil, err
+	}
+
+	return loadConfigs(repo.Directory(), prowConfigPath, jobConfigPaths), nil
+}
+
+// Squash original and modified configs at the same path, returning an array of configs
+// that contain only new and modified job configs.
+func squashConfigPaths(originalConfigs, modifiedConfigs jobConfigPath) []*config.Config {
```
dhiller: This method name sounds like there is more to come here? Is that intended? I mean, wouldn't we want to separate other things into other robots? We could share code in a common base though...

dhiller:
One afterthought here: In general I'd recommend that `prowjob-experiment` should be changed into something that clearly expresses what this is about, as I guess if it's working we need to do this anyway. This late bite would generate more work in the end, so I'd rather do it now.

Note: In OpenShift-CI they call check-running of presubmit jobs `rehearsals`, maybe we can borrow this phrase. I.e. sth. like `presubmit-rehearsal`.

PR #310, opened 2020-01-22
Add nmstate/kubernetes-nmstate branch protection

qinqon:

### DIFF
--- a/github/ci/prow/files/config.yaml
+++ b/github/ci/prow/files/config.yaml
@@ -180,6 +180,12 @@ branch-protection:
               protect: true
         kubectl-virt-plugin:
           protect: true
+    nmstate:
+      repos:
+        kubernetes-nmstate:
+          branches:
+            master:
+              protect: true
 
 push_gateway:
   endpoint: pushgateway


github/ci/prow/files/config.yaml
```diff
@@ -180,6 +180,14 @@ branch-protection:
               protect: true
         kubectl-virt-plugin:
           protect: true
+    nmstate:
+      repos:
+        kubernetes-nmstate:
+          protect: true
+          required_status_checks:
+            contexts:
+              - pull-kubernetes-nmstate-unit-test
```
dhiller: these are not necessary to mention, branchprotection will add them automatically from the presubmits and make them required, if they are in the job definition.

qinqon: so if they are marked at 'required' at job definitions they will bee added right ?

dhiller: Yes, that should work. Please see that these are picked up for kubevirt/kubevirt also automatically. We just added travis and coverage.

PR #319, opened 2020-01-28
Add ipv6 config for docker daemon

qinqon:
To run ipv6 jobs we need docker daemon to be properly configured as stated here  kubevirt/kubevirtci#241.


dankenigsberg:
I think that it is important state the purpose of this ( https://github.com/kubevirt/kubevirtci/pull/241 ), and to call in @yuvalk to review it.

yuvalk:
yep that the two I'm using
note that afaik this give dual ipv4/6

/lgtm

qinqon:
@yuvalk we are going to apply this changes tomorrow so we will be able to run the kind ipv6 jobs at your PR, although I have manually tested and it works fine, it's ok ?

PR #322, opened 2020-01-29
K8s fedora jobs

qinqon:


github/ci/prow/files/jobs/kubevirtci/kubevirtci-presubmits.yaml
```diff
@@ -257,6 +257,22 @@ presubmits:
         resources:
           requests:
             memory: "15Gi"
+        volumeMounts:
```
dhiller: Altough this has nothing to do with fedora, I guess we're still good here.

PR #329, opened 2020-02-05
Make NFS CDI lane required.

awels:


fabiand:
/lgtm
/approve

Let's see if this works ;)

awels:
You need to add /approved with a d.

fabiand:
I think I also need to be on the approvers list... which is ok that I am
not on it ;)

PR #330, opened 2020-02-05
Bump prow to v20200204-7e8cd997a

rmohr:
Bump to latest prow.

### DIFF
--- a/github/ci/prow/files/config.yaml
+++ b/github/ci/prow/files/config.yaml
@@ -7,10 +7,10 @@ plank:
     timeout: 2h
     grace_period: 15s
     utility_images:
-      clonerefs: "gcr.io/k8s-prow/clonerefs:v20190703-132f32f32"
-      initupload: "gcr.io/k8s-prow/initupload:v20190703-132f32f32"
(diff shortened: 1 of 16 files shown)


dhiller:
What was the matter with the failing prow check? Was it just, that the check was old, like it was not the right one that should be applied by the PR?

dhiller:
Otherwise 

/lgtm

rmohr:
> What was the matter with the failing prow check? Was it just, that the check was old, like it was not the right one that should be applied by the PR?

The checkconfig job is not yet updated. It will pass after we merged this PR. Here my local check:

```
$ docker run --rm -v ${PWD}/:/prow  -it gcr.io/k8s-prow/checkconfig:v20200204-7e8cd997a  --config-path /prow/files/config.yaml --job-config-path /prow/files/jobs --plugin-config /prow/files/plugins.yaml --strict
{"component":"checkconfig","file":"prow/cmd/checkconfig/main.go:325","func":"main.main","level":"info","msg":"checkconfig passes without any error!","time":"2020-02-06T09:11:34Z"}
```

dhiller:
So then, fasten your seat belts :)

dhiller:
And always wear a helmet :-D

PR #348, opened 2020-02-13
add presubmit for knmstate 0.15

phoracek:
We branched out, let's keep testing.


phoracek:
@qinqon good catch, thanks. I added it.

PR #356, opened 2020-02-19
knmstate: Let's remove k8s memory req to see if we good

qinqon:


rmohr:
/hold

please not. That will just increase the mess.

PR #361, opened 2020-02-20
knmstate: Run k8s lane on VM

qinqon:


rmohr:
/hold

We need to be careful with the presets here. Let me review.

github/ci/prow/files/jobs/nmstate/kubernetes-nmstate-presubmits.yaml
```diff
@@ -49,7 +49,8 @@ presubmits:
         preset-shared-images: "true"
```
rmohr: You can keep `preset-dind-enabled`, but `preset-docker-mirror` needs to  be removed, so that you don't try to use the docker mirror in the other part of the datacenter.

PR #373, opened 2020-02-25
adding kustomize tool to bootstrap Docker file

RamLavi:
[kustomize](https://kustomize.io/) is a tool used to customize application configuration files. 

Currently [kubemacpool project](https://github.com/k8snetworkplumbingwg/kubemacpool) uses this tool during the ci tests, and so there is a need to add it to the Bootstrap Docker file (currently failing due to this). 

### DIFF
--- a/images/kubevirt-infra-bootstrap/Dockerfile
+++ b/images/kubevirt-infra-bootstrap/Dockerfile
@@ -16,3 +16,10 @@ RUN apt-get update \
         python3 \
         python3-pip \
         && rm -rf /var/lib/apt/lists/*
+RUN export KUSTOMIZE_DIR=/opt/kustomize \
+    && mkdir -p $KUSTOMIZE_DIR \
+    && cd $KUSTOMIZE_DIR \
+    && wget "https://github.com/kubernetes-sigs/kustomize/releases/download/kustomize%2Fv3.5.4/kustomize_v3.5.4_linux_amd64.tar.gz" \
+    && tar xzf ./kustomize_v3.5.4_linux_amd64.tar.gz \
+    && rm kustomize_v3.5.4_linux_amd64.tar.gz \
+    && echo "export PATH=$KUSTOMIZE_DIR:$PATH" >> $HOME/.bashrc


images/kubevirt-infra-bootstrap/Dockerfile
```diff
@@ -16,3 +16,11 @@ RUN apt-get update \
         python3 \
         python3-pip \
         && rm -rf /var/lib/apt/lists/*
+RUN wget https://dl.google.com/go/go1.13.3.linux-amd64.tar.gz \
```
phoracek: Maybe it would be easier using gimme for the installation as done here https://github.com/kubevirt/cluster-network-addons-operator/blob/master/automation/check-patch.e2e.setup.sh? Not sure what is the better approach. Just for your consideration.

RamLavi: I agree - gimme approach more elegant.

images/kubevirt-infra-bootstrap/Dockerfile
```diff
@@ -16,3 +16,14 @@ RUN apt-get update \
         python3 \
         python3-pip \
         && rm -rf /var/lib/apt/lists/*
+RUN export GOROOT=/tmp/go/root \
```
phoracek: We better use some standard directory. I guess /opt would fit. There are no guarantees about /tmp surviving

images/kubevirt-infra-bootstrap/Dockerfile
```diff
@@ -16,3 +16,14 @@ RUN apt-get update \
...
+    && mkdir -p $GOROOT \
+    && export GOPATH=/opt/go/path \
+    && mkdir -p $GOPATH \
+    && export PATH=$GOPATH/bin:$GOROOT/bin:$PATH \
+    && export GIMME_GO_VERSION=1.12 \
+    && GIMME=/opt/go/gimme \
+    && mkdir -p $GIMME \
+    && curl -sL https://raw.githubusercontent.com/travis-ci/gimme/master/gimme | HOME=${GIMME} bash >> ${GIMME}/gimme.sh \
+    && . ${GIMME}/gimme.sh \
+    && GO111MODULE=on go get sigs.k8s.io/kustomize
```
slintes: just curious, why not just downloading kustomize from its release page? Installing Go just for this seems quite some overhead to me :) 
https://github.com/kubernetes-sigs/kustomize/releases/tag/kustomize%2Fv3.5.4

RamLavi: Totally agree. Thanks!

images/kubevirt-infra-bootstrap/Dockerfile
```diff
@@ -16,3 +16,9 @@ RUN apt-get update \
         python3 \
         python3-pip \
         && rm -rf /var/lib/apt/lists/*
+RUN export KUSTOMIZE_DIR=/opt/kustomize \
```
phoracek: Would https://docs.docker.com/engine/reference/builder/#env help with all the exports?

phoracek: Sorry, it is not actually that helpful since it would be lost after the build is done.

images/kubevirt-infra-bootstrap/Dockerfile
```diff
@@ -16,3 +16,9 @@ RUN apt-get update \
...
+    && mkdir -p $KUSTOMIZE_DIR \
+    && cd $KUSTOMIZE_DIR \
+    && curl -s "https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/hack/install_kustomize.sh"  | bash \
+    && echo "export PATH=$KUSTOMIZE_DIR:$PATH" >> $HOME/.bashrc \
+    && . $HOME/.bashrc
```
phoracek: Not needed

images/kubevirt-infra-bootstrap/Dockerfile
```diff
@@ -16,3 +16,8 @@ RUN apt-get update \
         python3 \
         python3-pip \
         && rm -rf /var/lib/apt/lists/*
+RUN export KUSTOMIZE_DIR=/opt/kustomize \
+    && mkdir -p $KUSTOMIZE_DIR \
+    && cd $KUSTOMIZE_DIR \
+    && curl -s "https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/hack/install_kustomize.sh"  | bash \
```
slintes: We usually try to avoid using master / latest. It introduces some risk when it updates to a new, maybe incompatible version, which will break CI. Do you mind using a specific version?

slintes: (and with this `curl ... | bash` flow it introduces even bigger risks)

RamLavi: sure!

slintes: and looking at the script, it always downloads the latest release.... so, please download a verioned kustomize binary from the release page, see e.g. https://github.com/kubernetes-sigs/kustomize/releases/tag/kustomize%2Fv3.5.4

slintes:
/lgtm

maybe @dhiller wants to have a look as well :)

PR #379, opened 2020-03-05
kubevirt: Add cnao k8s 1.17 lane.

qinqon:


phoracek:
/lgtm
/hold

Holding so we can test it against kubevirt

qinqon:
@phoracek at the end I am just adding a new lane instead of replace, so we can merge this and when job is ok after few runs we set this at mandatory and remove the old one.

phoracek:
Ok, then feel free to cancel the hold

PR #383, opened 2020-03-09
Add macvtap-cni postsubmit jobs

maiqueb:
Signed-off-by: Miguel Duarte Barroso <mdbarroso@redhat.com>

### DIFF
--- a/github/ci/prow/files/jobs/macvtap-cni/macvtap-cni-postsubmits.yaml
+++ b/github/ci/prow/files/jobs/macvtap-cni/macvtap-cni-postsubmits.yaml
@@ -1,0 +1,26 @@
+postsubmits:
+  kubevirt/macvtap-cni:
+    - name: publish-macvtap-cni
+      branches:
+      - master
+      always_run: true
+      decorate: true
+      max_concurrency: 1
+      labels:
+        preset-dind-enabled: "true"
+        preset-docker-mirror: "true"
+        preset-kubevirtci-quay-credential: "true"
+      spec:
+        nodeSelector:
+          type: vm
+          zone: ci
+        containers:
+          - image: gcr.io/k8s-testimages/bootstrap:v20190516-c6832d9
+            command:
+              - "/usr/local/bin/runner.sh"
+              - "/bin/bash"
+              - "-c"
+              - "cat $QUAY_PASSWORD | docker login --username $(cat $QUAY_USER) --password-stdin=true quay.io && ./automation/publish.sh"
+            # docker-in-docker needs privileged mode
+            securityContext:
+              privileged: true


github/ci/prow/files/jobs/macvtap-cni/macvtap-cni-postsubmits.yaml
```diff
@@ -0,0 +1,24 @@
+postsubmits:
+  kubevirt/macvtap-cni:
```
phoracek: We gotta limit it to master. Otherwise minor-branch merged will override master.

maiqueb: Done.

github/ci/prow/files/jobs/macvtap-cni/macvtap-cni-postsubmits.yaml
```diff
@@ -0,0 +1,26 @@
...
+        nodeSelector:
+          type: vm
+          zone: ci
+        containers:
+          - image: gcr.io/k8s-testimages/bootstrap:v20190516-c6832d9
+            command:
+              - "/usr/local/bin/runner.sh"
+              - "/bin/bash"
+              - "-c"
+              - "cat $QUAY_PASSWORD | docker login --username $(cat $QUAY_USER) --password-stdin=true quay.io && ./automation/publish.sh"
```
qinqon: If we publish at quay.io without tag the containers get deleted, do we implement that mechanism ? ( I know it's not part of this PR, just to bring some concerns here).

phoracek: And if we use :latest? Does that solve the issue? Can we just add that tag as the default to macvtap-cni Makefile?

qinqon: Best approach is something like this https://github.com/kubevirt/kubevirtci/blob/master/cluster-provision/ocp/4.4/publish.sh so you can know by the tag the date and commit id.

phoracek: The issue is that nobody would use that tag. In readme of macvtap, we would just reference to latest to try out master (we cannot refresh all the references on every PR) and on CNAO, we would use a specific tag anyway. I don't see a big benefit in data-based tagging unless there is a user for it.

phoracek: :latest is the default in macvtap-cni Makefile. I think it is good enough, even if we remove old images. :latest is just for users to try out master. The proper way of using it is to pin to released versions.

qinqon:
/lgtm
/approve

fair enough

github/ci/prow/files/jobs/macvtap-cni/macvtap-cni-postsubmits.yaml
qinqon: fair enough, important part is to ensure that is tagged if not quay.io will remove it.

maiqueb: I understand both use cases - and even think they're both useful.

Could we push the versioned tag (at least w/ commit-id) *and* move the latest tag all the time ? 

As a developer I would get traceability between containers & code; as a user I get a simple workflow (pull latest).

phoracek: We can continue the discussion on https://github.com/kubevirt/macvtap-cni/pull/12

PR #392, opened 2020-03-14
switch cnao from okd to ocp

phoracek:
Needed for: https://github.com/kubevirt/cluster-network-addons-operator/pull/350


phoracek:
@stu-gott this change is irrelevant now. We should just drop OKD instead.  Closing, thanks for checking it

PR #398, opened 2020-03-25
Move unit test to prow

qinqon:


github/ci/prow/files/jobs/kubevirt/kubevirt-presubmits.yaml
```diff
@@ -540,3 +540,29 @@ presubmits:
...
+      nodeSelector:
+        type: vm
+        zone: ci
+      containers:
+        - image: kubevirt/builder@sha256:7585e27b5a5a019f5749eaa1ef29da3278878eefc862b45d6e1c2e343d9f8859
+          env:
+            - name: "KUBEVIRT_RUN_UNNESTED"
+              value: "true"
+          args:
+            - automation/unit-test.sh
```
rmohr, line 568: Can't we just do `make test` here, or is this to allow modifying it without having to touch the job?

github/ci/prow/files/jobs/kubevirt/kubevirt-presubmits.yaml
```diff
@@ -540,3 +540,29 @@ presubmits:
...
+    always_run: true
+    optional: false
+    decorate: true
+    decoration_config:
+      timeout: 3h
+      grace_period: 5m
+    max_concurrency: 6
+    labels:
+      preset-docker-mirror: "false"
+      preset-shared-images: "true"
```
rmohr, line 557: I don't think that we need the shared images here.

github/ci/prow/files/jobs/kubevirt/kubevirt-presubmits.yaml
```diff
@@ -540,3 +540,29 @@ presubmits:
           resources:
             requests:
               memory: "29Gi"
+  - name: pull-kubevirt-unit-test
+    skip_branches:
+      - release-\d+\.\d+
+    annotations:
+      fork-per-release: "true"
+    always_run: true
```
rmohr, line 548: Can we set skip_report to true? We can then merge it and see how it goes.

github/ci/prow/files/jobs/kubevirt/kubevirt-presubmits.yaml
qinqon: Yep, so we can reduce changes at project-infra, for example next PR should be related to storing junit xml artifacts and that would be dont at unit-test.sh

rmohr:
Superseeded by #472.

@qinqon thanks again for starting this.

PR #399, opened 2020-03-25
Add vm-import-operator job

machacekondra:


machacekondra:
@dhiller Hi, can you please enable the trigger on: https://github.com/kubevirt/vm-import-operator/

dhiller:
@machacekondra you need to enable prow trigger plugin for your repo [here](https://github.com/kubevirt/project-infra/blob/master/github/ci/prow/files/plugins.yaml#L24)
and you need to add your repo to the trigger config [here](https://github.com/kubevirt/project-infra/blob/master/github/ci/prow/files/plugins.yaml#L234)

rmohr:
/hold

Does this repo need the openshift console for some testing? If not I would suggest to not use an openshift lane. They are very unstable, slow and extremely resource intensive (our CI resources are limited for e2e tests). I would consider it as not suitable for testing on PRs if you want reasonable feedback.

machacekondra:
It doesn't need openshift console, but it needs openshift templates to have some real case scenarios for testing.

machacekondra:
@cynepco3hahue @danielBelenky Hi, can you please review?

pkliczewski:
@rmohr Unfortunetally we need OCP because common templates are part of core functionality. We understand it is unstable but it is much better to have something than nothing. We want to run the test daily. I would appreciate review and/or ack on this PR. We need it to make progress with the operator.

rmohr:
> We understand it is unstable but it is much better to have something than nothing.

Ack.

> We want to run the test daily.

If you want to run it daily you will have to make it a periodic job.

pkliczewski:
@rmohr please rereview

github/ci/prow/files/jobs/vm-import-operator/vm-import-operator-periodics.yaml
```diff
@@ -0,0 +1,32 @@
...
+  spec:
+    nodeSelector:
+      type: bare-metal-external
+    containers:
+      - image: gcr.io/k8s-testimages/bootstrap:v20190516-c6832d9
+        command:
+          - "/usr/local/bin/runner.sh"
+          - "/bin/sh"
+          - "-c"
+          - "export TARGET=ocp-4.3 && automation/test.sh"
```
rmohr, line 26: Since this is now a periodic, the git repo is not checked out. I think that it needs to looks similar to this one:

```
      command:
      - "/usr/local/bin/runner.sh"
      - "/bin/sh"
      - "-c"
      - "git clone https://github.com/kubevirt/vm-import-operator.git && cd kubevirt && export TARGET=okd-4.1 && automation/test.sh"
```

pkliczewski:
@rmohr What do you think now?

pkliczewski:
:tada:

PR #403, opened 2020-04-01
Add labels support to kubernetes-nmstate project

qinqon:

### DIFF
--- a/github/ci/prow/files/labels.yaml
+++ b/github/ci/prow/files/labels.yaml
@@ -428,3 +428,59 @@ repos:
         target: prs
         prowPlugin: approve
         addedBy: approvers
+  nmstate/kubernetes-nmstate:
+    labels:
+      - name: priority/highest
+        color: 9999e6
+        target: both
+      - name: priority/high
+        color: adadeb
+        target: both
+      - name: priority/medium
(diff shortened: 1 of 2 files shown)


qinqon:
/cc @phoracek 
labels

github/ci/prow/files/labels.yaml
```diff
@@ -428,3 +428,56 @@ repos:
         target: prs
         prowPlugin: approve
         addedBy: approvers
+  nmstate/kubernetes-nmstate:
+    labels:
+      - name: priority/highest
```
dhiller, line 433: This is a label that exists in a different form in the [target repo](https://github.com/nmstate/kubernetes-nmstate/labels) where it's `priority:highest` instead of `priority/highest`. Maybe you should rename your labels before applying this?

dhiller: I mean, you should either change them in the target repo if that is an option (I think that GitHub support renaming labels and keeps the issues with the updated labels). Or you should change them here.

qinqon: Super good point, have renamed the labels to the ones we have at this PR so they track same issues/prs

PR #413, opened 2020-04-23
Make k8s-1.17 job report, but don't run it always

rmohr:
Let's reduce the CI load while this lane can't succeed. The Job will now report to github, but it will only be run if you explicitly do

```
/test pull-kubevirt-e2e-k8s-1.17
```

### DIFF
--- a/github/ci/prow/files/jobs/kubevirt/kubevirt-presubmits.yaml
+++ b/github/ci/prow/files/jobs/kubevirt/kubevirt-presubmits.yaml
@@ -101,9 +101,9 @@ presubmits:
       - release-\d+\.\d+
     annotations:
       fork-per-release: "true"
-    always_run: true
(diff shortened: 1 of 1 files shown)


rmohr:
@vladikr @jean-edouard @davidvossel would that makes sense? The lane just can't pass and runs for ages due to the test failures which creates a lot of extra load to our CI. That should also help with flaky tests.

oxtbuilder:
@rmohr lgtm, but `always_run` should probably be set back to true once a multiqueue workaround is in place.

rmohr:
@oxtbuilder definitely. As soon as all tests can pass on the lane, we can enable it permanently and kick out the 1.14 lane.

vladikr:
make sense to me, at least until we will merge #3299

PR #414, opened 2020-04-23
Kubernetes 1.14 is EOL

rmohr:
Disabling the testing against it.


rmohr:
@davidvossel @jean-edouard @vladikr @dhiller we may already want to disable k8s 1.14, since it is EOL even though we don't have 1.17 ready.

I think that it was a mistake from us to make 1.17 immediately dependent on selinux. It delayed the introduction of the lane and holds us back to test on 1.17, so let's not do something like this in the future. However since 1.14 is already EOL it may probably still make sense to already disable the lane.

PR #420, opened 2020-04-30
knmstate: Operator e2e jobs

bcrochet:
Add new operator e2e jobs and rename existing jobs as handler jobs.

Signed-off-by: Brad P. Crochet <brad@redhat.com>

### DIFF
--- a/github/ci/prow/files/jobs/nmstate/kubernetes-nmstate-presubmits.yaml
+++ b/github/ci/prow/files/jobs/nmstate/kubernetes-nmstate-presubmits.yaml
@@ -31,7 +31,7 @@ presubmits:
               - "-c"
               - "automation/check-patch.unit-test.sh"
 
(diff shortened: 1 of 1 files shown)


github/ci/prow/files/jobs/nmstate/kubernetes-nmstate-presubmits.yaml
```diff
@@ -127,3 +127,100 @@ presubmits:
...
+            resources:
+              requests:
+                memory: "29Gi"
+            command:
+              - "/usr/local/bin/runner.sh"
+              - "/bin/sh"
+              - "-c"
+              - "automation/check-patch.e2e-operator-k8s.sh"
+
+    - name: pull-kubernetes-nmstate-e2e-operator-okd
```
phoracek, line 163: I don't think we should add okd and ocp lanes that would start their life disabled. Let's just add the k8s one for the operator.

PR #437, opened 2020-05-16
Since ipv6 lane is green - make it required

AlonaKaplan:
Since ipv6 was introduced to kubevirt few PRs broke the lane.
To avoid this situation in the future, the lane should be required.


rmohr:
I see it green few times lately. Let's see how it goes.

/approve

PR #438, opened 2020-05-17
Make pull-kubevirt-e2e-k8s-1.17 lane mandatory

oshoval:

### DIFF
--- a/github/ci/prow/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
+++ b/github/ci/prow/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
@@ -102,7 +102,7 @@ presubmits:
     annotations:
       fork-per-release: "true"
     always_run: true
-    optional: true
+    optional: false
     skip_report: false
     decorate: true
     decoration_config:


danielBelenky:
/hold

I'm not 100% sure what is our process with regards to flake-finder.
@dhiller can you elaborate if we need to run tests before we mark a lane as mandatory?

oshoval:
Thanks, the line does run automatically,
but i saw it doesn't have the "Required" gray thumbnail

oshoval:
@qinqon 
Hi, do you know please if there is a reason that this lane isn't "Required"?

EdDev:
/lgtm

Makes sense to make this lane mandatory, even if flaky. It should cause us to fix the needed or skip the flaky tests (because they are badly written or because there is a bug in production that needs an opened ticket).

If no one give a reason to block this, lets proceed and in the worse case we can revert.

oshoval:
@danielBelenky 
@dhiller 
Hi, can we proceed with this one please ?

dhiller:
I would leave the final call to @jean-edouard because he's looking at the stability of this page. Otherwise lgtm.

dhiller:
> /hold
> 
> I'm not 100% sure what is our process with regards to flake-finder.
> @dhiller can you elaborate if we need to run tests before we mark a lane as mandatory?

Forgot to answer directly, sorry for that. With regards to the [last weekly flakefinder report](https://storage.googleapis.com/kubevirt-prow/reports/flakefinder/kubevirt/kubevirt/flakefinder-2020-05-19-168h.html), the lane looks slightly more flaky than the others. AFAIU @jean-edouard was fxing SELinux issues on this lane, so the test failures might be remnants of that effort, and therefore we might disregard them. But like I said, I'd leave that call to @jean-edouard .

jean-edouard:
/lgtm
The lane has SELinux running in enforcing mode, making it mandatory will ensure new code doesn't introduce AVC denials. If that happens, the denials can be found in the artifacts (auditlog).
I've been keeping an eye on the flaky tests to make sure none ever failed because of a denial, but that would be improbable anyway. While SELinux could definitely break tests, it would do so consitently.

PR #441, opened 2020-05-21
sriov: don't show results on PRs

phoracek:
This lane is unstable. We want to collect logs, so it runs on the
background, but it should not add noise to PRs.


phoracek:
@maiqueb @qinqon makes sense?

maiqueb:
> @maiqueb @qinqon makes sense?

Having the lane being reported, and bothering everyone in the organization can be seen as 'shame driven development' (forcing us to fix this).

But bothering everyone seems a bit of a stretch. Let's be polite.

PR #443, opened 2020-05-21
Release Blocker plugin

davidvossel:
This is the initial work for a flow I'm working on for automating kubevirt releases.

This plugin isn't perfect yet. I'm hoping to get this initial work merged and deployed so I can begin iterating on the idea further. For now, this plugin has no behavior change other than setting/removing a label and that label functionally means nothing at the moment.

I'm more interested in going through the process of wiring up a new plugin, deploying it, and seeing it in action than I am making this feature complete right away.


external-plugins/release-blocker/server.go
```diff
@@ -0,0 +1,219 @@
...
+	"k8s.io/test-infra/prow/config"
+	"k8s.io/test-infra/prow/github"
+	"k8s.io/test-infra/prow/pluginhelp"
+	"k8s.io/test-infra/prow/plugins"
+)
+
+const pluginName = "release-blocker"
+const baseLabel = "release-blocker"
+
+var releaseBlockRe = regexp.MustCompile(`(?m)^(?:/releaseblock|/release-block|/release-blocker)\s+(.+)$`)
```
rmohr, line 21: I think you are missing `releaseblocker` in addition to  `releaseblock`.

rmohr:
@davidvossel sorry I created a conflict by merging something else.

external-plugins/release-blocker/server.go
davidvossel: yup, fixed

davidvossel:
@rmohr i took out the ansible deployment/service parts for now. should be ready to go.

PR #226, opened 2019-10-15
[wip] Dependency Mirror

rmohr:
Right now more a draft like anything else, which shows how things can be connected. However, the `upload` binary works and was successfully used on kubevirt.io/kubevirt and kubevirt.io/containerized-data-importer.

To use it do:

```
export GOOGLE_APPLICATION_CREDENTIALS=credential-path-to.json
cd plugins/cmd/uploader/
go build .
./uploader -workspace ~/go/src/kubevirt.io/kubevirt/WORKSPACE # you will see a dry-run
./uploader -workspace ~/go/src/kubevirt.io/kubevirt/WORKSPACE -dry-run=false # will do the upload and modify WORKSPACE 
```

The PR is a mess right now, and only the upload command works, which parsed the WORKSPACE, detected missing mirrored artifacts, uploaded them and dumped the changed WORKSPACE file. Checking the shasum on upload is planned. Right now bazel verifies it when downloading.

The final goal of kubevirt/project-infra#226 is this:

A prow plugin, which puts a hold/non-mirrored-dependency label, if it finds a non-mirrored dependency.
Maintainers can write /upload
The plugin does the upload and pushes a commit to the PR (if that was forbidden by the creator it would post the diff and complain)
The label is removed
We can continue with /lgtm and /approve

PR #444, opened 2020-05-22
Dependency upload tool for bazel

rmohr:
The binary checks the specified WORKSPACE file for unmirrored
dependencies, uploads them to the specified GCS bucket and finally adds
the upload location as a fallback download location to the WORKSPACE
file.

```
export GOOGLE_APPLICATION_CREDENTIALS=credential-path-to.json
bazel run //plugins/cmd/uploader:uploader -- -workspace ~/go/src/kubevirt.io/kubevirt/WORKSPACE # you will see a dry-run
bazel run //plugins/cmd/uploader:uploader -- -workspace ~/go/src/kubevirt.io/kubevirt/WORKSPACE -dry-run=false # will do the upload and modify WORKSPACE
```


rmohr:
/cc @dhiller 

This is a breakout of #226 just with the tool which we already use. The prow plugin which is still very wip in #226 can be added separately.

Would be great if you can have a look.

plugins/cmd/uploader/uploader.go
```diff
@@ -0,0 +1,77 @@
...
+	rawFile, err := ioutil.ReadFile(options.workspacePath)
+	workspace, err := build.ParseWorkspace("workspace", rawFile)
+	if err != nil {
+		log.Fatalf("could not load workspace file: %v", err)
+	}
+	artifacts, err := mirror.GetArtifacts(workspace)
+	if err != nil {
+		log.Fatalf("could not read artifacts: %v", err)
+	}
+	invalid := mirror.FilterArtifactsWithoutMirror(artifacts, regexp.MustCompile(`^https://storage.googleapis.com/.+`))
```
dhiller, line 62: Nit: choose a better name here? I.e. `unmirrorredArtifacts`?

plugins/cmd/uploader/uploader.go
```diff
@@ -0,0 +1,77 @@
...
+
+	for _, artifact := range invalid {
+		newFileUrl := mirror.GenerateFilePath(options.bucket, &artifact)
+		err := mirror.WriteToBucket(options.dryRun, ctx, client, artifact.URLs()[0], options.bucket, artifact.SHA256())
+		if err != nil {
+			log.Fatalf("failed to upload %s to %s", artifact.URLs()[0], newFileUrl)
+		}
```
dhiller, line 69: Shouldn't we better exit here? Or at least collect the errors and return them?

plugins/cmd/uploader/uploader.go
```diff
@@ -0,0 +1,77 @@
...
+		artifact.AppendURL(newFileUrl)
+	}
+
+	err = mirror.WriteWorkspace(options.dryRun, workspace, options.workspacePath)
+	if err != nil {
+		log.Fatalf("could not write workspace file: %v", err)
+	}
```
dhiller, line 76: Same here: handle error somehow? At least store it or return it, so that the job fair with error, otherwise none will notice the failure?

plugins/cmd/uploader/uploader.go
rmohr: I agree. Fatalf exits. So taht should be the behaviour you want.

plugins/cmd/uploader/uploader.go
rmohr: I agree. Fatalf exits. So taht should be the behaviour you want.

rmohr:
@dhiller resolved the merge conflict and ensured that bazel works with the latest changes in the repo.

PR #446, opened 2020-05-27
Add anchors to test reports

rmohr:
In order to point people to specific lines in the reports, add anchors to every line.
anchors


rmohr:
/close

Went by accident in in another PR, but that one should be fine.

rmohr:
went in as part of #450

PR #450, opened 2020-05-28
Go 1.13 for flakefinder

rmohr:
New code we merged into vendor requires go 1.13 already.


rmohr:
/override pull-project-infra-test-flakefinder

Exactly this failing lane will be fixed with the change.

PR #451, opened 2020-05-28
release-blocker plugin improvements

davidvossel:
The release-blocker plugin now

- prevents manually adding/removing release-blocker labels to issues and PRs
- checks to see if branch exists before allowing the label to be added
 - reduces permissions to approvers instead of all org members.
- adds deployment/service ansible yamls
- integrate with bazel


external-plugins/release-blocker/server.go
```diff
@@ -133,25 +149,103 @@ func (s *Server) handleEvent(eventType, eventGUID string, payload []byte) error
...
 		}
 	} else if hasLabel && !add {
-		err := s.GHC.RemoveLabel(org, repo, num, label)
+		err := s.ghc.RemoveLabel(org, repo, num, label)
+		if err != nil {
+			return "", err
+		}
+	}
+
+	return "", nil
```
dhiller, line 198: Just being curious: are you returning an empty string here to avoid any errors regarding `nil`. Is that also considered good style in Go as it is in other languages?

davidvossel: I wanted to separate API errors (which are returned in the second argument) from github comment responses (which are returned in the first arg). Those are being treated independently from one another

PR #452, opened 2020-05-29
Bump flakefinder to the latest version with anchors

rmohr:
In order to point people to specific lines in the reports, add anchors to every line.
anchors

Further back-links will now ease the navigation between the generated tables.


rmohr:
@dhiller I think that the flakefinder jobs belong to project-infra. Something we should change.

dhiller:
> @dhiller I think that the flakefinder jobs belong to project-infra. Something we should change.

not sure of that, as it's information for the analyzed project? Or is that meant rather in terms of who controls this>

PR #458, opened 2020-06-01
hide ipv6 lane

phoracek:
The IPv6 lane on kubevirt is utterly broken with pass rate around 10 %.
We should disable it to reduce the noise and investigate why is the
provider and lane failing.

### DIFF
--- a/github/ci/prow/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
+++ b/github/ci/prow/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
@@ -500,8 +500,8 @@ presubmits:
     annotations:
       fork-per-release: "true"
     always_run: true
-    optional: false
-    skip_report: false
+    optional: true
+    skip_report: true
     decorate: true
     decoration_config:
       timeout: 7h


danielBelenky:
If it’s not reporting and not blocking, let’s run it on demand. It will free some resources.

github/ci/prow/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
```diff
@@ -500,8 +500,8 @@ presubmits:
     annotations:
       fork-per-release: "true"
     always_run: true
```
danielBelenky, line 502: Let’s make it on demand lane since it doesn’t report and doesn’t block the merge anyway.

phoracek: We want to enable it shortly again, collected information about failed runs would be helpful for debugging.

oshoval: The errors are due to prow problems, not because of ipv6 lane specific problems, i would suggest not to make changes to this lane.

phoracek:
Closing since the instability seems to be affecting all lanes.

PR #469, opened 2020-06-09
Add automation to configure remote cluster

danielBelenky:
We have a managed K8s cluster on IBM cloud. We want to utilize it as a remote
Prow jobs executor, enabling complete separation between the Prow instance
itself (and it's secrets) and the place where the jobs actually run on.

Currently, the deploy process is split into 2 parts where the user have to
manually switch K8s context between the clusters. In the future, we might want
to add a pre-defined service account on both of those clusters and store the
token for this SA with our secrets. This will enable fully automated deployment.


danielBelenky:
/hold

Need to run the playbook again to verify latest changes.
The problem is that currently, our managed cluster facing some issues
(not related to us).

rmohr:
@danielBelenky from the doc changes I am not sure if I can simply run the playbook with the latest var file if I don't want to do changes to cluster for prow, or if I have to do something special.

danielBelenky:
/hold cancel
Jobs are starting on the remote cluster as expected now.
Once someone lgtm's this, we are good to go.

PR #472, opened 2020-06-11
Run Travis test script on Prow

danielBelenky:
This is the first step of migrating all our CI workloads to Prow. Our
worker nodes have more juice than what Travis gives us and it will allow
us to make a pipeline out of this job in the future.

### DIFF
--- a/github/ci/prow/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
+++ b/github/ci/prow/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
@@ -609,3 +609,120 @@ presubmits:
           resources:
             requests:
               memory: "29Gi"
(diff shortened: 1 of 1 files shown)


rmohr:
We could also consider running the tasks unnested. We would only need bazel/bazelisk in the builder image.

danielBelenky:
@rmohr I've split the job as you suggested.
About running without dind, I agree that we should go there but maybe we should first migrate the jobs and then remove the dind requirement.

rmohr:
> About running without dind, I agree that we should go there but maybe we should first migrate the jobs and then remove the dind requirement.

works for me.

rmohr:
/lgtm
/approve
/hold

Looks like a good first iteration. Feel free to remove the hold if you think it should go in.

PR #478, opened 2020-06-17
Add new build-officer label

nunnatsa:

### DIFF
--- a/github/ci/prow/files/labels.yaml
+++ b/github/ci/prow/files/labels.yaml
@@ -148,6 +148,10 @@ default:
     - name: topic/virtualization
       color: c5def5
       target: both
+    - name: build-officer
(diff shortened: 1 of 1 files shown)


github/ci/prow/files/labels.yaml
```diff
@@ -148,6 +148,10 @@ default:
     - name: topic/virtualization
       color: c5def5
       target: both
+    - name: build-officer
+      color: 8452c9
+      target: issues
```
dhiller, line 153: nit: maybe `both`? As the build-officer might also create PRs directly?

nunnatsa: In theory yes, but practically, build officer open bug. Teams fixes. So maybe it's the same person, but with different "hat"

PR #487, opened 2020-06-19
k8s-1.18 does not exist on release-0.30 of kubevirt

rmohr:
Removing it since it blocks merges on release-0.30.


jean-edouard:
@rmohr was it intentional to also remove k8s-1.17 here?

rmohr:
> @rmohr was it intentional to also remove k8s-1.17 here?

No, was not intentional. Thanks for discovering.

PR #494, opened 2020-06-22
Add optional k8s-1.18 lanes to CDI.

maya-r:
Soon to replace k8s-1.16 lanes, we'd like to keep up to date with the latest k8s releases.
This is copy-pasted from the k8s-1.16 lanes with the version swapped and changing optional to true.

Would quickly error out until https://github.com/kubevirt/containerized-data-importer/pull/1252 is merged.


maya-r:
cc @awels 
I thought I should first split out the file to a release-v1.19 - v1.14 file, but as long as it's an optional lane, I imagine this is fine

awels:
Right, that is the point of making it optional. Then we can run these on some real PRs and see how they do.

maya-r:
Needs an approval

PR #495, opened 2020-06-22
CLI Release Tool 

davidvossel:
The cli release tool is works with git, the release-blocker plugin and the github API in order to provide release automation.

Example: cut a new branch with a new tag
```
./release-tool --org kubevirt --repo kubevirt --dry-run=false --new-branch=release-0.99 --release-tag=v0.99.0-rc.1
```

This will cut the new branch ```release-0.99``` and cut release ```v0.99.0-rc.1``` from that new branch. Release notes will automatically get generated, verified that release and tag semver align, verified that no blocker issues or PRs exist, and prow configs are forked.


davidvossel:
this should be ready to go now. the basics work

releng/release-tool/release-tool.go
```diff
@@ -0,0 +1,1186 @@
...
+		r.autoDetectData(*autoReleaseCadance, *autoPromoteAfterDays)
+	}
+
+	// If this is a promotion, we need to set the tag to promote
+	if r.promoteRC != "" {
+		if r.tag != "" {
+			log.Fatal("--release-tag and --promote-rc can not be used together. --promote-rc detects the correct tag to make an official release out of")
+		}
+
+		re := regexp.MustCompile(`^v\d*\.\d*.\d*-rc.\d*$`)
```
danielBelenky, line 1130: Maybe it's better to try and instantiate a `semver.Version` and let the semver library to handle the errors/compile.

releng/release-tool/release-tool.go
```diff
@@ -0,0 +1,1186 @@
...
+	log.Printf("\tforce: %t", r.force)
+	log.Printf("\tgithubTokenFile: %s", r.githubTokenPath)
+}
+
+func main() {
+	newBranch := flag.String("new-branch", "", "New branch to cut from master.")
+	releaseTag := flag.String("new-release", "", "New release tag. Must be a valid semver. The branch is automatically detected from the major and minor release")
+	org := flag.String("org", "", "The project org")
+	repo := flag.String("repo", "", "The project repo")
+	dryRun := flag.Bool("dry-run", true, "Should this be a dry run")
```
danielBelenky, line 1033: I really like this option! I used this tool today to do the release and it gave me confidence to play around with the tool.

releng/release-tool/release-tool.go
```diff
@@ -0,0 +1,1186 @@
...
+			return fmt.Errorf("Unable to find promote-rc tag [%s]", r.promoteRC)
+		} else if invalid {
+			return fmt.Errorf("RC [%s] is ineligible to be promoted due to blockers.", r.promoteRC)
+		}
+	} else {
+		// if this is not a promotion, ensure the release is either an RC or a patch release.
+		re := regexp.MustCompile(`^v\d*\.\d*.\d*-rc.\d*$`)
+		match := re.FindString(r.tag)
+		if match == "" && tagSemver.Patch() == 0 && !r.force {
+			return fmt.Errorf("The tag [%s] must be promoted from a release candidate since it is the first release of a patch series.", r.tag)
```
danielBelenky, line 939: Maybe it makes sense to have a defined exit codes in those cases? it will allow us to run this tool automatically and handle different errors.

releng/release-tool/release-tool.go
```diff
@@ -0,0 +1,1186 @@
...
+			}
+		}
+		if !found {
+			return fmt.Errorf("Unable to find promote-rc tag [%s]", r.promoteRC)
+		} else if invalid {
+			return fmt.Errorf("RC [%s] is ineligible to be promoted due to blockers.", r.promoteRC)
+		}
+	} else {
+		// if this is not a promotion, ensure the release is either an RC or a patch release.
+		re := regexp.MustCompile(`^v\d*\.\d*.\d*-rc.\d*$`)
```
danielBelenky, line 936: Maybe use the semver library here as well?

releng/release-tool/release-tool.go
```diff
@@ -0,0 +1,1170 @@
...
+
+	// github cached results
+	allReleases      []*github.RepositoryRelease
+	allBranches      []*github.Branch
+	blockerListCache map[string]*blockerListCacheEntry
+
+	now time.Time
+}
+
+func gitCommandStdin(stdin string, arg ...string) error {
```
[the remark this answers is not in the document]
davidvossel: nope, i'll remove it. i used it at one point, but removed whatever behavior it was used for.

releng/release-tool/release-tool.go
```diff
@@ -0,0 +1,1186 @@
...
+		githubTokenPath: *githubTokenFile,
+		dryRun:          *dryRun,
+		force:           *force,
+
+		blockerListCache: make(map[string]*blockerListCacheEntry),
+
+		now: time.Now(),
+	}
+
+	_, err = gitCommand("config", "--global", "user.name", r.gitUser)
```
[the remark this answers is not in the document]
davidvossel: ah, interesting. so the tool actually is only meant to be run in a container but I can look into a per repo git config just in case someone does run it locally

releng/release-tool/release-tool.go
```diff
@@ -0,0 +1,1186 @@
...
+	}
+
+	if *autoRelease {
+		r.autoDetectData(*autoReleaseCadance, *autoPromoteAfterDays)
+	}
+
+	// If this is a promotion, we need to set the tag to promote
+	if r.promoteRC != "" {
+		if r.tag != "" {
+			log.Fatal("--release-tag and --promote-rc can not be used together. --promote-rc detects the correct tag to make an official release out of")
```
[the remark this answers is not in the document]
davidvossel: yup, good eye :+1:

releng/release-tool/release-tool.go
davidvossel: yeah, i decided to leave dry-run on by default just in case, so it forces people to disable dry-run to do anything

releng/release-tool/release-tool.go
davidvossel: the automation parts are meant to handle all this already. If we hit an error like this, it would be because a user tried to do something unexpected.

releng/release-tool/release-tool.go
davidvossel: i'll look into it

davidvossel:
i'm going to take another pass at addressing a few comments and improving testing a bit.

releng/release-tool/release-tool.go
davidvossel: this regex is actually to enforce that we only promote an rc which is structured the way we expect it `<major>.<minor>.<patch>-rc.<rc version>`. It has to match this regardless if it is a valid semver or not.

releng/release-tool/release-tool.go
davidvossel: semver doesn't help us here because we're specifically interested in the `-rc.<rc number` postfix which semver doesn't enforce. That's why a regex is used.

davidvossel:
feedback addressed, and I added some more unit tests

danielBelenky:
/approve

I don’t see a reason to keep it open any longer. Thanks David 😄

PR #504, opened 2020-06-23
Add testing for CDI release-v1.13 branch.

maya-r:
This is based on an older version of the test file.
This is done as the KUBEVIRT_PROVIDER lanes do not exist in those names on the release-v1.13 branch.


maya-r:
/hold
made a mistake about which lanes are problematic (1.18 is fine, 1.13 is the one that had issues with the rename). thanks awels.

maya-r:
/hold cancel

Now it only addresses release-v1.13.

PR #527, opened 2020-06-28
Update pull-kubevirt-e2e-k8s-1.18 to be mandatory

oshoval:
Should be stable enough, and we need it as a guard.

### DIFF
--- a/github/ci/prow/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
+++ b/github/ci/prow/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
@@ -103,7 +103,7 @@ presubmits:
     annotations:
       fork-per-release: "true"
     always_run: true
-    optional: true
+    optional: false
     skip_report: false
     decorate: true
     decoration_config:


rmohr:
/hold

We need to resolve the issue on one test until we can enable it: https://storage.googleapis.com/kubevirt-prow/reports/flakefinder/kubevirt/kubevirt/flakefinder-2020-06-28-168h.html#row1. 1.18

rmohr:
Here the issue: https://github.com/kubevirt/kubevirt/issues/3610

oshoval:
> /hold
> 
> We need to resolve the issue on one test until we can enable it: https://storage.googleapis.com/kubevirt-prow/reports/flakefinder/kubevirt/kubevirt/flakefinder-2020-06-28-168h.html#row1. 1.18

Its as bad on 1.17 as its bad on 1.18, but up to you, thanks.
I understand it might stall some PRs when fails, but rest of the tests are pretty stable, and the above reason as well,
just we need to make sure, manually, 1.18 doesn't start a regression, since its not a gating atm.

rmohr:
/hold cancel
/approve

The flaky test is resolved. @oshoval it was necessary to block it. I agree that 1.17 and 1.18 showed the symptoms. Making both mandatory would have blocked merges since the probabilities of failures are AND connected.

PR #531, opened 2020-06-29
Re-enable k8s-1.17 on the 0.30 lane

jean-edouard:


rmohr:
Something seems to be wrong with the yaml

jean-edouard:
> Something seems to be wrong with the yaml

Sorry about that, should be fixed now.

Issue #539, opened 2020-07-03
cnao flakefinder jobvs are missing org and repo configs


dhiller:
Jobs for flakefinder report generation are missing [configuration for `org` and `repo`](https://github.com/kubevirt/project-infra/blob/master/robots/flakefinder/main.go#L52).

Example for missing configuration:
https://github.com/kubevirt/project-infra/blob/412094f96c3c034dd86e810fda1aa5942a0ec558/github/ci/prow/files/jobs/kubevirt/cluster-network-addons-operator/cluster-network-addons-operator-periodics.yaml#L49

Example for correct configuration: https://github.com/kubevirt/project-infra/blob/bef173d869b8e54c5ca3d76b49d94e2c3605da81/github/ci/prow/files/jobs/nmstate/kubernetes-nmstate/kubernetes-nmstate-periodics.yaml#L55

Relates to: cluster-network-addons-operator#447

/triage build-officer

PR #540, opened 2020-07-05
Specify repo, so we don't default to "kubevirt"

maya-r:
This is inspired by looking at [flakefinder reports for CNAO](https://storage.googleapis.com/kubevirt-prow/reports/flakefinder/kubevirt/cluster-network-addons-operator/index.html)

They seem to point at kubevirt/kubevirt rather than kubevirt/cluster-network-addons-operator, I was blindly searching for what could've caused it.


maya-r:
cc @dhiller in case I'm wrong about the cause for the issue :)

RamLavi:
@maya-r I came to the same conclusion as you as to the cause of the issue.

dhiller:
@maya-r thanks for the PR to fix this, exactly what is required. ICMY I created the issue here for that last week already: kubevirt/project-infra#539 Feel free to close the issue when it is solved. /cc @RamLavi

PR #544, opened 2020-07-07
Update test lanes for NMO

slintes:
Removed k8s 1.13, does not exist anymore
Updated k8s 1.17 to 1.18
Updated ocp 4.3 to 4.4 (4.3 does not get healthy)


beekhof:
/approve
/lgtm 

Should we keep some older versions around so we don’t accidentally land incompatibilities?

PR #554, opened 2020-07-16
Prow jobs rehearsal plugin

danielBelenky:
The purpose of this plugin is to rehearse modified Prow jobs from a PR.
There are two working modes: run on every PR or run only when someone
comments /rehearse.

This is eventually a web-server that accepts github webhooks for modified pull-requests
and comments, fetches the changes, calculates the modified prow jobs and applies them
to the cluster.


external-plugins/rehearse/events/events.go
```diff
@@ -0,0 +1,397 @@
...
+	"strings"
+
+	v1 "k8s.io/test-infra/prow/client/clientset/versioned/typed/prowjobs/v1"
+
+	"k8s.io/test-infra/prow/pjutil"
+
+	"github.com/sirupsen/logrus"
+	prowapi "k8s.io/test-infra/prow/apis/prowjobs/v1"
+	"k8s.io/test-infra/prow/config"
+	git2 "k8s.io/test-infra/prow/git/v2"
```
dhiller, line 20: nit: `gitv2`?

external-plugins/rehearse/events/events.go
```diff
@@ -0,0 +1,397 @@
...
+		if err := json.Unmarshal(eventPayload, &event); err != nil {
+			eventLog.WithError(err).Error("Could not unmarshal event.")
+		}
+		s.onUpdateEvent(eventLog, &event)
+	default:
+		eventLog.Info("Skipping event.")
+	}
+}
+
+func validateDryRunEvent(w http.ResponseWriter, r *http.Request, t func() []byte) (string, string, []byte, bool, int) {
```
dhiller, line 93: nit: use named return values, then you just can always use `return`, this would make it clearer that each of these returns is just a guard clause for an invalid event.

external-plugins/rehearse/events/events.go
```diff
@@ -0,0 +1,397 @@
...
+		dryRun:           dryRun,
+		tokenGenerator:   tokenGenerator,
+		prowConfigPath:   prowConfigPath,
+		jobsConfigBase:   jobsConfigBase,
+	}
+}
+
+type eventValidatingFunc func(http.ResponseWriter, *http.Request, func() []byte) (string, string, []byte, bool, int)
+
+func (s *GitHubEventsServer) ServeHTTP(w http.ResponseWriter, r *http.Request) {
```
dhiller, line 59: nit: regarding readability: I'm not a big fan of one character function arguments, as when you scroll the function declaration out of sight, you easily tend to forget what the s, w, r stood for.

external-plugins/rehearse/events/events.go
```diff
@@ -0,0 +1,397 @@
...
+
+func (s *GitHubEventsServer) generateProwJobs(
+	headConfigs, baseConfigs map[string]*config.Config, event *github.PullRequestEvent) []prowapi.ProwJob {
+	var jobs []prowapi.ProwJob
+
+	for path, headConfig := range headConfigs {
+		baseConfig, _ := baseConfigs[path]
+		jobs = append(jobs, s.generatePresubmits(headConfig, baseConfig, event)...)
+		jobs = append(jobs, s.generatePeriodics(headConfig, baseConfig, event)...)
+	}
```
dhiller, line 180: Post submits? Not yet I guess?

external-plugins/rehearse/rehearse_test.go
```diff
@@ -0,0 +1,131 @@
...
+			"")
+	})
+
+	AfterEach(func() {
+		if gitClientFactory != nil {
+			gitClientFactory.Clean()
+		}
+	})
+
+	FIt("With a running rehearsals server", func() {
```
dhiller, line 52: Remove focus here?

danielBelenky: Thanks :)

external-plugins/rehearse/events/events.go
danielBelenky: I find named return values to be very confusing. You need to remember what exactly you return and where you calculated that value....

external-plugins/rehearse/events/events.go
danielBelenky: +1, I totally agree with that. Fixed.

external-plugins/rehearse/events/events.go
danielBelenky: Not yet. I actually also removed the support for periodics for now, but we'll add it back soon.

external-plugins/rehearse/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM golang:1.14.6-alpine3.12 as builder
+
+RUN mkdir /build
+COPY plugin/ /build
+RUN cd /build/ && go build -v .
+
+FROM alpine:3.12.0
```
qinqon, line 7: We can even use scratch

external-plugins/rehearse/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM golang:1.14.6-alpine3.12 as builder
+
+RUN mkdir /build
+COPY plugin/ /build
+RUN cd /build/ && go build -v .
```
qinqon, line 5: Deactivate CGO so we make it a static binary.

external-plugins/rehearse/plugin/main.go
```diff
@@ -0,0 +1,189 @@
...
+	"os"
+	"time"
+
+	"k8s.io/test-infra/prow/git/v2"
+
+	"github.com/sirupsen/logrus"
+	"k8s.io/test-infra/pkg/flagutil"
+	"k8s.io/test-infra/prow/config/secret"
+	prowflagutil "k8s.io/test-infra/prow/flagutil"
+	"kubevirt.io/project-infra/external-plugins/rehearse/events_server"
```
qinqon, line 23: Is good practice to re-organize imports separating groups by newline
- first golang
- second third parties grouped by project
- third app

external-plugins/rehearse/plugin/events_server/events_server_test.go
```diff
@@ -0,0 +1,171 @@
+package events_server
```
qinqon, line 1: Golang packages does not containe underscroes they just concatenate `envetsserver`

external-plugins/rehearse/plugin/main.go
```diff
@@ -0,0 +1,189 @@
...
+	prowConfigPath string
+	jobsConfigBase string
+	kubeconfig     string
+	jobsNs         string
+	alwaysRun      bool
+	cacheDir       string
+	github         prowflagutil.GitHubOptions
+}
+
+func (o *options) validate() {
```
qinqon, line 40: What about using third parties for arguments handling that implementes the mandatory attribute ?

external-plugins/rehearse/plugin/main.go
danielBelenky: You are right. I let the IDE to organize this but I think that something was misconfigured there.

external-plugins/rehearse/Dockerfile
stu-gott: True, but will an admin ever need to login to debug? Unless there's a truly compelling reason to strip this down to nothing, it would be a good idea to leave at least some way to execute a shell inside the pod's environment.

qinqon: maybe we don't even need a container to run this, this is pure golang we can run it with `go run` and it will compile it too like scripting golang.

danielBelenky: @qinqon it's a web-server that is supposed to run on K8s. We need a container.
I can't run it in a scratch container because I need git to be installed and as Stu mentioned already,
we want the ability to SSH into that container to debug.

external-plugins/rehearse/plugin/main.go
danielBelenky: I inherit some flags from Prow's module and I need flag to register those. Otherwise, I would have used spf13/cobra.

external-plugins/rehearse/plugin/main.go
```diff
@@ -0,0 +1,197 @@
...
+	logger.Println("Rehearse server was gracefully shut down")
+}
+
+func mustSucceed(err error, message string) {
+	if err != nil {
+		logrus.WithError(err).Fatal(message)
+	}
+}
+
+func startSecretsAgnet(ghTokenPath, hmacTokenPath string) *secret.Agent {
```
qinqon, line 183: `s/Agnet/Agent/g`

external-plugins/rehearse/plugin/main.go
qinqon: Fair enough

external-plugins/rehearse/plugin/main.go
```diff
@@ -0,0 +1,197 @@
...
+		"/etc/webhook/hmac",
+		"Path to the file containing the GitHub HMAC secret.")
+	fs.StringVar(&o.kubeconfig,
+		"kubeconfig",
+		"",
+		"Path to kubeconfig. If empty, will try to use K8s defaults.")
+	fs.StringVar(&o.prowConfigPath,
+		"prow-config-path",
+		"",
+		"Path to Prow configuration (required).")
```
stu-gott, line 92: Is it common to make flags required? Isn't that usually what positional arguments are for?

external-plugins/rehearse/plugin/server/eventsserver.go
```diff
@@ -0,0 +1,43 @@
...
+		return
+	}
+
+	s.eventsChan <- &handler.GitHubEvent{
+		Type:    eventType,
+		GUID:    eventGUID,
+		Payload: eventPayload,
+	}
+	writer.WriteHeader(http.StatusOK)
+	writer.Write([]byte("Event received. Have a nice day."))
```
stu-gott, line 42: Thank you! You should have a nice day too!

external-plugins/rehearse/plugin/main.go
danielBelenky: You are correct, and I absolutely agree with you. Unfortunately, that's what this library supports and I prefer to use it because the other Prow components use it as well and we expose some default Prow flags through it.

external-plugins/rehearse/plugin/handler/handler.go
```diff
@@ -0,0 +1,504 @@
...
+				presubmitKey, headPresubmit.Name)
+		}
+
+		job.Spec.ExtraRefs = addExtraRefs(job.Spec.ExtraRefs, org, repo, pr.Base.Ref)
+		jobs = append(jobs, job)
+	}
+	return jobs
+}
+
+func (h *gitHubEventsHandler) generatePeriodics(headConfig, baseConfig *config.Config) []prowapi.ProwJob {
```
rmohr, line 321: This is unused right now?

external-plugins/rehearse/plugin/handler/handler.go
```diff
@@ -0,0 +1,504 @@
...
+	_, err = tmpfile.Write(content)
+	if err != nil {
+		log.WithError(err).Errorf("Could not write data to file: %s", tmpfile.Name())
+		return "", err
+	}
+	tmpfile.Sync()
+	return tmpfile.Name(), nil
+}
+
+func addExtraRefs(refs []prowapi.Refs, org, repo, ref string) []prowapi.Refs {
```
rmohr, line 466: How about something like `addTargetRepoRef`?

external-plugins/rehearse/plugin/handler/handler.go
```diff
@@ -0,0 +1,504 @@
...
+	reflect.Copy(r, reflect.ValueOf(refs))
+
+	copiedRefs := r.Interface().([]prowapi.Refs)
+
+	if copiedRefs != nil {
+		// We need to check if there is an extra ref that is set to be the WorkDir.
+		// If there is no such ref, then we set the repo for which this presubmit is
+		// configured for as the work dir. Otherwise, we don't want to override the
+		// users configuration.
+		copiedRefs = append(
```
rmohr, line 478: You can keep just this and remove the surrounding if/else. You can iterate on `nil` arrays and `append` to nil arrays.
See this example: https://play.golang.org/p/gLkZP3vocO1

external-plugins/rehearse/plugin/handler/handler.go
```diff
@@ -0,0 +1,504 @@
...
+		log.WithError(err).Errorf("Could not write data to file: %s", tmpfile.Name())
+		return "", err
+	}
+	tmpfile.Sync()
+	return tmpfile.Name(), nil
+}
+
+func addExtraRefs(refs []prowapi.Refs, org, repo, ref string) []prowapi.Refs {
+	// We don't want to modify the refs on the original object, so we make our copy of it
+	r := reflect.ValueOf(make([]prowapi.Refs, len(refs)))
```
rmohr, line 468: Why do you not just add it directly? If you don't want to add it directly, consider something like this:

```go
func generateTargetRepoRef (job v1.ProwJob, org, repo, ref string) prowapi.Refs {
    return prowapi.Refs{
prowapi.Refs{
				Repo:    repo,
				Org:     org,
				WorkDir: !workdirAlreadyDefined(refs),
				BaseRef: ref,
}
}
```

and then

```
job.Refs  = append(job.Refs, generateTargetRepoRef(job, org, repo, ref))
```

external-plugins/rehearse/plugin/handler/handler.go
```diff
@@ -0,0 +1,504 @@
...
+
+	for presubmitKey, headPresubmit := range headPresubmits {
+		basePresubmit, exists := basePresubmits[presubmitKey]
+
+		if exists && reflect.DeepEqual(basePresubmit.Spec, headPresubmit.Spec) {
+			continue
+		}
+		// This is a new or modified presubmit
+		// TODO: check the repo configuration on the generated Prow job - will it report the correct place?
+		log.Infof("Detected modified presubmit: %s.", headPresubmit.Name)
```
rmohr, line 303: `Detected modified or new presubmit:`

external-plugins/rehearse/plugin/handler/handler.go
```diff
@@ -0,0 +1,504 @@
...
+
+	prowjobs := h.generateProwJobs(headConfigs, baseConfigs, pr, eventGUID)
+	log.Infof("Will create %d jobs", len(prowjobs))
+	for _, job := range prowjobs {
+		if job.Labels == nil {
+			job.Labels = make(map[string]string)
+		}
+		job.Labels["rehearsal"] = "true"
+		job.Labels["rehearsal-for-pull-request"] = strconv.Itoa(pr.Number)
+		job.Spec.Job = strings.Join([]string{"rehearsal", job.Spec.Job}, "-")
```
rmohr, line 263: You also need to set `Spec.Context` to `strings.Join([]string{"rehearsal", job.Spec.Job}, "-")` to report it with a `rehearsal` prefix on the github PR. This way one can clearly see that this is a rehearsal job on the PR.

external-plugins/rehearse/plugin/handler/handler.go
```diff
@@ -0,0 +1,504 @@
...
+	// the repo is checked out with any refspec. Instead, we use git cat-file to read the files that we need to the
+	// memory and write them to a temp file. We need the temp file because the current version of Prow's config module
+	// can't load the configs from the memory.
+	prowConfigBytes, ret := catFile(log, git.Directory(), h.prowConfigPath, ref)
+	if ret != 0 && ret != 128 {
+		log.WithError(err).Errorf("Could not load Prow config [%s] at ref: %s", h.prowConfigPath, ref)
+		return nil, err
+	}
+	log.Infoln("File:", string(prowConfigBytes))
+	prowConfigTmp, err := writeTempFile(log, tmpdir, prowConfigBytes)
```
rmohr, line 365: Will this be clean up too?

external-plugins/rehearse/plugin/handler/handler.go
danielBelenky: Yes, I removed it.

external-plugins/rehearse/plugin/handler/handler.go
danielBelenky: Yes, when the directory will be removed, the files in it will be removed as well.
See https://github.com/kubevirt/project-infra/pull/554/files#diff-1cf555f4744be30bc95023fc222bd9d3R354

external-plugins/rehearse/plugin/handler/handler.go
danielBelenky: I always tend to forget that iteration on nil arrays works in Go :laughing: 
Thanks.

external-plugins/rehearse/plugin/main.go
```diff
@@ -0,0 +1,216 @@
...
+	stopChan <- true
+	logger.Println("Rehearse server was gracefully shut down")
+}
+
+func helpProvider(_ []config.OrgRepo) (*pluginhelp.PluginHelp, error) {
+	pluginHelp := &pluginhelp.PluginHelp{
+		Description: `The rehearse plugin is used run modified Prow jobs from a PR to provide pre-merge feedback.`,
+	}
+	pluginHelp.AddCommand(pluginhelp.Command{
+		Usage:       "Send a PR or type /rehearse on the PRs page",
```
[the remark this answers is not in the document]
danielBelenky: Right, I've added just the permitted label for now because I think that it keeps things simpler.
You either permit or you don't. And if we will want rehearsals for jobs, we will add them as we modify them... I think it should work just fine.

external-plugins/rehearse/plugin/server/eventsserver.go
```diff
@@ -0,0 +1,43 @@
...
+func (s *GitHubEventsServer) ServeHTTP(writer http.ResponseWriter, request *http.Request) {
+
+	eventType, eventGUID, eventPayload, eventOk, _ := github.ValidateWebhook(writer, request, s.tokenGenerator)
+
+	if !eventOk {
+		writer.WriteHeader(http.StatusBadRequest)
+		return
+	}
+
+	s.eventsChan <- &handler.GitHubEvent{
```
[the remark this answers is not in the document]
danielBelenky: Your comment caused me to re-think that I did and actually question why I even used a channel here and I don't remember that I had a good reason for it except for separation of concerns between the events handler and the events server itself, but I assume that by making the handler a member of the events server instead of the channel still solves this problem and even simplifies the testing.

external-plugins/rehearse/plugin/handler/handler.go
```diff
@@ -0,0 +1,463 @@
...
+	eventLog := log.WithField("event-guid", incomingEvent.GUID)
+	switch incomingEvent.Type {
+	case "pull_request":
+		eventLog.Infoln("Handling pull request event")
+		var event github.PullRequestEvent
+		if err := json.Unmarshal(incomingEvent.Payload, &event); err != nil {
+			eventLog.WithError(err).Error("Could not unmarshal event.")
+			return
+		}
+		h.handlePullRequestUpdateEvent(eventLog, &event)
```
qinqon, line 89: What about returning errors here and composing them with `github.com/pkg/errors` using `errors.Wrap` and friends ? we will do the logrus.WithError just here.

external-plugins/rehearse/plugin/handler/handler.go
```diff
@@ -0,0 +1,463 @@
...
+	case "issue_comment":
+		logrus.Infoln("Handling issue comment event")
+		var event github.IssueCommentEvent
+		if err := json.Unmarshal(incomingEvent.Payload, &event); err != nil {
+			log.WithError(err).Error("Could not unmarshal event.")
+			return
+		}
+		h.handleIssueComment(eventLog, &event)
```
qinqon, line 97: ditto

external-plugins/rehearse/plugin/handler/handler.go
danielBelenky: Who should handle the error that is returned?

external-plugins/rehearse/plugin/server/eventsserver.go
```diff
@@ -0,0 +1,37 @@
...
+		writer.WriteHeader(http.StatusBadRequest)
+		return
+	}
+
+	event := &handler.GitHubEvent{
+		Type:    eventType,
+		GUID:    eventGUID,
+		Payload: eventPayload,
+	}
+	go s.eventsHandler.Handle(event)
```
qinqon, line 35: If we use error propagating to facilitate unit testing we can log the error inside the goroutine so it's just at one place

```golang
go func() {
   err := s.eventsHandler.Handle(event)
   log.WithError(err)....
}
```

external-plugins/rehearse/plugin/handler/handler.go
qinqon: I suppose the border is here https://github.com/kubevirt/project-infra/pull/554/files#r464933931

qinqon:
/lgtm

Error handling feels a little weird but looks like it's prow's components way.

external-plugins/rehearse/plugin/handler/handler.go
```diff
@@ -0,0 +1,463 @@
...
+
+	baseConfigs, err := h.loadConfigsAtRef(changedJobConfigs, git, pr.Base.SHA)
+	if err != nil {
+		log.WithError(err).Errorf(
+			"Could not load job configs from base ref: %s", pr.Base.SHA)
+	}
+	log.Infoln("Base configs:", baseConfigs)
+
+	prowjobs := h.generateProwJobs(headConfigs, baseConfigs, pr, eventGUID)
+	log.Infof("Will create %d jobs", len(prowjobs))
```
rmohr, line 248: I would suggest to add a check here if a rehearsal is already going on and filter out jobs which are not yet finished.

external-plugins/rehearse/plugin/handler/handler.go
```diff
@@ -0,0 +1,463 @@
...
+		job.Spec.Context = rehearsalName
+		_, err := h.prowClient.Create(&job)
+		if err != nil {
+			log.WithError(err).Errorf("Failed to create prow job: %s", job.Spec.Job)
+		}
+		log.Infof("Created a rehearse job: %s", job.Name)
+	}
+}
+
+const rehearsalPermittedLabel = "rehearsal.allowed"
```
rmohr, line 272: Can you make this an annotation? Labels are not really a good fit for something like this.

rmohr: For presubmits I think we only have jobs in kubevirtci which would require blacklisting. I wonder if we should start with a denylist instead of an allowlist.

danielBelenky: We currently don't rehearse postsubmits and periodics. I think that it should be ok to start with explicit "allowed" annotation. Otherwise, someone can add a job by mistake that shouldn't be rehearsed and it will run.

Let's start with that and if it will become an issue, I'll update the code?

rmohr: Yes, since there are no postsubmits and periodics we basically add to every presubmit an allowed annotation. The only presubmits where we would maybe deny it are the push jobs from kubevirtci. It sounds better to me to explicitly deny the push jobs (if we we even want that. They "release" something but it is not referenced anywhere, so it is not dangerous).

rmohr:
/approve

Discussed offline that deploying the plugin properly will be tackled in a dedicated PR.

PR #567, opened 2020-08-03
Release tool improvments

davidvossel:
- Adds `#` character to PRs in release notes so they appear as links. 

- Fixes issue with branch cache. We need to clear
  branch cache and re-generate in the event a new branch
  is created

- Fixes issue with branching prow configs. Only attempt
  to pull the latest branch code for the prow config if
  the branch already exists.

### DIFF
--- a/releng/release-tool/release-tool.go
+++ b/releng/release-tool/release-tool.go
@@ -598,7 +598,7 @@ func (r *releaseData) verifyPromoteRC() error {
 	re := regexp.MustCompile(`^v\d*\.\d*.\d*-rc.\d*$`)
 	match := re.FindString(r.promoteRC)
 	if match == "" {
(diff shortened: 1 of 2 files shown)


releng/release-tool/release-tool.go
```diff
@@ -386,17 +395,20 @@ func (r *releaseData) forkProwJobs() error {
 	gitbranch := fmt.Sprintf("%s_%s_%s_configs", r.org, r.repo, r.newBranch)
 
 	_, err := gitCommand("-C", r.infraDir, "checkout", "-b", gitbranch)
-	// if err, then branch probably already exists
+
+	// if err, then branch likely already exists
```
rmohr, line 399: Could this potentially lead to a situation where we take the branch from a not-anymore correct commit? Would it make sense to use `-B` instead. Basically create-if-necessary and set to the current commit?

davidvossel: yep, fixed

davidvossel:
@rmohr this should be good for another look

PR #568, opened 2020-08-03
Fix the bot command list link

jean-edouard:
The prow version we use has a bug where
the link for the list of bot commands is
broken if not explicitely set.
Setting it to its "default" value.

Fixed upstream by:
https://github.com/kubernetes/test-infra/pull/18097


stu-gott:
> The full list of commands accepted by this bot can be found [here](?repo=kubevirt%2Fproject-infra).
> Needs approval from an approver in each of these files:
> 

ironic....

PR #570, opened 2020-08-05
Add an optional lane with a newer version of base image.

tomob:
The CDI tests now require libdevmapper.so to run. The base image all the
lanes are running from does not include it, but a latest version does.
This PR adds an optional lane to verify that a/ the tests run
correctly and b/ the CI works with the updated image.

Signed-off-by: Tomasz Baranski <tbaransk@redhat.com>

### DIFF
--- a/github/ci/prow/files/jobs/kubevirt/containerized-data-importer/containerized-data-importer-presubmits.yaml
+++ b/github/ci/prow/files/jobs/kubevirt/containerized-data-importer/containerized-data-importer-presubmits.yaml
@@ -192,4 +192,36 @@ presubmits:
           resources:
             requests:
               memory: "29Gi"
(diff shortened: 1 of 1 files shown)


github/ci/prow/files/jobs/kubevirt/containerized-data-importer/containerized-data-importer-presubmits.yaml
```diff
@@ -192,4 +192,36 @@ presubmits:
           resources:
             requests:
               memory: "29Gi"
+  - name: pull-containerized-data-importer-e2e-k8s-1.18-hpp-test-image
+    skip_branches:
+      - release-v\d+\.\d+
+    annotations:
+      fork-per-release: "true"
+    always_run: true
```
mhenriks, line 200: this should probably be false

PR #575, opened 2020-08-10
If things look bad in CI, the one hour retest is an issue

rmohr:
Let's only retrigger jobs all three hours for now and see if it is a
good compromise.


rmohr:
/cc @danielBelenky 

A small thing which may help a little in overload scenarios.

PR #579, opened 2020-08-11
Schedule our caches on a dedicated node.

rmohr:
The label is set on the caching node only and the caching node has the
specified taint applied too.


rmohr:
/hold

This will probably also try to schedule the caches on our second cluster to a dedicated node, which does not exist.

rmohr:
/hold cancel

As a simple solution all nodes on the second cluster have the label for now.

PR #580, opened 2020-08-12
labels: Rename topic/network to sig/network

EdDev:
The label plugin supports labeling using a command only specific types
and "topic" is not one of them.

"sig" is a supported type and is equivalent in meaning to the "topic"
one.
With this change, one can label a PR or issue using the
`/sig network` comment.

### DIFF
--- a/github/ci/prow/files/labels.yaml
+++ b/github/ci/prow/files/labels.yaml
@@ -145,18 +145,30 @@ default:
       target: both
       prowPlugin: lifecycle
       addedBy: prow
-    - name: topic/network
+    - name: sig/network
       color: c5def5
       target: both
-    - name: topic/security
+      previously:
+        - name: topic/network
+          color: c5def5
+    - name: sig/security
       color: c5def5
       target: both
-    - name: topic/storage
+      previously:
+        - name: topic/security
+          color: c5def5
+    - name: sig/storage
       color: c5def5
       target: both
-    - name: topic/virtualization
+      previously:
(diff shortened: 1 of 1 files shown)


phoracek:
/lgtm

But I'd like to hear from somebody who understands Prow config better, to make sure we don't break anything.

danielBelenky (changes requested):
For the sake of consistency, let's either change the label for everyone or keep it as `topic/network`.
I personally don't have any preference on that... just want things to be consistent.

EdDev:
> For the sake of consistency, let's either change the label for everyone or keep it as `topic/network`.
> I personally don't have any preference on that... just want things to be consistent.

I think each team can only control their own labels and I cannot enforce this on all. I especially do not think I can remove the old ones for the others.
If it is just adding, I am fine with adding the others as well.

danielBelenky:
> > For the sake of consistency, let's either change the label for everyone or keep it as `topic/network`.
> > I personally don't have any preference on that... just want things to be consistent.
> 
> I think each team can only control their own labels and I cannot enforce this on all. I especially do not think I can remove the old ones for the others.
> If it is just adding, I am fine with adding the others as well.

I'm all for aligning our verticals with how K8s defines them (SIGs).
Let's raise it on the mailing list. I just don't want to have confusing labels on PRs and issues.

rmohr:
Labels are synchronized twice a day. It will take a few hours until the changes are effective.

PR #587, opened 2020-08-20
add task for building prow_builder_arm64 image

zhlhahaha:
The prow_builder_arm64 image containes necessery prow
utility cmd binaries for ARM64. Then we can use it to
build prow utility images.


zhlhahaha:
Hi, @rmohr @danielBelenky 
I have verified image build part of this PR in a periodics prow jobs in my local multi-arch prow cluster without any tags(e.g. preset-dind-enabled, preset-kubevirtci-docker-credential, preset-docker-mirror) and it works fine.
I have done some research on how the tags works, I think it should be fine.
Here are two items that have not been checked.
1. whether the tags work? 
-- The tag, preset-docker-mirror, use configmap "docker-daemon-mirror-config". And the tag, preset-kubevirtci-docker-credential, use secret "kubevirtci-docker-credential". Do we need to set the configmap and the secret in arm-cluster? 

2. If the presubmit job and postsubmit job can be triggered successfully?
-- I do not have env to trigger presubmit job and postsubmit job, so it need to be verified.

As all new prow jobs are assigned to arm-cluster, it should be safe to current prow cluster.

images/dind-env-build-arm64.sh
```diff
@@ -0,0 +1,51 @@
...
+# Used to set Docker in docker environment
+
+#
+# BEGIN: DOCKER IN DOCKER SETUP
+#
+
+# Install Docker deps, some of these are already installed in the image but
+# that's fine since they won't re-install and we can reuse the code below
+# for another image someday.
+apt-get update && apt-get install -y --no-install-recommends \
```
rmohr, line 12: @zhlhahaha I have to extract our builder image from kubevirt/kubevirt and make it multi-arch-fit. I think we can do something nice with `qemu-static-user`. Let me come back to you in a few hours.

zhlhahaha: > @zhlhahaha I have to extract our builder image from kubevirt/kubevirt and make it multi-arch-fit. I think we can do something nice with `qemu-static-user`. Let me come back to you in a few hours.

Hi Roman, I understand your idea. But for prow utility images, do we need to build them for multi-arch?

images/kubevirt-infra-prow-arm64/builder/nodejs.patch
```diff
@@ -0,0 +1,36 @@
+diff --git a/WORKSPACE b/WORKSPACE
+index d500c17519..fc11d57dde 100644
+--- a/WORKSPACE
++++ b/WORKSPACE
+@@ -62,7 +62,12 @@ load("@io_bazel_rules_k8s//k8s:k8s.bzl", "k8s_repositories")
+
+ k8s_repositories()
+
+-load("@build_bazel_rules_nodejs//:defs.bzl", "yarn_install")
++load("@build_bazel_rules_nodejs//:defs.bzl", "node_repositories", "yarn_install")
```
rmohr, line 10: The changes here look minimal, any chance that we can just push them right now into test-infra? They may not care right now about multi-arch builds, but they should not mind merging a fix. What exactly are you fixing?

images/dind-env-build-arm64.sh
rmohr: I would at least have the chance to build them for arm and ppc64 without having to care about the cluster-architecture. I am a little bit worried about  too much friction here, which may make it hard for us to update things in the future. Multi-arch image is basically just the little extra-plus which eases maintenance a little bit more.

rmohr: At least it looks to me like your spcripts expects this to be run on an actual arm machine.

Did you think about cross-compiling it with bazel? Something like

```
bazel build --platforms=@io_bazel_rules_go//go/toolchain:linux_arm64_cgo ...
```

should already do the job.

images/kubevirt-infra-prow-arm64/builder/nodejs.patch
zhlhahaha: > The changes here look minimal, any chance that we can just push them right now into test-infra? They may not care right now about multi-arch builds, but they should not mind merging a fix. What exactly are you fixing?

The bazel rule "@build_bazel_rules_nodejs//:defs.bzl" only support x86 arch. Add arm64 arch nodejs source code and binary, thus "bazel build" in k8s.io/test-infra project can be used in Arm64 platform. 
May need to send PR to bazel project to fix it. I am not quite familar about it, need to do more research.

images/dind-env-build-arm64.sh
zhlhahaha: > I would at least want to have the chance to build them for arm and ppc64 without having to care about the cluster-architecture. I am a little bit worried about too much friction here, which may make it hard for us to update things in the future. Multi-arch image is basically just the little extra-plus which eases maintenance a little bit more.



> At least it looks to me like your spcripts expects this to be run on an actual arm machine.
> 
> Did you think about cross-compiling it with bazel? Something like
> 
> ```
> bazel build --platforms=@io_bazel_rules_go//go/toolchain:linux_arm64_cgo ...
> ```
> 
> should already do the job.

Yes, as an arm server has joined in ci system, this script is using it to build arm image.
If we want to use "--platforms" to crossbuild prow utility **image** directly without any patch, it does not work, because the base image for prow utility image only support for x86. But I will try if we can use it crossbuild multiarch utility **binary**.

zhlhahaha:
Hi, @rmohr @danielBelenky The patch has updated.

zhlhahaha:
Hi @danielBelenky @dhiller please feel free to tell me if you have any suggetion or question on this PR. Thanks a lot.

images/kubevirt-infra-prow/publish_multiarch_prow_images.sh
```diff
@@ -0,0 +1,124 @@
...
+get_image_tag() {
+    local current_commit today
+    current_commit="$(git rev-parse HEAD)"
+    today="$(date +%Y%m%d)"
+    echo "v${today}-${current_commit:0:7}"
+}
+
+build_binaries(){
+    # Install bazel
+    apt update && apt install -y curl gnupg git python python3\
```
danielBelenky, line 69: Maybe we should prepare an image with all those tools pre-installed and a certain version of the test-infra repo pre-cloned?

images/kubevirt-infra-prow/publish_multiarch_prow_images.sh
```diff
@@ -0,0 +1,124 @@
+#!/bin/bash -x
```
danielBelenky, line 1: Maybe add `-e`?

danielBelenky (changes requested):

images/kubevirt-infra-prow/publish_multiarch_prow_images.sh
```diff
@@ -0,0 +1,124 @@
...
+
+BAZEL_VERSION=3.0.0
+PROW_COMMIT=master
+repo=kubevirtci
+archs=(arm64 ppc64le)
```
dhiller, line 6: Is there a reason why we are not building this image for amd64 also? Is that done somewhere implicitly?

images/kubevirt-infra-prow/publish_multiarch_prow_images.sh
```diff
@@ -0,0 +1,124 @@
...
+    push_images
+    push_manifest
+}
+
+help() {
+    cat <<EOF
+    Usage:
+        ./publish_multiarch_prow_images.sh [OPTIONS]
+
+    Build and publish mutliarch prow utility images.
```
dhiller, line 41: nit: spelling

dhiller:
> Hi, @rmohr @danielBelenky
> I have verified image build part of this PR in a periodics prow jobs in my local multi-arch prow cluster without any tags(e.g. preset-dind-enabled, preset-kubevirtci-docker-credential, preset-docker-mirror) and it works fine.
> I have done some research on how the tags works, I think it should be fine.
> Here are two items that have not been checked.
> 
> 1. whether the tags work?
>    -- The tag, preset-docker-mirror, use configmap "docker-daemon-mirror-config". And the tag, preset-kubevirtci-docker-credential, use secret "kubevirtci-docker-credential". Do we need to set the configmap and the secret in arm-cluster?
> 2. If the presubmit job and postsubmit job can be triggered successfully?
>    -- I do not have env to trigger presubmit job and postsubmit job, so it need to be verified.
> 
> As all new prow jobs are assigned to arm-cluster, it should be safe to current prow cluster.

@danielBelenky I faintly remember we had to remove the arm cluster? So will this work somehow on our cluster?

images/kubevirt-infra-prow/publish_multiarch_prow_images.sh
dhiller: OK, so this is just for building the images for other arch's outside of amd64. Understood. But still I wonder whether this could have some value for amd64 envs also.

zhlhahaha: > Is there a reason why we are not building this image for amd64 also? Is that done somewhere implicitly?
Hi, dhiller, prow utility images are already support amd64 and they are maintained by kubernetes community. We tried to push PR and make these images support multiarch, but there is no progress. Following link is discussion about it https://github.com/kubernetes/test-infra/issues/16588

images/kubevirt-infra-prow/publish_multiarch_prow_images.sh
zhlhahaha: > Maybe we should prepare an image with all those tools pre-installed and a certain version of the test-infra repo pre-cloned?

Ok

images/kubevirt-infra-prow/publish_multiarch_prow_images.sh
zhlhahaha: > nit: spelling

Should it be "Building and publishing mutliarch prow utility images"?

zhlhahaha:
> > Hi, @rmohr @danielBelenky
> > I have verified image build part of this PR in a periodics prow jobs in my local multi-arch prow cluster without any tags(e.g. preset-dind-enabled, preset-kubevirtci-docker-credential, preset-docker-mirror) and it works fine.
> > I have done some research on how the tags works, I think it should be fine.
> > Here are two items that have not been checked.
> > 
> > 1. whether the tags work?
> >    -- The tag, preset-docker-mirror, use configmap "docker-daemon-mirror-config". And the tag, preset-kubevirtci-docker-credential, use secret "kubevirtci-docker-credential". Do we need to set the configmap and the secret in arm-cluster?
> > 2. If the presubmit job and postsubmit job can be triggered successfully?
> >    -- I do not have env to trigger presubmit job and postsubmit job, so it need to be verified.
> > 
> > As all new prow jobs are assigned to arm-cluster, it should be safe to current prow cluster.
> 
> @danielBelenky I faintly remember we had to remove the arm cluster? So will this work somehow on our cluster?

Really?

images/kubevirt-infra-prow/publish_multiarch_prow_images.sh
dhiller: No, sorry for being unclear. I meant it needs to be mu*lt*iarch instead of mu*tl*iarch :blush:

images/kubevirt-infra-prow/publish_multiarch_prow_images.sh
dhiller: Well, I wasn't aware of this. Should be fine then.

dhiller (approved):
/lgtm

zhlhahaha:
I will use multi-image builder for prow image build after https://github.com/kubevirt/project-infra/pull/612 get merged.

zhlhahaha:
> I will use multi-image builder for prow image build after #612 get merged.

As building multiarch prow utility image job is unlikely to trigger often, I think a builder image is not necessary.

zhlhahaha:
@rmohr @dhiller 
I have verified it on my local x86 server, but need to test it on kubevirt CI system.

images/kubevirt-infra-prow/publish_multiarch_prow_images.sh
```diff
@@ -0,0 +1,128 @@
...
+}
+
+get_base_image() {
+    if [ $1 == "arm64" ]; then
+        echo "arm64v8/alpine"
+    elif [ $1 == "ppc64le" ] ; then
+        echo "ppc64le/alpine"
+    else
+        echo "unsupport arch"
+        exit
```
dhiller, line 57: Do we want to silently ignore this?

images/kubevirt-infra-prow/publish_multiarch_prow_images.sh
```diff
@@ -0,0 +1,128 @@
...
+    today="$(date +%Y%m%d)"
+    echo "v${today}-${current_commit:0:7}"
+}
+
+build_binaries(){
+    # Cleanup Env
+    rm -rf /usr/bin/bazel /usr/local/bin/python prow test-infra
+
+    # Install bazel
+    apt update && apt install -y curl gnupg git python python3\
```
dhiller, line 73: We have a bootstrap with bazel(isk) installed: https://github.com/kubevirt/project-infra/blob/master/github/ci/prow-deploy/files/jobs/kubevirt/project-infra/project-infra-postsubmits.yaml#L24
Maybe we can skip installing bazel when using that image?
Maybe you even want to checkout that image whether it has part of the tools you need?

dhiller (commented):
Looks good in general, I have added two nits inline.

PR #591, opened 2020-08-24
Fix make generate to run the right target

oshoval:
The right target also check the git directory is clean
after running make generate.
In order to see the expected status is met.


oshoval:
@danielBelenky 
why it added approved automatically ?

PR #593, opened 2020-08-25
Allow more executions in parallel of the core kubevirt/kubevirt lanes

rmohr:
Make use of the two new nodes which got added.


rmohr:
/hold

Still verifying that the second node is healthy too.

PR #598, opened 2020-08-26
Fix generate verify job

danielBelenky:
This change ensures that the targets that we execute after we run
generate, will be executed in the same environment and will be able to
consume the changes that were done by generate.


danielBelenky:
/hold
let's wait until we fix the master branch before we merge this so we won't introduce noise.

PR #606, opened 2020-08-28
Allow non org admins but project owners to override jobs

qinqon:
Right now users need to be github org admins to be able to override
a job since [1] merged at prow users at OWNERS file can override
them if configured, this commit activate that feature.

[1] https://github.com/kubernetes/test-infra/pull/14538


qinqon:
/cc @danielBelenky @phoracek 

That's why we cannot override jobs at kubemacpool.

rmohr:
Nice.

/lgtm
/approve

PR #608, opened 2020-08-28
Redesign of Prow deployment

gabrielecerami:
This is a series of patches to add a redesign of the prow deployment, to lower maintenance and add tests.


gabrielecerami:
/cc @dhiller 
/cc @rmohr 
Github identified the token in my code and revoked it. I'm exploring how to pass secret to jobs in prow.

rmohr:
@danielBelenky I think you PR does not rebase, can that be? It sees all these jobs as added. In fact I removed all these jobs and the only exist on the PR of @gabrielecerami since it is based on an older state. @gabrielecerami I think if you rebase that awkwardness would  disappear. @danielBelenky something which we should fix.

github/ci/prow-deploy/README.md
```diff
@@ -0,0 +1,209 @@
...
+## How to launch a deployment
+
+First you need to create a deployment configuration file.
+The role will look by default in
+
+    /etc/prow-deploy/deploy_config.yaml
+    
+An example can be found in directory
+
+    molecule/default/vars/deploy_config.yaml
```
fgimenez, line 53: I can't find this file, is it added as part of this PR?

fgimenez:
/lgtm

All ok in the last build with the current status https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_project-infra/608/prow-deploy-test/1334035927168716800 

@rmohr this is ready to go PTAL.

rmohr:
/lgtm
/approve

@gabrielecerami @fgimenez thank you for you hard work. :+1:

rmohr:
/hold

Let's see if wa can sort out the owner file issue

PR #612, opened 2020-09-02
add task for mulitarch image builder

zhlhahaha:
This image would be used on building multi-arch
prow utility images and bootstrap images


zhlhahaha:
@danielBelenky @dhiller 
Hi Daniel, this is an image with all those tools pre-installed and a certain version of the test-infra repo pre-cloned.

images/kubevirt-infra-multiarch-image-builder/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM gcr.io/k8s-testimages/bootstrap:v20190516-c6832d9
+
+RUN apt update && apt install -y curl gnupg git python python3 \
+        && curl -fsSL https://bazel.build/bazel-release.pub.gpg | gpg --dearmor > bazel.gpg \
+        && mv bazel.gpg /etc/apt/trusted.gpg.d/ \
+        && echo "deb [arch=amd64] https://storage.googleapis.com/bazel-apt stable jdk1.8" | tee /etc/apt/sources.list.d/bazel.list \
+        && apt update && apt install -y bazel-3.0.0 \
+        && ln -s /usr/bin/bazel-3.0.0 /usr/bin/bazel
```
qinqon, line 8: Why not bazelisk ?

qinqon (changes requested):

images/kubevirt-infra-multiarch-image-builder/Dockerfile
zhlhahaha: > Why not bazelisk ?

Hi Enrique, thanks for reply. I do not use bazelisk before, and bazel is fair enough for build multi-arch binary.
https://github.com/kubevirt/project-infra/pull/587#discussion_r481090049 Here is some discussion about this image.

danielBelenky: I think that bazelisk is better here because it will determine and install the right bazel version for us.
But, since it is a container that we build anyway, I don't mind installing the correct bazel version ourselves.

Just for reference, here's the repo: https://github.com/bazelbuild/bazelisk

danielBelenky:
/lgtm
@qinqon I'm ok if we won't use bazelisk at this point but it's up to you.

zhlhahaha:
Hi @danielBelenky  @qinqon I have updated the PR. bazelisk is quite easy to use.

zhlhahaha:
Hi @danielBelenky  @qinqon 
Would you mind to review this PR when you have time?
Thanks a lot.

PR #617, opened 2020-09-03
labels: change good first issue label name

alonSadan:
The label `good first issue` , is not triggred by the bot when
asked. Changing it's name to `good-first-issue` should solve it.

Signed-off-by: alonSadan <asadan@redhat.com>


dhiller (approved):
/lgtm
/approve

I saw that we have more like this. If this works, I'm going to fix the others also.

PR #626, opened 2020-09-07
Rename build-officer to build-watcher

oshoval:
Renaming according the rebrand.

### DIFF
--- a/github/ci/prow/files/labels.yaml
+++ b/github/ci/prow/files/labels.yaml
@@ -2,13 +2,13 @@
 default:
   labels:
     - color: 8452c9
-      name: triage/build-officer
+      name: triage/build-watcher
       target: issues
-      description: Discovered by a kubevirt build-officer. Pay attention to these issues to keep CI healthy.
+      description: Discovered by a kubevirt build-watcher. Pay attention to these issues to keep CI healthy.
       prowPlugin: label
       addedBy: anyone
       previously:
-        - name: build-officer
+        - name: triage/build-officer
           color: 8452c9
     - color: c2e0c6
       name: release-note


github/ci/prow/files/labels.yaml
```diff
@@ -2,13 +2,13 @@
...
-      name: triage/build-officer
+      name: triage/build-watcher
       target: issues
-      description: Discovered by a kubevirt build-officer. Pay attention to these issues to keep CI healthy.
+      description: Discovered by a kubevirt build-watcher. Pay attention to these issues to keep CI healthy.
       prowPlugin: label
       addedBy: anyone
       previously:
-        - name: build-officer
+        - name: build-watcher
```
rmohr, line 11: Here we need to keep build-officer to let the rename happen.

rmohr (changes requested):
Almost there.

github/ci/prow/files/labels.yaml
oshoval: nice, thanks, on it

oshoval: Done

github/ci/prow/files/labels.yaml
```diff
@@ -2,9 +2,9 @@
 default:
   labels:
     - color: 8452c9
-      name: triage/build-officer
+      name: triage/build-watcher
       target: issues
-      description: Discovered by a kubevirt build-officer. Pay attention to these issues to keep CI healthy.
```
dhiller, line 7: According to [documentation](https://github.com/kubernetes/test-infra/tree/master/label_sync#configuration) you should add a `previously:` section, as you are modifying an existing label.

dhiller (changes requested):

github/ci/prow/files/labels.yaml
oshoval: also for the description ?
because we do have previously

(see also https://github.com/kubevirt/project-infra/pull/626#discussion_r484357055)

dhiller: The `previously` was there from a previous label-sync run. IIUC we now need to update that section from

```
previously:
   - name: build-officer
      ...
```

to 

```
previously:
   - name: triage/build-officer
      ...
```

as the label currently is `triage/build-officer`.

dhiller (changes requested):

github/ci/prow/files/labels.yaml
dhiller: Sorry for marking the wrong line BTW.

oshoval: on it, 
we need only triage/build-officer then ?
no need to cascade both right ? (in case there is an option to even)

oshoval: Done thanks

dhiller (commented):
/lgtm
/approve

dhiller (approved):

github/ci/prow/files/labels.yaml
rmohr: it would have been an array, you normally put all previous labels in that array. But no old mentions are left, so should not be an issue.

oshoval: Thanks both of you

PR #633, opened 2020-09-15
Adding persistent storage volume for Grafana data

rhrazdil:
Adding a PVC and mounting it to grafana so that the grafana database survives restarts of the pod.
Mounting the PVC volume to /var/lib/grafana, where the database is located.

### DIFF
--- a/github/ci/prometheus/templates/grafana.yaml
+++ b/github/ci/prometheus/templates/grafana.yaml
@@ -1,4 +1,16 @@
-
+kind: PersistentVolumeClaim
+apiVersion: v1
+metadata:
+  labels:
+    app: grafana
+  name: grafana-data
+spec:
+  accessModes:
+    - ReadWriteOnce
+  resources:
+    requests:
+      storage: 32Gi
+  volumeName: "storage01-pv103-32g"
 ---
 apiVersion: apps/v1beta1
 kind: StatefulSet
@@ -36,6 +48,8 @@ spec:
             - cfg:default.paths.data=/var/lib/grafana
             - cfg:default.log.level=debug
           volumeMounts:
+            - name: grafana-storage
+              mountPath: /var/lib/grafana
             - name: config-volume
               mountPath: /etc/grafana
               readOnly: true
@@ -53,6 +67,9 @@ spec:
               memory: 100Mi            
       terminationGracePeriodSeconds: 300
       volumes:
+        - name: grafana-storage
+          persistentVolumeClaim:
+            claimName: grafana-data
         - name: config-volume
           secret:
             secretName: grafana-config


phoracek:
@rhrazdil have you manage to verify this?

/cc @danielBelenky

rhrazdil:
No, how can I run this on the real cluster

phoracek:
I don't think we have to run on a real cluster. I was thinking that we should just dpeloy it on kubevirtci, make some changes to the config, kill the pod and see if our changes were kept after the boot.

But maybe Daniel has a better suggestion on how to verify this (if needed at all).

rhrazdil:
One thing that's missing at this moment is making grafana to actually use the hostDir.
I see the config file path is `--config=/etc/grafana/grafana.ini`, we will probably need to move or this to the created hostDir and point grafana to the new config location.
Daniel, could you suggest how to tackle this?

rhrazdil:
@danielBelenky sorry for the confusion in the previous comments, the point of this PR is not to move the config, but the data of course.

I have changed the `cfg:default.paths.data` setting to `/hoststorage` and tried running the grafana pod, but I didn't see grafana to actually populate the `/hoststorage` with any data.

I don't see any other related setting for sqllite3 database location int the [docs](https://grafana.com/docs/grafana/v5.0/installation/configuration/).

Am I missing something?

phoracek:
Have you tried to write in the directory manually, to see if it is accessible and data synced properly?

danielBelenky:
> didn

Did you see any data that was sent to Grafana? Also, maybe you could just mount the host path to the already configured `/var/lib/grafana`?

rhrazdil:
> Also, maybe you could just mount the host path to the already configured /var/lib/grafana?
@danielBelenky Interesting idea, tried this and all appears to be working.

rhrazdil:
@danielBelenky @jean-edouard could you review please?

github/ci/prometheus/templates/grafana.yaml
```diff
@@ -26,6 +26,9 @@ spec:
         zone: ci
       containers:
         - name: grafana
+          securityContext:
+            runAsUser: 0
+            privileged: true
```
danielBelenky, line 31: I'm not sure that I like this... maybe we should allow some host paths in our cluster for unprivileged users.
@dhiller @rmohr what do you think?

danielBelenky (changes requested):

github/ci/prometheus/templates/grafana.yaml
dhiller: Same here. Didn't we talk about adding a PVC here BTW?

rhrazdil: @dhiller @danielBelenky Hello, if I use a PVC, what storage class can/should I use for it?

PR #635, opened 2020-09-16
Mount lib modules

qinqon:
As part of https://github.com/kubevirt/kubevirtci/pull/445 we need prow to mount `/lib/modules` so dnsmasq.sh can load ip6table kernel modules in case they are not loaded already at cluster-up.

### DIFF
--- a/github/ci/prow/files/jobs/kubevirt/cluster-network-addons-operator/cluster-network-addons-operator-presubmits.yaml
+++ b/github/ci/prow/files/jobs/kubevirt/cluster-network-addons-operator/cluster-network-addons-operator-presubmits.yaml
@@ -17,6 +17,7 @@ presubmits:
         preset-dind-enabled: "true"
         preset-docker-mirror: "true"
         preset-shared-images: "true"
+        preset-mount-lib-modules: "true"
       spec:
         nodeSelector:
           type: bare-metal-external
@@ -79,6 +80,7 @@ presubmits:
         preset-dind-enabled: "true"
         preset-docker-mirror: "true"
         preset-shared-images: "true"
+        preset-mount-lib-modules: "true"
       spec:
         nodeSelector:
           type: bare-metal-external
--- a/github/ci/prow/files/config.yaml
+++ b/github/ci/prow/files/config.yaml
@@ -328,3 +328,14 @@ presets:
     - name: kubevirtci-cred
       mountPath: /etc/kubevirtci-cred
       readOnly: true
+- labels:
+    preset-mount-lib-modules: "true"
+  volumes:
+    - name: modules
+      hostPath:
+        path: /lib/modules
+        type: Directory
+  volumeMounts:
+    - name: modules
+      mountPath: /lib/modules
+      readOnly: true
--- a/github/ci/prow/files/jobs/k8snetworkplumbingwg/kubemacpool/kubemacpool-presubmits.yaml
+++ b/github/ci/prow/files/jobs/k8snetworkplumbingwg/kubemacpool/kubemacpool-presubmits.yaml
@@ -47,6 +47,7 @@ presubmits:
         preset-dind-enabled: "true"
         preset-docker-mirror: "true"
         preset-shared-images: "true"
+        preset-mount-lib-modules: "true"
       spec:
         nodeSelector:
           type: bare-metal-external
--- a/github/ci/prow/files/jobs/kubevirt/containerized-data-importer/containerized-data-importer-presubmits.yaml
+++ b/github/ci/prow/files/jobs/kubevirt/containerized-data-importer/containerized-data-importer-presubmits.yaml
@@ -16,6 +16,7 @@ presubmits:
       preset-dind-enabled: "true"
       preset-docker-mirror: "true"
       preset-shared-images: "true"
+      preset-mount-lib-modules: "true"
     spec:
       nodeSelector:
         type: bare-metal-external
@@ -48,6 +49,7 @@ presubmits:
       preset-dind-enabled: "true"
       preset-docker-mirror: "true"
       preset-shared-images: "true"
+      preset-mount-lib-modules: "true"
     spec:
       nodeSelector:
         type: bare-metal-external
@@ -80,6 +82,7 @@ presubmits:
       preset-dind-enabled: "true"
       preset-docker-mirror: "true"
       preset-shared-images: "true"
+      preset-mount-lib-modules: "true"
     spec:
       nodeSelector:
         type: bare-metal-external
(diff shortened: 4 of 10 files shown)


github/ci/prow/files/jobs/kubevirt/kubevirtci/kubevirtci-presubmits.yaml
```diff
@@ -421,6 +421,15 @@ presubmits:
         resources:
           requests:
             memory: "8Gi"
+        volumeMounts:
```
dhiller, line 424: To avoid duplication maybe create a preset for this?

qinqon: Better, will do like that.

qinqon:
@dhiller @rmohr Another option is to make mounting `/lib/modules` optional at gocli but then we have to be sure that ip6tables kernel modules are loaded at prow nodes since it's run by dnsmasq.sh.

github/ci/prow/files/jobs/kubevirt/cluster-network-addons-operator/cluster-network-addons-operator-presubmits.yaml
```diff
@@ -94,37 +65,6 @@ presubmits:
               - "/bin/sh"
               - "-c"
               - "automation/check-patch.e2e-workflow-k8s.sh"
-    - name: pull-e2e-cluster-network-addons-operator-workflow-okd
```
phoracek, line 97: Please remove these changes from this PR. I'd be happy to take them in right away on a different PR.

phoracek (changes requested):

qinqon:
@phoracek done

phoracek (approved):
/lgtm

Does @dhiller buy this?

qinqon:
@phoracek let's wait for @dhiller and @rmohr LGTMs

rmohr:
I would prefer not to do that if possible. How often do we think that we hit a machine like this? How common is it to not have ip6modules support enabled? I have it for instance on my fedora and never enabled or loaded it explicitly.

phoracek:
It happened on our CI: https://storage.googleapis.com/kubevirt-prow/pr-logs/pull/k8snetworkplumbingwg_kubemacpool/244/pull-kubemacpool-e2e-k8s/1306141158832869376/build-log.txt?Expires=1600261120&GoogleAccessId=prow-46%40kubevirt-240711.iam.gserviceaccount.com&Signature=GP1ZLIViahiNZ7YDEz%2BTmbk8PJDhHB94YKPfk32y1LHZkT0jOJXNj6XZA95pxW3LMxEmnp8TsEs%2FQ3Y%2F1z2mjqqrzfwpKFily0JD6NFUJleW5ff3beZwlyut5yQux5tqiLYqJAOdnyvWGI3DQ2yp%2FKNydTKbkw34Ed73HvZUBgFbpUsAukrHngOjJ8fULuf%2BIBU0MiLN%2BtMN%2Fg1mMklYBsAlUf7XAI1X1ClTtJXKS6VG8CnCM2CJZsjFZa7VHrNDLyPak%2FWUC7VfGM7Adiwsb58BLaZOLSbCYFbviXM7uH5cBMKitomEV2fpfM1GYRLvNwTUkxJHBmAktfJ7gsWsDg%3D%3D

I have not hit this issue locally neither, but there were couple people who did. That is not to be resolved by this infra PR though.

qinqon:
@rmohr we found it at Alona's f31 machine, so we decide to put in place a solution, but I am pretty sure that prow nodes have those modules loaded, the alternative to this is to make mounting `/lib/modules` optional at gocli and not doing it on prow with and env var or the like, what do you think ?

phoracek:
Oh right, thanks @qin

rmohr:
> @rmohr we found it at Alona's f31 machine, so we decide to put in place a solution, but I am pretty sure that prow nodes have those modules loaded, the alternative to this is to make mounting `/lib/modules` optional at gocli and trying to mount it only if it exists.
> 
> @phoracek that issue is created by the gocli PR that points `/lib/modules` not by modules not being loaded.

How about a pre-flight check + an extra command which we provide which can "forcefully" setup your local machine? I am not super-concerned about your approach looking at the position from where we are right now. The containers run with extensive privileges, but we do not modify something out side of them normally. However there are three points:

 * We should become more secure over time (even if it is just a test environment) on local machine.
 * We should be come more secure on CI, and we can definitely right now ensure that the module is no the machine and that we don't mount this extra path.
 * I find it rude if we do modifications outside of our container without asking for permission.

rmohr:
> * I find it rude if we do modifications outside of our container without asking for permission.

Oh, one question here. Would the module be loaded if I would run dnsmasq non-privileged in user-space outside of the container?

qinqon:
@rmohr I have just remember that we have some "preflight" checks at kubevirtci on `cluster-up/check.sh` we can add the check there and fail if modules are not loaded with instructions about loading them or even do a `sudo ip6tables` command there to load the modules.

About dnsmasq.sh running out of container without privileged, we cannot run unprivileged ip6tables it fails.

```bash
[ellorent@localhost kubevirtci]$ ip6tables -L
Fatal: can't open lock file /run/xtables.lock: Permission denied
```

rmohr:
> About dnsmasq.sh running out of container without privileged, we cannot run unprivileged ip6tables it fails.

Ok, so it would probably auto-load the module if outside of our container as privileged.

rmohr:
Ha, I was completely off track. You already mounted the gocli part. Now I get the full context. I think we should assume properly prepared machines. Can you just make it optional in kubevirtci? If we start provisioning our own machines (we don't own machines right now), we can just add  such a kernel load requirement to ansible or the like. Would that make sense?

qinqon:
But that's for CI, we found an issue at user doing cluster-up at kubevirt, and there is not that problematic mounting modules I would say.
For CI we need machines to be prepared for users maybe is not that bad that dnsmasq.sh mount some modules with `ip6tables` command.

qinqon:
So we have three options I would say:
- Mount /lib/modules if it exists
- Add a preflight check that check if ip6tables modules are there.
- Make mounting /lib/modules optional passing a flag 

From CI perspectiove we don't want to change prow nodes, from users perspective loading a module is not that bad.

rmohr:
> From CI perspectiove we don't want to change prow nodes, from users perspective loading a module is not that bad.

 Agreed. In prow we have a CI=true variable. Maybe you can check for it in gocli and do things dynamically and just close this PR? How does that sound?

qinqon:
Closing this in favor of https://github.com/kubevirt/kubevirtci/pull/446

PR #657, opened 2020-10-01
Make travis ci run required for the last ten release branches 

dhiller:
This elaborates on commit 8c56d2f, where travis ci run was made non required for any branch.

As we still need to run Travis on the release branches we add rules for
the last ten release branches.

/cc @rmohr

### DIFF
--- a/github/ci/prow/files/config.yaml
+++ b/github/ci/prow/files/config.yaml
@@ -195,6 +195,51 @@ branch-protection:
       repos:
         kubevirt:
           protect: true
+          branches:
+            release-0.33:
+              required_status_checks:
+                contexts:
+                  - continuous-integration/travis-ci/pr
+            release-0.32:
+              required_status_checks:
+                contexts:
+                  - continuous-integration/travis-ci/pr
+            release-0.31:
+              required_status_checks:
+                contexts:
+                  - continuous-integration/travis-ci/pr
+            release-0.30:
+              required_status_checks:
+                contexts:
+                  - continuous-integration/travis-ci/pr
+            release-0.29:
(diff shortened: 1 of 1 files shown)


github/ci/prow/files/config.yaml
```diff
@@ -195,6 +195,9 @@ branch-protection:
       repos:
         kubevirt:
           protect: true
+          required_status_checks:
```
rmohr, line 198: Maybe we can just name the last 10  release branches explicitly here and not protect master and new releases?

dhiller: I thought we just were talking about removing the required-status-checks: for master only? So you are saying we should also remove the **protection** for master?

I understood we should do sth like this, no?
```suggestion
          branches:
            release-0.33:
              required_status_checks:
                contexts:
                - continuous-integration/travis-ci/pr
            release-0.32:
              required_status_checks:
                contexts:
                - continuous-integration/travis-ci/pr
            ...
```

rmohr: Never mind, we can do that later.

rmohr: No wait, yes please do it like this. I don't want to disturb tide more than necessary.

dhiller:
@rmohr updated with requiring travis for the last ten branches, PTAL!

rmohr:
/lgtm
/approve
/hold

Would you mind updating the PR title and the PR description?

dhiller:
> /lgtm
> /approve
> /hold
> 
> Would you mind updating the PR title and the PR description?

@rmohr done

PR #669, opened 2020-10-09
Login to dockerhub in the kubevirtci release job

rmohr:
If the job is not logged in to dockerhub it can't push.


rmohr:
Ok, another change. Had to change from `/bin/sh` to `/bin/bash` to make the login work. Job is running here: https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/logs/publish-kubevirtci/1314533833831354368

PR #684, opened 2020-11-03
SRIOV, Change to not always run

oshoval:
As we are working on stabilizing the lane,
and the report is off:

1. there is no reason that each PR will run
it, without it being shown in the PR.
2. we dont have enough resources to run it
when it flakes, we need to wait 1 day
sometimes when we do need it.
3. its reports are causing noise on flake finder,
we know its not stable but flake finder watcher
shouldnt invest time when he sees it.

See please
https://github.com/kubevirt/project-infra/pull/680/files#r516458644

### DIFF
--- a/github/ci/prow/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
+++ b/github/ci/prow/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
@@ -169,7 +169,7 @@ presubmits:
     annotations:
       fork-per-release: "true"
       k8s.v1.cni.cncf.io/networks: multus-cni-ns/sriov-passthrough-cni
-    always_run: true
+    always_run: false
     optional: true
     skip_report: true
     decorate: true


ormergi:
We need this lane up to get insights about it behavior so we can stabilize it.
Disabling it now is going two steps back on the effort for stabilizing this lane.

oshoval:
> We need this lane up to get insights about it behavior so we can stabilize it.
> Disabling it now is going two steps back on the effort for stabilizing this lane.

we run it mamually each time we need it, and wait in Q for 1 day (see https://github.com/kubevirt/kubevirt/pull/4472)
see please PR description reason,
i just want to make us able to move faster, each time you need it simply run it,
this is what we are doing anyway because of the timeouts (see my remark on 680, about a possible direction we should investigate for that please, related to timeouts)

oshoval:
One reason to keep it "always run" is in order to debug the problems mentioned on the PR description (point 2).
see please all the gray mark errors on kubevirt https://github.com/kubevirt/kubevirt/pull/4472 from yesterday.
But once the reason will be known, at least until its solved it might be right to stop "always run" until we stabilize the lane imo.

A progress was started about those gray marks.

PR #687, opened 2020-11-03
Added uploader periodic job and uploader plugin readme

fgimenez:
Signed-off-by: Federico Gimenez <federico.gimenez@gmail.com>

### DIFF
--- a/github/ci/prow/files/jobs/kubevirt/project-infra/project-infra-periodics.yaml
+++ b/github/ci/prow/files/jobs/kubevirt/project-infra/project-infra-periodics.yaml
@@ -209,3 +209,54 @@ periodics:
         secret:
           secretName: gcs
 
+- name: periodic-kubevirt-mirror-uploader
+  cron: "05 7 * * *"
+  max_concurrency: 1
+  decorate: true
+  decoration_config:
+    timeout: 1h
+    grace_period: 5m
+  extra_refs:
+  - org: kubevirt
+    repo: project-infra
+    base_ref: master
+    workdir: true
+  - org: kubevirt
+    repo: kubevirt
+    base_ref: master
+  - org: kubevirt
+    repo: containerized-data-importer
+    base_ref: master
+  spec:
+    containers:
+      - image: "kubevirt/builder@sha256:e6a76ec71f2de87c3bc648bae87ab43a113540bf57da69f07f76bd0edfd500ab"
+        env:
+        - name: GIT_ASKPASS
+          value: ../project-infra/hack/git-askpass.sh
+        - name: GOOGLE_APPLICATION_CREDENTIALS
+          value: /etc/gcs/service-account.json
+        command: ["/bin/sh"]
+        args:
+          - "-c"
+          - mkdir ${HOME}/bazel &&
+            curl -L https://github.com/bazelbuild/bazelisk/releases/download/v1.7.4/bazelisk-linux-amd64 --output ${HOME}/bazel/bazelisk &&
+            chmod a+x ${HOME}/bazel/bazelisk &&
+            export PATH=${PATH}:${HOME}/bazel &&
+            hack/git-pr.sh -c "bazelisk run //plugins/cmd/uploader:uploader -- -workspace ${PWD}/../kubevirt/WORKSPACE -dry-run=false" -p ../kubevirt -r kubevirt &&
+            hack/git-pr.sh -c "bazelisk run //plugins/cmd/uploader:uploader -- -workspace ${PWD}/../containerized-data-importer/WORKSPACE -dry-run=false" -p ../containerized-data-importer -r containerized-data-importer
+        volumeMounts:
+        - name: token
+          mountPath: /etc/github
+        - name: gcs
+          mountPath: /etc/gcs
+          readOnly: true
+        resources:
+          requests:
+            memory: "200Mi"
+    volumes:
+    - name: token
+      secret:
+        secretName: oauth-token
+    - name: gcs
+      secret:
+        secretName: gcs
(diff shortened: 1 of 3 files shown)


fgimenez:
/hold

Testing with the branch proposed on #686, I'll revert this to master once all is working.

github/ci/prow/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml
```diff
@@ -266,3 +266,41 @@ periodics:
...
+  spec:
+    containers:
+      - image: "kubevirt/builder@sha256:e6a76ec71f2de87c3bc648bae87ab43a113540bf57da69f07f76bd0edfd500ab"
+        env:
+        - name: GIT_ASKPASS
+          value: "hack/git-askpass.sh"
+        command:
+          - "hack/git-pr.sh"
+          - "-c"
+          - "bazelisk run //plugins/cmd/uploader:uploader -- -workspace ../kubevirt/WORKSPACE"
```
rmohr, line 294: I don't think that we have bazelisk installed in the builder image (the image contains bazel only at the moment). Running `go get -u github.com/bazelbuild/bazelisk` may solve that.

rmohr (changes requested):
Once we are sure that bazelisk is present, I will create a job. Looks great.

github/ci/prow/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml
fgimenez: Thx @rmohr following our online conversation i've pushed changes to use the bazel version present in the image, which is the required one.

fgimenez: Finally the bazel version installed in the image doesn't match the required, pushed changes to install it as suggested

github/ci/prow/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml
```diff
@@ -266,3 +266,52 @@ periodics:
...
+      - image: "kubevirt/builder@sha256:e6a76ec71f2de87c3bc648bae87ab43a113540bf57da69f07f76bd0edfd500ab"
+        env:
+        - name: GIT_ASKPASS
+          value: "hack/git-askpass.sh"
+        - name: GOOGLE_APPLICATION_CREDENTIALS
+          value: /etc/gcs/service-account.json
+        command: ["/bin/sh"]
+        args:
+          - "-c"
+          - curl -L https://golang.org/dl/go1.15.3.linux-amd64.tar.gz --output go1.15.3.linux-amd64.tar.gz &&
```
rmohr, line 296: That is a nice idea. in general we have golang already in there.  The issue is that the environment is not initialized properly, since go was installed via gimme.

github/ci/prow/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml
```diff
@@ -266,3 +266,52 @@ periodics:
...
+    base_ref: master
+  spec:
+    containers:
+      - image: "kubevirt/builder@sha256:e6a76ec71f2de87c3bc648bae87ab43a113540bf57da69f07f76bd0edfd500ab"
+        env:
+        - name: GIT_ASKPASS
+          value: "hack/git-askpass.sh"
+        - name: GOOGLE_APPLICATION_CREDENTIALS
+          value: /etc/gcs/service-account.json
+        command: ["/bin/sh"]
```
rmohr, line 293: I think that if you put `/entrypoint.sh` here (https://github.com/kubevirt/kubevirt/blob/master/hack/builder/Dockerfile#L109) you should have a working env.

fgimenez: Thanks a lot for the pointer, the problem is that `/entrypoint.sh` uses `eval` and currently the generic PR script relies on `eval` too,  using `/entrypoint.sh` leads to nested evals and escaping issues. 

Instead I've pushed changes to download the bazelisk binary (no need for go in the path), I've also fixed some paths that were wrong bc of the different orgs for each repo (this should be removed once the generic PR script is merged), PTAL.

github/ci/prow/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml
```diff
@@ -266,3 +266,49 @@ periodics:
...
+        args:
+          - "-c"
+          - curl -L https://github.com/bazelbuild/bazelisk/releases/download/v1.7.4/bazelisk-linux-amd64 --output /usr/bin/bazelisk &&
+            chmod a+x /usr/bin/bazelisk &&
+            hack/git-pr.sh -c "bazelisk run //plugins/cmd/uploader:uploader -- -workspace ../../kubevirt/kubevirt/WORKSPACE" -p ../../kubevirt/kubevirt
```
rmohr, line 298: Why is `-p` added here? Are you sure that this is needed?` We would also need `-dry-run=false`.

rmohr (changes requested):

github/ci/prow/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml
rmohr: Ah never mind regarding the `-p`. I thought for a moment that it was passed to `uploader`, which it is not.

fgimenez: yep, added `-dry-run=false`, moved to project-infra periodics and added containerized-data-importer as extra_ref plus a call to execute the uploader on it

PR #704, opened 2020-11-13
Add bootstrap image with docker mirror proxy support

fgimenez:
The image is based on `gcr.io/k8s-testimages/bootstrap:v20201106-b01d6e7` and includes a wrapper around `/usr/local/bin/runner.sh` that:
* verifies if there are env vars defined for setting up HTTP proxy, renaming them as needed. The docker daemon process requires `HTTP_PROXY` and `HTTPS_PROXY` env vars to be defined, we plan to use the more specific `DOCKER_HTTP_PROXY` and `DOCKER_HTTPS_PROXY` in presets.
* optionally adds a CA certificate to the system-wide trusted root authority, this will allow connecting the docker daemon with the mirror proxy using TLS.

### DIFF
--- a/images/bootstrap/Dockerfile
+++ b/images/bootstrap/Dockerfile
@@ -1,0 +1,20 @@
+# Copyright 2020 The KubeVirt Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+FROM gcr.io/k8s-testimages/bootstrap:v20201106-b01d6e7
+
+RUN cp /usr/local/bin/runner.sh /usr/local/bin/runner_orig.sh
+
+COPY "runner.sh" \
+  "/usr/local/bin/"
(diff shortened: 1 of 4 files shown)


images/bootstrap/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+FROM gcr.io/k8s-testimages/bootstrap:v20201106-b01d6e7
+
+COPY [
+  "runner_wrapper.sh", \
```
rmohr, line 18: I see that we run `runner.sh` in pretty much all our prowjobs. Would it make sense to rename `runner.sh` and name the wrapper `runner.sh` instead?

images/bootstrap/runner_wrapper.sh
```diff
@@ -0,0 +1,44 @@
...
+setup_ca(){
+    if [ -f "${CA_CERT_FILE}" ]; then
+        cp "${CA_CERT_FILE}" /usr/local/share/ca-certificates/
+
+        update-ca-certificates
+    fi
+}
+
+main(){
+    setup_env
```
rmohr, line 37: I think that there is a debian service for docker in that image. See `service docker start` there. I think we only want to inject the proxies for this service and not for anything else which happens in there.

rmohr (changes requested):

images/bootstrap/Dockerfile
fgimenez: Good point, done PTAL

rmohr:
/lgtm
/approve
/hold

Looks good to me. @fgimenez feel free to remove the hold if you could try out the image already.

rmohr:
/lgtm
/approve

@fgimenez it takes some time until job changes are propagated. We will have to trigger the postsubmit manually, since it will miss the code-change of  this merge.

fgimenez:
/hold cancel

> @fgimenez it takes some time until job changes are propagated. We will have to trigger the postsubmit manually, since it will miss the code-change of this merge.

sure will do, thanks @rmohr!

PR #706, opened 2020-11-13
Add docker mirror proxy

fgimenez:
Creates the mirror manifests, the secret for the proxy CA cert and the job presets. Currently using the image `fgimenez/docker-mirror-proxy:v0.1.0` built from https://github.com/fgimenez/docker-registry-proxy

### DIFF
--- a/github/ci/prow/files/config.yaml
+++ b/github/ci/prow/files/config.yaml
@@ -354,6 +354,23 @@ presets:
     mountPath: /etc/default
     readOnly: true
 - labels:
+    preset-docker-mirror-proxy: "true"
+  env:
+  - name: DOCKER_HTTP_PROXY
+    value: http://docker-mirror-proxy.kubevirt-prow.svc:3128
+  - name: DOCKER_HTTPS_PROXY
+    value: http://docker-mirror-proxy.kubevirt-prow.svc:3128
+  - name: CA_CERT_FILE
+    value: /etc/docker-mirror-proxy/ca.crt
+  volumes:
+  - name: docker-mirror-proxy-ca-cert
+    configMap:
+      name: mirror-proxy-config
+  volumeMounts:
+  - name: docker-mirror-proxy-ca-cert
+    mountPath: /etc/docker-mirror-proxy
+    readOnly: true
+- labels:
     preset-shared-images: "true"
   volumes:
   - name: shared-iso
--- a/github/ci/prow/tasks/main.yml
+++ b/github/ci/prow/tasks/main.yml
@@ -138,6 +138,20 @@
           type: Opaque
           data:
             service-account.json: "{{ gcsServiceAccount | b64encode }}"
+    - name: Create Docker mirror proxy secret
+      k8s:
+        kubeconfig: '{{ automationKubeconfig }}'
+        context: '{{ masterClusterContext }}'
+        state: present
+        namespace: "{{ prowNamespace }}"
+        definition:
+          apiVersion: v1
+          kind: Secret
+          metadata:
+            name: docker-mirror-proxy
+          type: Opaque
+          data:
+            ca.key: "{{ dockerMirrorProxyCA.key }}"
     - name: Deploy limit range
       command: "oc --kubeconfig {{ automationKubeconfig }} --context {{ masterClusterContext }} -n {{prowNamespace}} apply -f -"
       args:
--- a/github/ci/prow/templates/mirror-proxy-config.yaml
+++ b/github/ci/prow/templates/mirror-proxy-config.yaml
@@ -1,0 +1,7 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: mirror-proxy-config
+data:
+  ca.crt: |
+{{ dockerMirrorProxyCA.crt | indent(width=4, indentfirst=True) }}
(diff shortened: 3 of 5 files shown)


github/ci/prow/templates/mirror-proxy.yaml
```diff
@@ -0,0 +1,49 @@
...
+            memory: 3Gi
+          limits:
+            memory: 3Gi
+      volumes:
+      - name: ca-cert
+        secret:
+          secretName: docker-mirror-proxy
+      - name: storage
+        hostPath:
+          path: /var/lib/docker-mirror-proxy
```
rmohr, line 49: Can you use a sub-directory of where the docker mirror right now writes its data? We have there an extra partition just for this, so that we don't impact other applications if something goes wrong with the cache.

github/ci/prow/tasks/main.yml
```diff
@@ -138,6 +138,22 @@
...
+        namespace: "{{ prowNamespace }}"
+        definition:
+          apiVersion: v1
+          kind: Secret
+          metadata:
+            name: docker-mirror-proxy
+          type: Opaque
+          data:
+            ca.crt: "{{ dockerMirrorProxyCA.cert }}"
+            ca.key: "{{ dockerMirrorProxyCA.key }}"
```
rmohr, line 155: Could you create an extra config-map where you put ca.crt and the publick part of the ca into? Let's keep the key and the ca in an extra secret.

rmohr (changes requested):
Very good start.

github/ci/prow/tasks/main.yml
fgimenez: Sure done, now the config map embeds the certificate contents and is under templates while no variables are interpolated, let me know if we should use a separate file. I've also updated the secret PR to only include ca.key.

github/ci/prow/templates/mirror-proxy-service.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: docker-mirror-proxy
+spec:
+  selector:
+    app: docker-mirror-proxy
+  ports:
+  - port: 3128
```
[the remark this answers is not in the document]
fgimenez: The proxy uses a single port for both http and https, from the PoV of the daemons this will be transparent, they will still use the designated ports for connecting with the target registries, sorry but I don't see clearly what is the benefit of declaring these different ports.

github/ci/prow/files/config.yaml
```diff
@@ -353,6 +353,23 @@ presets:
   - name: docker-config
     mountPath: /etc/default
     readOnly: true
+- labels:
+    preset-docker-mirror-proxy: "true"
+  env:
+  - name: DOCKER_HTTP_PROXY
+    value: http://docker-mirror-proxy.kubevirt-prow.svc:3128
```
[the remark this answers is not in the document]
fgimenez: As I understand it nope, we should configure both `http_proxy` and `https_proxy` to the same URL using http protocol and the port on which the proxy is listening see https://github.com/rpardini/docker-registry-proxy#what and the suggested configuration in https://github.com/rpardini/docker-registry-proxy#configuring-the-docker-clients--kubernetes-nodes--linux-clients, the proxy acts as a man-in-the-middle creating an http tunnel. Tested and works fine :+1:

github/ci/prow/templates/mirror-proxy.yaml
fgimenez: Done the docker mirror is using `/var/lib/docker-mirror`, i've changed the proxy to use `/var/lib/docker-mirror/proxy`

github/ci/prow/files/config.yaml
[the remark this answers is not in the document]
rmohr: Perfect. Thanks for the details.

github/ci/prow/templates/mirror-proxy-service.yaml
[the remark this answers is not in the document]
rmohr: Makes sense to me now.

github/ci/prow/templates/mirror-proxy-config.yaml
```diff
@@ -0,0 +1,41 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: mirror-proxy-config
+data:
+  ca.srl: "0A"
+  ca.crt: |
```
rmohr, line 7: I am undecided if it is better to have it together with the rest, or just adding it here.

rmohr (commented):
/lgtm
/approve
/hold

Let's coordinate this with the secrets update.

github/ci/prow/files/config.yaml
```diff
@@ -353,6 +353,23 @@ presets:
...
+  - name: DOCKER_HTTPS_PROXY
+    value: http://docker-mirror-proxy.kubevirt-prow.svc:3128
+  - name: CA_CERT_FILE
+    value: /etc/docker-mirror-proxy/ca.crt
+  volumes:
+  - name: docker-mirror-proxy-ca-cert
+    secret:
+      secretName: docker-mirror-proxy
```
rmohr, line 368: This would probably have to be the configmap now.

rmohr (commented):
One small issue. The rest looks good.

github/ci/prow/files/config.yaml
fgimenez: Makes total sense, done.

PR #714, opened 2020-11-18
Add requirements file for ansible roles

fgimenez:
It includes doc to create a virtual env and install the requirements (including ansible) on it. Also sets the `ansible_python_interpreter` var to the right value so that the interpreter in the virtual env and the libraries installed there are used.


fgimenez:
`--system-site-packages` is added to the suggested command for creating the venv to prevent this error `Aborting, target uses selinux but python bindings (libselinux-python) aren't installed!`

dhiller (approved):
/lgtm
/approve

PR #717, opened 2020-11-18
Modify jobs to use docker mirror proxy

fgimenez:

### DIFF
--- a/github/ci/shared-deployments/files/docker-daemon-mirror-proxy.conf
+++ b/github/ci/shared-deployments/files/docker-daemon-mirror-proxy.conf
@@ -1,0 +1,6 @@
+DOCKER_OPTS="${DOCKER_OPTS} --data-root=/docker-graph"
+DOCKER_OPTS="${DOCKER_OPTS} --mtu=1450"
+DOCKER_OPTS="${DOCKER_OPTS} --ipv6"
+DOCKER_OPTS="${DOCKER_OPTS} --fixed-cidr-v6 2001:db8:1::/64"
+export http_proxy=http://docker-mirror-proxy.kubevirt-prow.svc:3128
+export https_proxy=http://docker-mirror-proxy.kubevirt-prow.svc:3128
--- a/github/ci/prow/files/config.yaml
+++ b/github/ci/prow/files/config.yaml
@@ -368,19 +368,24 @@ presets:
 - labels:
     preset-docker-mirror-proxy: "true"
   env:
-  - name: DOCKER_HTTP_PROXY
-    value: http://docker-mirror-proxy.kubevirt-prow.svc:3128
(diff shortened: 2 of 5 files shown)


github/ci/shared-deployments/files/docker-daemon-mirror-proxy.conf
```diff
@@ -0,0 +1,2 @@
+export http_proxy=http://docker-mirror-proxy.kubevirt-prow.svc:3128
+export https_proxy=http://docker-mirror-proxy.kubevirt-prow.svc:3128
```
rmohr, line 2: I think we also want to add:

```
DOCKER_OPTS="${DOCKER_OPTS} --data-root=/docker-graph"
DOCKER_OPTS="${DOCKER_OPTS} --mtu=1450"
DOCKER_OPTS="${DOCKER_OPTS} --ipv6"
DOCKER_OPTS="${DOCKER_OPTS} --fixed-cidr-v6 2001:db8:1::/64"
```

github/ci/prow/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
```diff
@@ -15,13 +15,13 @@ presubmits:
     max_concurrency: 11
     labels:
       preset-dind-enabled: "true"
-      preset-docker-mirror: "true"
+      preset-docker-mirror-proxy: "true"
```
rmohr, line 18: Can you revert that for now? We can trigger a job extra without having to modify it right now.

rmohr (changes requested):
Two small requests.

github/ci/prow/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
fgimenez: Sure, done.

github/ci/shared-deployments/files/docker-daemon-mirror-proxy.conf
fgimenez: We are currently merging `docker-daemon-mirror.conf` (which contains the options you mention) and `docker-daemon-mirror-proxy.conf` in the docker-proxy entry of the configmap, so that we don't need to repeat the settings in 2 different files https://github.com/kubevirt/project-infra/pull/717/files#diff-22cedcd0268e1984d6386801c0148fc967762c4e32bc56ab6d872a568122b11eR19 is that correct?

rmohr: ok, fine by me then. See it now.

rmohr (approved):
/lgtm
/approve
/hold

PR #720, opened 2020-11-19
Do not configure docker on bootstrap containers

fgimenez:
The configuration with all the required settings will be mounted on `/etc/default/docker`


rmohr:
/lgtm
/approve
/hold

Feel free to remove the hold when ready.

PR #721, opened 2020-11-19
local registry as --insecure-registry in proxy docker config

fgimenez:


fgimenez:
`docker-mirror.kubevirt-prow.svc` is not the local registry, it is the mirror and we are bypassing it when using the proxy, so no need to add it as an insecure registry.

PR #724, opened 2020-11-20
Include all the required modules in ansible requirements

fgimenez:
Instead of relying on virtual environments created with `--system-site-packages` (and depending on what is installed in the host) declare all the needed packages in requirements.txt.

cc/ @dhiller 


dhiller (approved):
/lgtm
/approve

PR #728, opened 2020-11-24
Gosec: always run subset of rules which where fixed

xpivarc:


github/ci/prow/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
```diff
@@ -371,6 +371,42 @@ presubmits:
             requests:
               memory: "4Gi"
 
+  - name: pull-kubevirt-gosec-subset
```
dhiller, line 374: So I understand we are creating another job so that we can run subsets. At some point in the future this job will then become obsolete? Then we'll remove this one and make the other one required right?

xpivarc: Yes, I want to keep the original lane so we can use it on PRs and see the converge to final switch.

xpivarc:
/hold 
Depends on https://github.com/kubevirt/kubevirt/pull/4589 and https://github.com/kubevirt/kubevirt/pull/4574

dhiller:
/lgtm
/approve

cancel the hold when ready

ezrasilvera:
@xpivarc is there a list defining the included tests somewhere ?  
Maybe we should have that as a file we can just PR and commit when we want to add/update new tests

github/ci/prow/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
```diff
@@ -371,6 +371,42 @@ presubmits:
...
+            - "-c"
+            - "cp /etc/bazel.bazelrc ./ci.bazelrc && make gosec"
+          # docker-in-docker needs privileged mode
+          securityContext:
+            privileged: true
+          resources:
+            requests:
+              memory: "4Gi"
+          env:
+            - name: GOSEC
```
xpivarc, line 407: @ezrasilvera  here we can add/remove the targets which we want to block on.

ezrasilvera: I understand, I thought it would be nice to have a text file with the test numbers. Maybe when you define the env variable you can read that file. Just a thought ...

github/ci/prow/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
```diff
@@ -371,6 +371,42 @@ presubmits:
...
+    decoration_config:
+      timeout: 1h
+      grace_period: 5m
+    labels:
+      preset-dind-enabled: "true"
+      preset-docker-mirror: "true"
+      preset-bazel-cache: "true"
+    spec:
+      containers:
+        - image: gcr.io/k8s-testimages/bootstrap:v20200430-be2a8a9
```
dhiller, line 394: You need to use our own bootstrap image now to use the new image pull-through cache
```suggestion
        - image: kubevirtci/bootstrap:v20201119-a5880e0
```

dhiller:
/lgtm cancel 

see comment at image

PR #731, opened 2020-11-25
Deploy docker proxy components in both clusters

fgimenez:
Moves the tasks and templates required for the docker proxy components from prow role to shared-deployments, so that they are deployed on both clusters.

cc/ @rmohr @dhiller 


github/ci/remote-cluster/templates/mirror-proxy.yaml
```diff
@@ -0,0 +1,51 @@
...
+  template:
+    metadata:
+      labels:
+        app: docker-mirror-proxy
+    spec:
+      terminationGracePeriodSeconds: 180
+      serviceAccountName: "caches"
+      containers:
+      - name: mirror-proxy
+        image: fgimenez/docker-registry-proxy:v0.2.0
```
dhiller, line 19: In the long run we should push that to kubevirtci org on quay also.

dhiller (approved):
/lgtm

github/ci/remote-cluster/templates/mirror-proxy.yaml
fgimenez: makes total sense :+1:

dhiller (approved):
/lgtm
/approve

dhiller (approved):
/lgtm

PR #735, opened 2020-11-25
SRIOV Lanes Jobs: Remove Rehearsal Plugin Labels

ormergi:
There are rehearsal plugin labels leftovers on sriov lane jobs
this PR removes them to prevent invoking the lane on any rehearsal plugin usage

### DIFF
--- a/github/ci/prow/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
+++ b/github/ci/prow/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
@@ -182,7 +182,6 @@ presubmits:
       preset-docker-mirror: "true"
       preset-shared-images: "true"
       sriov-pod: "true"
-      rehearsal.allowed: "true"
     spec:
       nodeSelector:
         hardwareSupport: sriov-nic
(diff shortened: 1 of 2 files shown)


qinqon:
Why we don't want to rehearsal it?

ormergi:
> Why we don't want to rehearsal it?

When you use rehearsal plugin on a PR it will run all the jobs that has the 'reahersal.allow=true' label.
I want to remove this label so it will not invoke sriov lane jobs when not needed

qinqon:
But if we change the job at project-infra we want to rehearsal them it right ?

qinqon:
> When you use rehearsal plugin on a PR it will run all the jobs that has the 'reahersal.allow=true' label.
I want to remove this label so it will not invoke sriov lane jobs when not needed

It reharse all the jobs **changed** in the PR, so there is no issue I think.

ormergi:
> > When you use rehearsal plugin on a PR it will run all the jobs that has the 'reahersal.allow=true' label.
> > I want to remove this label so it will not invoke sriov lane jobs when not needed
> 
> It reharse all the jobs **changed** in the PR, so there is no issue I think.

Thanks for the heads up, I guess I got it wrong

/close

PR #761, opened 2020-12-03
Increase lifecycle job timeout to 4h

qinqon:
Addind a new releas to CNAO reach the 3h limit since an upgrade of all
the releases is done, this PR increment the timeout to 4h to fix that.


github/ci/prow/files/jobs/kubevirt/cluster-network-addons-operator/cluster-network-addons-operator-presubmits.yaml
```diff
@@ -41,7 +41,7 @@ presubmits:
       optional: true
       decorate: true
       decoration_config:
-        timeout: 3h
+        timeout: 4h
```
RamLavi, line 44: why do we need it on pull-e2e-cluster-network-addons-operator-lifecycle-okd? I thought this lane is not running

PR #773, opened 2020-12-11
Update jobs to use docker proxy

fgimenez:

### DIFF
--- a/github/ci/kubernetes_crud/vars/main.yml
+++ b/github/ci/kubernetes_crud/vars/main.yml
@@ -1,0 +1,4 @@
+default_fatal_validation: yes
+default_strict_validation: yes
+
+default_crud_operation: apply_manifest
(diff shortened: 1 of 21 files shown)


fgimenez:
cc/ @rmohr @dhiller 

This is ready to go, the error in `pull-e2e-ovs-cni` is not related to the changes in this PR.

fgimenez:
all green now :broccoli:

github/ci/kubernetes_crud/vars/main.yml
```diff
@@ -0,0 +1,4 @@
+default_fatal_validation: yes
```
rmohr, line 1: is this intentional?

fgimenez: The file was lost during the changes required for the merge of this PR #608, I recovered it verbatim. It means that by default a failure validating k8s resources before applying them would fail the operation.

PR #777, opened 2020-12-14
Bump to a config-forker which creates unique job-names

rmohr:
The new referenced commit contains https://github.com/kubernetes/test-infra/pull/19009 which should make us more resilient to statusreconciler issues. Note that this PR got reverted later on due to some context expectations  in k8s release scripts in the later pipeline stages.


davidvossel (approved):
/lgtm
/approve

we'll see how it goes for jan release.

PR #781, opened 2020-12-15
Cleanup owners file

rmohr:
Remove some reviewer suggestions from people which are not or only
sporadically working on kubevirt at this point in time and move two
approvers to the emeritus section for the same reason.

This is needed to get more meaningful reviewers and approvers assigned.


fgimenez (approved):
/lgtm
/approve

PR #783, opened 2020-12-16
sriov cni, Allocate only one sriov PF for the target container

oshoval:
In order to be able to run multi sriov clusters on CI sriov node, at the same time,
each Job should get only one PF (per CNI call), and not all the available PFs.

**Notes**
Removed the resetting of the the dynamic VF num,
in case we use operator which will it it itself.
In case we want to be on the safe side, i can add it,
but it will need to be done with little refactoring comparing
to how it does it now, since the CNI is re-entrant,
and its forbidden to reset other container PFs / VFs.


oshoval:
@EdDev 
FYI

PR #784, opened 2020-12-16
Introduce sriov patch nodes script

oshoval:
The script is meant to create/remove/update extended resources
named `prow/sriov`, in order to control how many
jobs each sriov node can run simultaneously
according how many sriov PFs it has.

After creating the resources, they would appear as allocatable,
after around 30 seconds.

Verified that It survives node reboot.

### DIFF
--- a/cni-plugins/sriov-passthrough-cni/README.md
+++ b/cni-plugins/sriov-passthrough-cni/README.md
@@ -27,3 +27,43 @@ plugin on all the relevant K8s nodes.
 │   └── sriov-passthrough-cni
 └── README.md
 ```
+
+## Prow job usage:
+
+In order to use the CNI and allocate one SR-IOV PF:
+1. Add one of the following annotation to the prow job yaml (both are doing the same):
+
+    [1] `k8s.v1.cni.cncf.io/networks: multus-cni-ns/sriov-passthrough-cni`
+
+    [2] `k8s.v1.cni.cncf.io/networks: '[{"interface":"net1","name":"sriov-passthrough-cni","namespace":"multus-cni-ns"}]'`
+
+2. Check if `pull-kubevirt-e2e-kind-1.17-sriov` job at [3] has requests/limits of `prow/sriov`, if it does, make sure it request only 1 `prow/sriov`. see [4] for example.
+This will let k8s know that only one PF is allocated (each PF is represented by one `prow/sriov` resource).
+K8s uses these resources in order to know how many jobs can run simultaneously.
+Each SR-IOV node of the CI cluster has `prow/sriov` capacity according to the amount of it's available physical PFs.
+
+[3] https://github.com/kubevirt/project-infra/blob/master/github/ci/prow/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
+
+[4] 
+```bash
+requests:
+  prow/sriov: "1"
+limits:
+  prow/sriov: "1"
+```
+
+* In case 2 PFs are required, please do the following changes:
+
+1. Use the following annotation instead those in the previous [1] section.
+
+    `k8s.v1.cni.cncf.io/networks: '[{"interface":"net1","name":"sriov-passthrough-cni","namespace":"multus-cni-ns"}, {"interface":"net2","name":"sriov-passthrough-cni","namespace":"multus-cni-ns"}]'`
+
+2. Follow [2] of the previous section, but instead of 1 `prow/sriov`, use 2, in order to let k8s know that 2 PFs are
+allocated per job, each PF is represented by a `prow/sriov` resource.
+
+```bash
+requests:
+  prow/sriov: "2"
+limits:
+  prow/sriov: "2"
+```
--- a/hack/patch_node.sh
+++ b/hack/patch_node.sh
@@ -1,0 +1,84 @@
+#!/bin/bash -e
+
+# This script can be used in order to create/remove/update k8s extended resources named prow/sriov
+# see https://kubernetes.io/docs/tasks/administer-cluster/extended-resource-node for more details.
+# The resource will be created node wise, so each sriov node should be configured according
+# the desired capacity (according how many PFs it has, and how many jobs are actually desired to run simultaneously).
+# After creation, the resource would appear as allocatable on node's yaml after around 20 seconds.
+# The following command can be used in order to check it was created successfully.
+# timeout 60s bash -c "until oc get node $NODE -oyaml | grep prow/sriov | grep $CAPACITY | wc -l | grep 2; do sleep 1; done"
+
+# Usage:
+# ./patch_node.sh <NODE_NAME> <CAPACITY>
+
+PROXY_PORT=9999
+if which oc &> /dev/null; then
+  BINARY=oc
+elif which kubectl &> /dev/null; then
+  BINARY=kubectl
+else
+  echo "error: oc / kubectl not found"
+  exit 1
+fi
+
+function finish {
+  rc=$?
+  if jobs -p | xargs kill; then
+    echo -e "\nProxy deleted"
+  fi
+  exit "$rc"
+}
+
+function validate_parameters {
+  local node=$1
+  local capacity=$2
+  if ! $BINARY get node "$node" &> /dev/null; then
+    echo "error: node $node not found"
+    exit 1
+  fi
+
+  if [ -z "${capacity##*[!0-9]*}" ]; then
+    echo "error: capacity $capacity must be greater or equal 0"
+    exit 1
+  fi
+}
+
+function run_proxy {
+  $BINARY proxy -p $PROXY_PORT &
+  sleep 3
+  jobs 1 | grep -q Running
+}
+
+function patch_node {
+  local node=$1
+  local capacity=$2
+  if [ "$capacity" -ne 0 ]; then
+    curl --header "Content-Type: application/json-patch+json" \
+      --request PATCH \
+      --data '[{"op": "add", "path": "/status/capacity/prow~1sriov", "value": "'$capacity'"}]' \
+      http://localhost:$PROXY_PORT/api/v1/nodes/"$node"/status
+  else
+    curl --header "Content-Type: application/json-patch+json" \
+      --request PATCH \
+      --data '[{"op": "remove", "path": "/status/capacity/prow~1sriov"}]' \
+      http://localhost:$PROXY_PORT/api/v1/nodes/"$node"/status
+  fi
+}
+
+function main() {
+  local node=$1
+  local capacity=$2
+  if [ -z "$node" ] || [ -z "$capacity" ]; then
+    echo "syntax error, use: $0 <NODE_NAME> <CAPACITY>"
+    exit 1
+  fi
+
+  validate_parameters "$node" "$capacity"
+
+  trap finish EXIT
+  run_proxy
+
+  patch_node "$node" "$capacity"
+}
+
+main "$@"


oshoval:
@EdDev 
FYI

qinqon:
/hold
Add documentation here to unhold

oshoval:
/cc @qinqon @EdDev
Readme added.
`prow/sriov` extended resource PR is ready to review please.

oshoval:
Just squashed 2 commits.

oshoval:
/hold cancel
added documentation

cni-plugins/sriov-passthrough-cni/README.md
```diff
@@ -27,3 +27,37 @@ plugin on all the relevant K8s nodes.
...
+
+2. When the parallel sriov jobs mode is enabled, add the following requests/limits:
+```
+requests:
+  prow/sriov: "1"
+limits:
+  prow/sriov: "1"
+```
+
+### In order to run the CNI twice, so 2 sriov PFs would be allocated per job:
```
alonSadan, line 46: I think this headline, and the previous one should be about *what* will be detailed further on - allocate 2 (or 1) PF. The *how* should be in the description beneath  the headline.
something like: 

```
### Allocate 2 SR-IOV PFs per job:

One will have to run the cni twice in order for 2 SR-IOV PFs to be allocatable per job:
+1. USe the following ....
```

Strating a headline with "in order" seems weird. what do you think?

cni-plugins/sriov-passthrough-cni/README.md
```diff
@@ -27,3 +27,37 @@ plugin on all the relevant K8s nodes.
...
+1. Use the following annotation (added 2nd entry with just a different name for the interface),
```
alonSadan, line 47: "added" ---> "add  a"?

cni-plugins/sriov-passthrough-cni/README.md
```diff
@@ -27,3 +27,37 @@ plugin on all the relevant K8s nodes.
...
+must be in the same line, else the line will be deduped, as it has the same key.
```
alonSadan, line 48: interface),
must...

 ----> 

interface).
It must ...

cni-plugins/sriov-passthrough-cni/README.md
```diff
@@ -27,3 +27,37 @@ plugin on all the relevant K8s nodes.
...
+
+`k8s.v1.cni.cncf.io/networks: '[{"interface":"net1","name":"sriov-passthrough-cni","namespace":"multus-cni-ns"}, {"interface":"net2","name":"sriov-passthrough-cni","namespace":"multus-cni-ns"}]'`
+
+2. When the parallel sriov jobs mode is enabled, add the following requests/limits:
+update requests/limits accordingly, so the scheduler would preserve 2 prow/sriov resources for each job,
```
alonSadan, line 53: requests/limits: update ---> requests/limits. Update

cni-plugins/sriov-passthrough-cni/README.md
```diff
@@ -27,3 +27,37 @@ plugin on all the relevant K8s nodes.
...
+and obey the prow/sriov limit per node.
+```
+requests:
+  prow/sriov: "2"
+limits:
+  prow/sriov: "2"
+```
+
+Note:
+`sriov-pod: "true"` can be used as a safety catch if sriov jobs sequential mode is desired.
```
alonSadan, line 63: What do you mean?

Can you describe what exactly "safety catch" is and how to use `sriov-pod: "true"` as safety catch ?

hack/patch_node.sh
```diff
@@ -0,0 +1,60 @@
+#!/bin/bash -e
+
+PROXY_PORT=9999
+
+function finish {
+  rc=$?
+  if jobs 1 | grep -q Running; then
+    echo
```
alonSadan, line 8: Can you explain why this `echo` is needed?

hack/patch_node.sh
```diff
@@ -0,0 +1,60 @@
...
+    echo "Killing proxy (pid=$proxy_pid)..."
+    kill "$proxy_pid"
+  fi
+  exit "$rc"
+}
+
+# creates/removes prow/sriov resources for the given $node
```
alonSadan, line 15: Maybe it worth mentioning in the comment that this resource will actually effect the amount of PFs delegated to the cluster.

hack/patch_node.sh
```diff
@@ -0,0 +1,60 @@
...
+function patch_node {
+  local node=$1
+  local capacity=$2
```
alonSadan, line 18: The docs explain about allocate one or two jobs, here I see that the capacity can vary. 
Maybe we should mention in the doc that it can be more than 2, Or limit here the capacity to 2.

hack/patch_node.sh
```diff
@@ -0,0 +1,60 @@
...
+  if [ "$capacity" -ne 0 ]; then
```
alonSadan, line 19: Can someone pass  negative value to `capacity` ? 
Maybe change to `if [ "$capacity" -ge 0 ]; then`  ?

hack/patch_node.sh
```diff
@@ -0,0 +1,60 @@
...
+    curl --header "Content-Type: application/json-patch+json" \
+      --request PATCH \
+      --data '[{"op": "add", "path": "/status/capacity/prow~1sriov", "value": "'$capacity'"}]' \
+      http://localhost:$PROXY_PORT/api/v1/nodes/"$node"/status
+  else
+    curl --header "Content-Type: application/json-patch+json" \
+      --request PATCH \
+      --data '[{"op": "remove", "path": "/status/capacity/prow~1sriov"}]' \
```
alonSadan, line 27: maybe worth mentioning, for future vistors in this code, that the `~1` actually means `/` 
accordung to https://kubernetes.io/docs/tasks/administer-cluster/extended-resource-node/.
Maybe as an inline comment.

alonSadan (commented):
Thank you Or:)
I think it is great that you also explained the steps in the doc, and the script looks good to my unprofessional eyes. I added some notes, but most of them I style-related.

hack/patch_node.sh
oshoval: It doesn't affect the number of PFs delegated, it just affect how many jobs are running in the same time.
and this is determined becaused the request/limit is added to the job as explained in the README.
k8s will allocate the requested amount of prow/sriov, if it has enough free ones, else the job would be pending.

The CNI is responsible to allocate 1 PF, calling the CNI twice will allocate 2 PFs (explained in the README).

Here its only create the k8s resources.

hack/patch_node.sh
oshoval: Please see https://github.com/kubevirt/project-infra/pull/784#discussion_r568719214 (partially related).
the capacity here means only how many sriov/prow resources there will be,
the README just ask you that if you run the CNI twice, you need to request 2 resources out of the allocatable ones.
the capacity here is the configuration you apply to the node, based on how many PFs it physically have, and should not be limited.

hack/patch_node.sh
oshoval: negative values should not be used.
you meant `-gt` i guess, we can bu tthen we would need to protect against non integers, which -ne do protect them
and i wanted to keep it simple as was suggested on other PRs.

i can add something like in the main, wdyt?
```
  if [ -z "${capacity##*[!0-9]*}" ]; then
    echo "error: capacity $capacity must be greater or equal 0"
    exit 1
  fi
```

hack/patch_node.sh
oshoval: I dont think we want to keep comments ?
unless you insist, there is a comment already that say that the resource name will be prow/sriov
so it would be understandable that ~1 is the /

hack/patch_node.sh
oshoval: just to split between output of the previous stdout and this one, so the output will be nicer.

cni-plugins/sriov-passthrough-cni/README.md
```diff
@@ -27,3 +27,37 @@ plugin on all the relevant K8s nodes.
...
 ```
+
+## Prow job usage:
+
+### In order to use the CNI and allocate one sriov PF:
+1. Add the following annotation to the prow job yaml (or leave the current annotation unchanged).
+
+`k8s.v1.cni.cncf.io/networks: '[{"interface":"net1","name":"sriov-passthrough-cni","namespace":"multus-cni-ns"}`
+
+2. When the parallel sriov jobs mode is enabled, add the following requests/limits:
```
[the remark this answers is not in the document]
oshoval: Once we configure the prow jobs to use parallel mode, it means they will have a request/limits already,
and then in case one changes the number of times he calls the CNI, he will need to update the requests/limits accordingly.
It will appear here
https://github.com/kubevirt/project-infra/blob/6e55786ef91dce8f8719f63e5fac4cac13041b25/github/ci/prow/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml#L264

hack/patch_node.sh
alonSadan: ohhh. OK got it.
Thanks.

hack/patch_node.sh
alonSadan: If it is not limited to 2,  maybe you can generalize the explanation in the doc?

hack/patch_node.sh
alonSadan: > I dont think we want to keep comments ?
> unless you insist, 

> there is a comment already that say that the resource name will be prow/sriov

Maybe add it a short explanation there?

> so it would be understandable that ~1 is the /

I am not sure it will be understandable (also k8s people thought it could be confusing :) ) but I won't insist.

hack/patch_node.sh
alonSadan: I think it is better to play it safe.

hack/patch_node.sh
oshoval: The document doesn't explain the script, it explains only how to allocate more than one PF.

The README is part of this PR, just because the prow/sriov is introduced in this PR.
There is no other scenarios that we will need beside using one of 2 PFs, and even if there will be, one can derive them from the examples there.

I will add the the script itself some guidelines about the script usage itself if needed.

cni-plugins/sriov-passthrough-cni/README.md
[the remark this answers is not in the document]
alonSadan: Thanks for explaining. 
Maybe this explanation, or the summery of it, should appear also in the doc.

hack/patch_node.sh
alonSadan: OK. Thanks.

cni-plugins/sriov-passthrough-cni/README.md
oshoval: Removed it, its too much details, and not related the what the README should explain.

oshoval:
Addressed comments.

oshoval:
Fixed trap in case ctrl+c was used after proxy started but before trap was registered.

alonSadan (approved):
Thank you for adding the prolog for the script, it really make thing a lot clearer and, together with the changes you made in the doc,  addresses all of my previous  comments.

oshoval:
Thanks Alon,

/cc @fgimenez 
Would you take a look please?

hack/patch_node.sh
```diff
@@ -0,0 +1,84 @@
...
+# Usage:
+# ./patch_node.sh <NODE_NAME> <CAPACITY>
+
+PROXY_PORT=9999
+
+function finish {
+  rc=$?
+  if jobs 1 | grep -q Running; then
+    echo
+    echo "Killing proxy (pid=$proxy_pid)..."
```
fgimenez, line 20: This global variable hurts :) Could we do something like `jobs -p | xargs kill` to kill all the background jobs in the current shell?

fgimenez (commented):
Other than the `proxy_pid` global variable looks good.

hack/patch_node.sh
oshoval: I will verify that it doesn't touch jobs that were not created by the script, as if there are any,
the script shouldn't kill them.

fgimenez: Indeed, I think the command above is safe, `jobs` only shows processes started in the current shell.

oshoval: Right, done thanks

oshoval:
Addressed comments.

hack/patch_node.sh
```diff
@@ -0,0 +1,81 @@
...
+  if jobs -p | xargs kill; then
+    echo -e "\nProxy deleted"
+  fi
+  exit "$rc"
+}
+
+function validate_parameters {
+  local node=$1
+  local capacity=$2
+  if ! oc get node "$node" &> /dev/null; then
```
fgimenez, line 27: If we are planning to use this script in the new control plane and with the automated deployment tools we should use kubectl instead of  oc, is there any reason why you use oc here?

oshoval: Evgheni told me current CI cluster is using `oc`
maybe i will add auto detection according `which` in case needed ?
(tries to find oc, if not exists try to find kubectl, else fails)

fgimenez: Yes, autodetection sounds good :+1:

oshoval:
Addressed comments.

fgimenez (approved):
/lgtm
/approve

awesome thanks!

PR #796, opened 2020-12-24
Improve mirror download behaviour

rmohr:
* Check if we have to do an upload before opening a connection to the
   source.
 * Try to connect to the source with all available artifact urls and not
   just with the first one.


rmohr:
Merging to fix jobs and due to lack of available reviewers due to christmas. Worst case the mirror job will continue to fail. A local execution ran fine.

PR #797, opened 2020-12-24
Give kubevirtci bumpers explicit bump branch names

rmohr:
Ensure that every kubevirtci bump job has its own unique branch in
kubevirt-bot/kubevirti.


rmohr:
Merging to fix jobs and due to lack of available reviewers due to Christmas. Worst case the jobs will continue to create wrong PRs which they already do.

dhiller:
> Merging to fix jobs and due to lack of available reviewers due to Christmas. Worst case the jobs will continue to create wrong PRs which they already do.

Fair enough :)

PR #805, opened 2021-01-04
Release configs for kubevirt/kubevirt release branch release-0.37

davidvossel:
adds new release configs

### DIFF
--- a/github/ci/prow/files/jobs/kubevirt/kubevirt/kubevirt-presubmits-0.37.yaml
+++ b/github/ci/prow/files/jobs/kubevirt/kubevirt/kubevirt-presubmits-0.37.yaml
@@ -1,0 +1,595 @@
+periodics: null
+postsubmits: {}
+presubmits:
+  kubevirt/kubevirt:
+  - always_run: false
+    branches:
+    - release-0.37
+    decorate: true
+    decoration_config:
+      grace_period: 5m0s
+      timeout: 7h0m0s
+    labels:
+      preset-dind-enabled: "true"
+      preset-docker-mirror-proxy: "true"
+      preset-shared-images: "true"
+    max_concurrency: 11
+    name: pull-kubevirt-e2e-k8s-1.20
+    optional: true
+    skip_report: true
+    spec:
+      containers:
+      - command:
+        - /usr/local/bin/runner.sh
+        - /bin/sh
+        - -c
+        - export TARGET=k8s-1.20 && automation/test.sh
+        image: kubevirtci/bootstrap:v20201119-a5880e0
+        name: ""
+        resources:
+          requests:
+            memory: 29Gi
+        securityContext:
+          privileged: true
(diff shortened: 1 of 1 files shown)


rmohr:
/hold

This does not seem to take the new config-forker into account which should generate unique names.

github/ci/prow/files/jobs/kubevirt/kubevirt/kubevirt-presubmits-0.37.yaml
```diff
@@ -0,0 +1,595 @@
...
+      affinity:
+        podAntiAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+          - labelSelector:
+              matchExpressions:
+              - key: sriov-pod
+                operator: In
+                values:
+                - "true"
+            topologyKey: kubernetes.io/hostname
```
oshoval, line 208: can you please add 
```
            - labelSelector:
                matchExpressions:
                  - key: sriov-pod-multi
                    operator: In
                    values:
                      - "true"
              topologyKey: kubernetes.io/hostname
```
as https://github.com/kubevirt/project-infra/pull/843 introduced?
It will allow to run the multi sriov POC in parallel to the official jobs

This PR was generated manually or with some help tools ? 
(in case its with tools, i will update them if needed)

Thanks

oshoval (changes requested):

github/ci/prow/files/jobs/kubevirt/kubevirt/kubevirt-presubmits-0.37.yaml
rmohr: This PR gets generated when the release branch is cut and just takes the master configuration at this time.

rmohr: can you backport this please after this PR if you need it?

oshoval: sure thanks, will add it

btw, it means that in case the job is generated when the release is out, but then the master config is changed, we lose the changes (as this what happened now i think, since master config has it)

rmohr: > btw, it means that in case the job is generated when the release is out, but then the master config is changed, we lose the changes (as this what happened now i think, since master config has it)

Yes this is intentional. This ensures that old release branches continue working while new jobs can evolve without having to think about backward compatibility. Changes like yours, which you want to propagate back are very seldom compared to doing forward-looking-only changes.

rmohr:
/hold

I am still not convinced that we will not see the statusreconciler regression. Checking now.

rmohr:
/close

replaced with #872

PR #816, opened 2021-01-07
Add ci-search readme

fgimenez:
Besides $subject it tweaks the ci-search and cert-manager deployment and test jobs:
* tests are triggered on `WORKSPACE` changes too
* no need to extract kind config in tests, rely on the default config file.
* for deployments, use config file instead of env var.


dhiller (approved):
/lgtm
/approve

PR #819, opened 2021-01-08
WIP, sriov, side by side CNI with conf

oshoval:


oshoval:
conf is not needed, closing, the one without conf is working

PR #823, opened 2021-01-10
sriov cni, Add NetworkAttachmentDefinition missing namespace

oshoval:
Prow jobs use network multus-cni-ns/sriov-passthrough-cni. 
CI has the NAD deployed in multus-cni-ns as the jobs expect [1],
but the NAD manifest doesnt state which namespace to use, 
so the NAD will be created in the default cluster namespace,
and will be unreachable by the jobs.

[1]
```
#  oc get net-attach-def --all-namespaces
NAMESPACE           NAME                          AGE
multus-cni-ns       sriov-passthrough-cni         1y
```

### DIFF
--- a/cni-plugins/sriov-passthrough-cni/deploy/multus-cni.yaml
+++ b/cni-plugins/sriov-passthrough-cni/deploy/multus-cni.yaml
@@ -106,6 +106,7 @@ apiVersion: "k8s.cni.cncf.io/v1"
 kind: NetworkAttachmentDefinition
 metadata:
   name: sriov-passthrough-cni
+  namespace: multus-cni-ns
 spec:
   config: |
     {


cni-plugins/sriov-passthrough-cni/deploy/multus-cni.yaml
```diff
@@ -106,6 +106,7 @@ apiVersion: "k8s.cni.cncf.io/v1"
 kind: NetworkAttachmentDefinition
 metadata:
   name: sriov-passthrough-cni
+  namespace: multus-cni-ns
 spec:
   config: |
     {
```
oshoval, line 112: maybe need also the change next line to 0.3.1 as the cni binary itself outputs 0.3.1

EdDev (changes requested):
The change itself seems fine to me, but the commit itself is very confusing.
The terms are intermixed and it is hard to follow.

Please also make it clear that the correct namespace is applied in production and is just not present in the sources. The commit sentence that tries to express that is not clear.

oshoval:
Addressed comments

EdDev (approved):

oshoval:
Thanks Edy

oshoval:
/cc @qinqon 
Can you please take a look ?

oshoval:
Thanks Quique.

PR #824, opened 2021-01-11
sriov lanes, Add podAntiAffinity of sriov-pod-multi

oshoval:
There is an effort of creating a multi sriov POC.
In order to be able to run it on CI, with minimal interference to
the official jobs, this commit suggests adding a cross
`podAntiAffinity`, so official jobs won't be run when the POC
is running (manually triggered only for now), and vice versa.

This will allow us to continue testing the effort with minimal interruption.


oshoval:
Problems here are not due to this PR, but the testing on this PR reflected problems, i fixed already on the multi sriov POC.
(at least some of them)

PR #834, opened 2021-01-12
Add bazelisk to the bootstrap image

rmohr:
Ensure that bazelisk is present in the binary and that  the most
commonly used bazel version is precached. Finally create a symlink to
from bazelisk to bazel, which makes it easier for scripts to work with
plain bazel or bazelisk.


fgimenez (approved):
/lgtm
/approve

nice :)

PR #843, opened 2021-01-13
sriov lanes, Add podAntiAffinity of sriov-pod-multi

oshoval:
There is an effort of creating a multi sriov POC.
It will allow running multi sriov lane on the same node, simultaneously.

In order to be able to run it on CI, with minimal interference
to the official jobs, this commit suggests adding a new
`podAntiAffinity`, so official jobs won't be run when the POC
is running (manually triggered only for now), and vice versa.

For example, It will allow us to easily investigate the blocker of the POC,
which appears to be a real issue unrelated to sriov
https://github.com/kubevirt/kubevirt/issues/3886.
Running multi jobs at the same time, allows to simulate
this issue easier, so the investigation should be easier.

### DIFF
--- a/github/ci/prow/files/jobs/kubevirt/kubevirt/kubevirt-presubmits-0.20.yaml
+++ b/github/ci/prow/files/jobs/kubevirt/kubevirt/kubevirt-presubmits-0.20.yaml
@@ -384,6 +384,13 @@ presubmits:
                   values:
                   - "true"
               topologyKey: kubernetes.io/hostname
+            - labelSelector:
+                matchExpressions:
+                - key: sriov-pod-multi
+                  operator: In
+                  values:
+                  - "true"
+              topologyKey: kubernetes.io/hostname
       containers:
       - image: kubevirtci/bootstrap:v20201119-a5880e0
         command:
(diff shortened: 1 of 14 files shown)


oshoval:
We need to have the official jobs and the new jobs side by side for now, until the feature will be come public.
POC jobs must not run when official jobs are running and vice versa.
The feature allows running multi job on the same time, so it can't use `antiAffinity` internally between its jobs.

This is the reason i added the new `antiAffinity` label:
POC jobs will have `antiAffinity` labeled `sriov-pod: true`, as the official jobs - this way they will respect the official jobs.
POC jobs will also labeled `sriov-pod-multi: true`, so the official jobs will respect them when they are running.
POC jobs internally would respect each other using extended resources,
and the nodeSelector for the node that is tagged with the extended resources as you said.

Once the POC would finish, and the feature become public, we will be able to remove the new antiAffinity, and arrange
the yamls to the minimal needed.

Correct me please if there is a simpler way to cross respect between the POC and the official ones,
without breaking the feature which is running multi jobs on the same machine simultaneously.

fgimenez (approved):
/lgtm
/approve
/hold

Thanks for the explanation, feel free to unhold when ready

oshoval:
Thanks Federico

/hold cancel

PR #849, opened 2021-01-15
Update to set correct python image for user-guide builds

mazzystr:
Update to set correct python image for user-guide builds per Netlify limitation

@pep @dhiller


dhiller (approved):
/lgtm
/approve

PR #856, opened 2021-01-18
Ensure bash is used in user-guide jobs

vasiliy-ul:
'source' is a bash built-in command and it is not available in sh. It may lead to job failure if the interpreter does not point to bash:

    /bin/sh: 1: source: not found

Example of failing job in PR https://github.com/kubevirt/user-guide/pull/378 : https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_user-guide/378/user-guide-presubmit-build/1351077164841177088

### DIFF
--- a/github/ci/prow/files/jobs/kubevirt/user-guide/user-guide-periodics.yaml
+++ b/github/ci/prow/files/jobs/kubevirt/user-guide/user-guide-periodics.yaml
@@ -15,7 +15,7 @@ periodics:
         region: primary
       containers:
         - image: docker.io/library/python:3.7
-          command: ["/bin/sh", "-c"]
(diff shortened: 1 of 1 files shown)


rmohr:
/cc @mazzystr

seems like more things than just bash are not right yet.

mazzystr:
Thanks for catching that :)

vasiliy-ul:
> Thanks for catching that :)

:ok_hand: 

Still after switching to bash there is on more issue:
```
mkdir: cannot create directory ‘/opt/buildhome’: Permission denied
```
Any ideas about it?

vasiliy-ul:
Closing the PR as the jobs seem not used anymore (the respective files have been deleted in master).

PR #858, opened 2021-01-18
ci-search: set index-bucket to kubevirt-prow

fgimenez:

### DIFF
--- a/github/ci/services/ci-search/patches/statefulset.yaml
+++ b/github/ci/services/ci-search/patches/statefulset.yaml
@@ -26,6 +26,7 @@ spec:
         - --path=/var/lib/ci-search/
         - --deck-uri=https://prow.apps.ovirt.org/
         - --job-uri-prefix=https://prow.apps.ovirt.org/view/gs/
+        - --index-bucket=kubevirt-prow
         - --bugzilla-url=https://bugzilla.redhat.com/rest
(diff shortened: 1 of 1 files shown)


github/ci/services/ci-search/patches/statefulset.yaml
```diff
@@ -26,6 +26,7 @@ spec:
         - --path=/var/lib/ci-search/
         - --deck-uri=https://prow.apps.ovirt.org/
         - --job-uri-prefix=https://prow.apps.ovirt.org/view/gs/
+        - --index-bucket=kubevirt-prow
         - --bugzilla-url=https://bugzilla.redhat.com/rest
         - --bugzilla-search=OPEN version:4.7,4.6,4.5,4.4,4.3,4.2 product="Container Native Virtualization (CNV)" AND delta_ts>-2w
```
rmohr, line 31: Just realized that he versions here are wrong: We would probably have 1.4, 2,5 and 2.6

rmohr (changes requested):

github/ci/services/ci-search/patches/statefulset.yaml
fgimenez: Good catch! should be fixed now

rmohr (commented):
/lgtm
/approve

PR #870, opened 2021-01-20
Use a base image with the right config-forker

rmohr:
One commit landed which pulled in the right config-forker library to a
dockerfile, but this base image was then never for the build.

Follow up of #777.


davidvossel (approved):
/lgtm
/approve

PR #871, opened 2021-01-20
Periodically Check for a new Fedora version and create a new CDI

shwetaap:
Periodically Check for a new Fedora version and create a new Containerdiskimage for it in the registry

### DIFF
--- a/github/ci/prow/files/jobs/kubevirt/common-templates/common-templates-periodics.yaml
+++ b/github/ci/prow/files/jobs/kubevirt/common-templates/common-templates-periodics.yaml
@@ -1,0 +1,51 @@
+periodics:
+- name: periodic-common-templates-nightly-image-builder
+  cron: "*/30 * * * *"
+  annotations:
+    testgrid-create-test-group: "false"
+  decorate: true
+  decoration_config:
+    timeout: 1h
+    grace_period: 5m
+  max_concurrency: 1
+  labels:
+    preset-dind-enabled: "true"
+    preset-docker-mirror-proxy: "true"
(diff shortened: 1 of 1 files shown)


rmohr:
/cc @fgimenez 

If there are no annotations on the job, shoulld the testgrid-config pick this job up?

github/ci/prow/files/jobs/kubevirt/common-templates/common-templates-periodics.yaml
```diff
@@ -0,0 +1,48 @@
+periodics:
+- name: periodic-common-templates-nightly-image-builder
+  cron: "*/30 * * * *"
+  decorate: true
```
fgimenez, line 4:
```suggestion
  annotations:
    testgrid-create-test-group: "false" 
  decorate: true
```

fgimenez (changes requested):

fgimenez:
> If there are no annotations on the job, shoulld the testgrid-config pick this job up?

Good catch, periodics should either define the dashboard to which the job should be added or declare that we don't want it to be shown.  I've proposed a small change to not include this job in testgrid, it should fix https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_project-infra/871/pull-project-infra-check-testgrid-config/1352042776916660225 too.

fgimenez (approved):
/lgtm

rmohr:
/approve

the job is not ready. Appriving to allow @shwetaap iterating on it.

PR #875, opened 2021-01-21
Add sonar-scanner job

fgimenez:


rmohr:
Having the sonarqube code smells reported on branches sounds great to me. We need to enable our goveralls job which runs through bazel already. Why do we need another job for reporting unit tests somewhere?

fgimenez:
Makes total sense, also not sure if this concrete command would offer any information different from what we already have in the automatic analysis you mentioned.

/close

PR #877, opened 2021-01-21
user-guide: Fix venv path

mazzystr:
Another pull req to fix venv path.  Hopefully we can write to /tmp


mazzystr:
@dhiller If this doesn't fix the permissions issue then I'm out of ideas

PR #879, opened 2021-01-25
Minimalistic tide dashboard

rmohr:
Persist a minimalistic tide dashboard which looks like this:

tide-dashboard


fgimenez (approved):
/lgtm
/approve

awesome, thx!

PR #835, opened 2021-01-12
Run kubevirt e2e tests without the buider image to make them more lightweight

rmohr:
Change all main presubmits and postsubmits for kubevirt to the new fedora based bootstrap image. This new image allows running the e2e tests without nesting.

PR #894, opened 2021-01-28
A new fedora based bootstrap image

rmohr:
Pull in the test-infra boostrap image from https://github.com/kubernetes/test-infra/commit/7e74b33534f950170801fb3ce015c1e96c975471 and rework it to work with fedora.

The image is updated extremely seldom, so we don't miss much but can in contrast gain:

 * streight forward modifications (ca updates, adding bazelisk, ...)
 * much newer glibc versions than which exist in debian and the freedom to easier change these if needed

Unblocks #835 which needs new enough glibc versions.


rmohr:
/hold

Forgot some debian bits in the bazel cache file

fgimenez (approved):
/lgtm
/approve

thx!

PR #897, opened 2021-01-28
sriov, Change CNI annotation style

oshoval:
With this updated annotation style,
one can call the same CNI twice by duplicating the annotation,
(must be in the same annotation line) and changing only the interface name.
    
It will allows to call the updated CNI which allocate only one PF
twice, in order to allocate 2 PFs for migration.
Without it, duplicating the line, resulted in pod with just one annotation, as the
duplicated entry was deduped.

Example
`      k8s.v1.cni.cncf.io/networks: '[{"interface":"net1","name":"sriov-passthrough-cni","namespace":"multus-cni-ns"}, {"interface":"net2","name":"sriov-passthrough-cni","namespace":"multus-cni-ns"}]'`

### DIFF
--- a/github/ci/prow/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
+++ b/github/ci/prow/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
@@ -207,7 +207,7 @@ presubmits:
     - release-\d+\.\d+
     annotations:
       fork-per-release: "true"
-      k8s.v1.cni.cncf.io/networks: multus-cni-ns/sriov-passthrough-cni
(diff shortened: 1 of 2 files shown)


oshoval:
@qinqon 
As part of preparing the documention on how to use the new CNI, i updated the code in this PR
so it will be easier to adapt the changes later.
For some reason it doesn't run rehearse.
You can see here https://github.com/kubevirt/project-infra/pull/862/files#diff-52b69623cbee788fd0180870e9e649ba846c99c2855229706a8736e17848b8b7R210
That it did run rehearsal and worked with calling the CNI twice.

In this PR i just changed the annotation style, still using the same CNI.

PR #899, opened 2021-01-28
Add quay.io presets to the project-infra jobs

rmohr:
We are missing the quay preset in project-infra. It gets silently not injected, causing jobs to fail which try to upload something to quay.


dhiller (approved):
/lgtm
/approve

rmohr:
/close

I executed it wrong with mkpj.

PR #900, opened 2021-01-28
Don't use proxy for the bootstrap job for now

rmohr:
The proxy has issues with uploading big layers.


rmohr:
/cc @fgimenez 

Without the proxy the jobs succeeded: https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/logs/publish-bootstrap-image/1354792901300719616

fgimenez (approved):
/lgtm
/approve

thx, will take a look at the proxy

PR #904, opened 2021-01-28
Fix minor bootstrap issues

rmohr:
* Remove version dependent bazel configuration. We don't know the version upfront and all extra handling is for ancient bazel versions wich are not relevant for us
* Install sysctl and modprobe, which is needed for the ipv6-enabled docker startup


rmohr:
I mostly created this PR to fix the bazel config issue in my experimental jobs: https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/4914/pull-kubevirt-e2e-k8s-cnao-1.19/1354898596520005632

fgimenez (approved):
/lgtm
/approve

PR #907, opened 2021-01-29
Revert #871

rmohr:
#871 is not ready and produces a lot of noise in our monitoring channels.


fgimenez (approved):
/lgtm
/approve

@shwetaap let's catch up to try to make the job work

PR #915, opened 2021-02-01
handle new pr creator

davidvossel:
pr creator during our release process failed. The pr creator was updated recently and a parameter "match-title" was removed. This caused our automation to fail.


fgimenez (approved):
/lgtm
/approve
/hold

A side question, do you think it would make sense to have a presubmit prow job to run the release tool unit tests?

PR #918, opened 2021-02-01
use upstream docker-registry-proxy and set proxy-connect timeouts

fgimenez:
These changes will prevent errors like https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/logs/publish-bootstrap-image/1354780192060477440 when pushing big layers through our proxy. 

I've also taken the opportunity to switch to the upstream docker-registry-proxy image now that https://github.com/rpardini/docker-registry-proxy/issues/70 is fixed and our previous upstream PRs have already landed.

Finally, I've added the proxy deployment to the automated prow-deploy, was missing there. 

### DIFF
--- a/github/ci/prow-deploy/kustom/base/manifests/local/docker-mirror-proxy_deployment.yaml
+++ b/github/ci/prow-deploy/kustom/base/manifests/local/docker-mirror-proxy_deployment.yaml
@@ -76,5 +76,5 @@ spec:
         secret:
           secretName: docker-mirror-proxy
       - name: storage
-        hostPath:
-          path: /var/lib/docker-mirror/proxy
+        persistentVolumeClaim:
+          claimName: docker-mirror-proxy
--- a/github/ci/prow-deploy/kustom/base/manifests/local/docker-mirror-proxy_deployment.yaml
+++ b/github/ci/prow-deploy/kustom/base/manifests/local/docker-mirror-proxy_deployment.yaml
@@ -1,0 +1,80 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: docker-mirror-proxy
+spec:
+  selector:
+    matchLabels:
+      app: docker-mirror-proxy
+  replicas: 1
+  template:
+    metadata:
+      labels:
+        app: docker-mirror-proxy
+    spec:
+      terminationGracePeriodSeconds: 180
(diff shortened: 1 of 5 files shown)


github/ci/prow-deploy/kustom/base/manifests/local/docker-mirror-proxy_deployment.yaml
```diff
@@ -0,0 +1,80 @@
...
+            memory: 3Gi
+      volumes:
+      - name: ca-public
+        configMap:
+          name: mirror-proxy-config
+      - name: ca-private
+        secret:
+          secretName: docker-mirror-proxy
+      - name: storage
+        hostPath:
```
dhiller, line 79: Is it required for the proxy to use hostpath? I mean, if the node goes down our cache won't work any more? Or do we have more than one cache node?

dhiller (commented):
Looks good, one question though.

github/ci/prow-deploy/kustom/base/manifests/local/docker-mirror-proxy_deployment.yaml
fgimenez: You are completely right, when i created the docker-mirror-proxy deployment i just copied over the docker-mirror deployment, this `hostPath` setting was inherited from the old cluster constraints https://github.com/kubevirt/project-infra/blob/master/github/ci/prow/templates/mirror.yaml#L45-L47

I've just pushed changes to use a PVC instead, PTAL.

dhiller (approved):
/lgtm
/approve

github/ci/prow-deploy/kustom/base/manifests/local/docker-mirror-proxy_deployment.yaml
```diff
@@ -0,0 +1,80 @@
...
+        effect: "NoSchedule"
+      containers:
+      - name: mirror-proxy
+        image: rpardini/docker-registry-proxy:0.6.2
+        env:
+        - name: ENABLE_MANIFEST_CACHE
+          value: "true"
+        - name: ALLOW_PUSH
+          value: "true"
+        - name: PROXY_CONNECT_TIMEOUT
```
gw0, line 32: Just noticed, this is a duplicate env var. Also elsewhere in PR #918.

PR #920, opened 2021-02-02
Fix target calculation for kubevirt periodic master e2e

fgimenez:
There was an error in the way we calculated the target for the previous and previous-previous supported k8s versions, because of that all the kubevirt periodic master e2e were executed against the latest k8s version. 


dhiller (approved):
/lgtm
/approve

Thanks!

PR #931, opened 2021-02-04
Execute periodic e2e from master twice a day

fgimenez:
Now that we have verified that the periodic e2e lanes are working as expected let's increase the frequency to execute them twice a day, we have still margin thanks to the 7h timeout. 

Besides getting more feedback from them executing on working hours would run them on real conditions, the executions when no one else is using the cluster might give good results because of low resource usage. 


dhiller (approved):
/lgtm
/approve

PR #932, opened 2021-02-05
Add ci-health to prometheus scrape targets

fgimenez:
ci-health now publishes a metrics file in Prometheus format as part of the artifacts http://fgimenez.github.io/ci-health/output/metrics with these changes this file will be scraped by prometheus and we will be able to see the evolution of the metrics over time in grafana dashboards.


dhiller (commented):
/lgtm
/approve

PR #940, opened 2021-02-11
Add periodic job that syncs OpenSuse Crio repo

rhrazdil:
Add periodic job that downloads repositories for installing
cri-o version 1.20 and rsyncs the downloaded repository with gcs.

Since we can expect other versions to be needed later,
I'm adding `sync_crio_repo_for_version` so we can iterate
over a list of multiple versions when needed.

### DIFF
--- a/github/ci/prow/files/jobs/kubevirt/kubevirtci/kubevirtci-periodics.yaml
+++ b/github/ci/prow/files/jobs/kubevirt/kubevirtci/kubevirtci-periodics.yaml
@@ -1,0 +1,69 @@
+periodics:
+- name: periodic-kubevirtci-sync-crio-repository-weekly
+  cron: "25 0 * * 1"
+  decorate: true
+  annotations:
+      testgrid-create-test-group: "false"
+  decoration_config:
+    timeout: 1h
+    grace_period: 5m
+  max_concurrency: 1
+  labels:
+    preset-shared-images: "true"
+  spec:
+    containers:
+      - image: kubevirtci/bootstrap:v20201119-a5880e0
+        env:
+          - name: GOOGLE_APPLICATION_CREDENTIALS
+            value: /etc/gcs/service-account.json
+          - name: OS
+            value: CentOS_8_Stream
+          - name: BUCKET_DIR
+            value: kubevirtci-crio-mirror
+          - name: LOCAL_MIRROR_DIR
+            value: opensuse-mirror
+          - name: BASE_REPOID
+            value: devel_kubic_libcontainers_stable
+          - name: CRIO_VERSIONS
+            value: "1.20"
+        command:
+          - "/usr/local/bin/runner.sh"
+          - "/bin/sh"
+          - "-c"
+          - |
+            set -e;
+
+            sync_crio_repo_for_version () {
+              if [ -z "$1" ]
+              then
+                  CRIO_SUBDIR=""
+                  REPOID=$BASE_REPOID
+              else
+                  CRIO_SUBDIR=":cri-o:$1"
+                  REPOID=$BASE_REPOID_cri-o_$1
+              fi
+              curl -L -o $REPOID.repo https://download.opensuse.org/repositories/devel:kubic:libcontainers:stable$CRIO_SUBDIR/$OS/devel:kubic:libcontainers:stable$CRIO_SUBDIR.repo
+              reposync -c $REPOID.repo -p ./$LOCAL_MIRROR_DIR -n --repoid=$REPOID
+            }
+
+            # First sync the shared stable repository
+            sync_crio_repo_for_version
+
+            # Loop over comma-separated list of cri-o versions
+            for i in $(echo $CRIO_VERSIONS | sed "s/,/ /g")
+            do
+              sync_crio_repo_for_version $i
+            done
+
+            gsutil rsync -d -r $LOCAL_MIRROR_DIR gs://BUCKET_DIR
+        resources:
+          requests:
+            memory: "2Gi"
+        volumeMounts:
+          - name: gcs
+            mountPath: /etc/gcs
+            readOnly: false
+    volumes:
+      - name: gcs
+        secret:
+          secretName: gcs


github/ci/prow/files/jobs/kubevirt/kubevirtci/kubevirtci-periodics.yaml
```diff
@@ -0,0 +1,52 @@
+periodics:
+- name: periodic-kubevirtci-sync-crio-repository-weekly
+  interval: 168h
+  decorate: true
+  decoration_config:
+    timeout: 1h
+    grace_period: 5m
+  max_concurrency: 1
+  labels:
+    preset-docker-mirror-proxy: "true"
```
rhrazdil, line 10: Not sure if I need these labels or not

github/ci/prow/files/jobs/kubevirt/kubevirtci/kubevirtci-periodics.yaml
```diff
@@ -0,0 +1,52 @@
...
+    containers:
+      - image: kubevirtci/bootstrap:v20201119-a5880e0
+        env:
+          - name: GOOGLE_APPLICATION_CREDENTIALS
+            value: /etc/gcs/service-account.json
+        command:
+          - "/usr/local/bin/runner.sh"
+          - "/bin/sh"
+          - "-c"
+          - >
```
fgimenez, line 22: New lines would be removed with `>`
```suggestion
          - |
```

github/ci/prow/files/jobs/kubevirt/kubevirtci/kubevirtci-periodics.yaml
```diff
@@ -0,0 +1,52 @@
...
+            set -e;
+            BUCKET=kubevirtci-crio-mirror
+            LOCAL_MIRROR_DIR=opensuse-mirror
+            OS=CentOS_8_Stream
+            CRIO_VERSION=1.20
```
fgimenez, line 27: I think it would be better to define these env vars in the `env` block, wdyt?

github/ci/prow/files/jobs/kubevirt/kubevirtci/kubevirtci-periodics.yaml
```diff
@@ -0,0 +1,52 @@
...
+
+            DEVEL_LIBCONTAINERS_STABLE=devel_kubic_libcontainers_stable
+            curl -L -o $DEVEL_LIBCONTAINERS_STABLE.repo https://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/$OS/devel:kubic:libcontainers:stable.repo
+            reposync -c $DEVEL_LIBCONTAINERS_STABLE.repo -p ./$LOCAL_MIRROR_DIR -n --repoid=$DEVEL_LIBCONTAINERS_STABLE
```
fgimenez, line 31: There's repetition here, could you modify `sync_crio_repo_for_version` to take `DEVEL_LIBCONTAINERS_CRIO` as parameter? then call it twice, first with `devel_kubic_libcontainers_stable` and second with `devel_kubic_libcontainers_stable_cri-o_${CRIO_VERSION}`

github/ci/prow/files/jobs/kubevirt/kubevirtci/kubevirtci-periodics.yaml
```diff
@@ -0,0 +1,52 @@
...
+
+            sync_crio_repo_for_version () {
+              DEVEL_LIBCONTAINERS_STABLE_CRIO=devel_kubic_libcontainers_stable_cri-o_$1
+              curl -L -o DEVEL_LIBCONTAINERS_STABLE_CRIO.repo https://download.opensuse.org/repositories/devel:kubic:libcontainers:stable:cri-o:$1/$OS/devel:kubic:libcontainers:stable:cri-o:$1.repo
+              reposync -c DEVEL_LIBCONTAINERS_STABLE_CRIO.repo -p ./$LOCAL_MIRROR_DIR -n --repoid=DEVEL_LIBCONTAINERS_STABLE_CRIO
+            }
+            sync_crio_repo_for_version $CRIO_VERSION
+
+            gsutil mb gs://$BUCKET || true
```
fgimenez, line 40: I would remove this, we need to configure the bucket anyway to allow serve files with http from it.

fgimenez (changes requested):
Looks good, the only potential problem I see is what happens if we find the same connectivity issues with the upstream repo that we have during the test execution.

github/ci/prow/files/jobs/kubevirt/kubevirtci/kubevirtci-periodics.yaml
```diff
@@ -0,0 +1,52 @@
+periodics:
+- name: periodic-kubevirtci-sync-crio-repository-weekly
+  interval: 168h
+  decorate: true
+  decoration_config:
```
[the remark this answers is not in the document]
rhrazdil: Thanks!

github/ci/prow/files/jobs/kubevirt/kubevirtci/kubevirtci-periodics.yaml
```diff
@@ -0,0 +1,52 @@
+periodics:
+- name: periodic-kubevirtci-sync-crio-repository-weekly
+  interval: 168h
```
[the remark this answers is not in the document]
rhrazdil: Okay, done

github/ci/prow/files/jobs/kubevirt/kubevirtci/kubevirtci-periodics.yaml
rhrazdil: Thanks, changed

github/ci/prow/files/jobs/kubevirt/kubevirtci/kubevirtci-periodics.yaml
rhrazdil: Done

github/ci/prow/files/jobs/kubevirt/kubevirtci/kubevirtci-periodics.yaml
rhrazdil: Removed

github/ci/prow/files/jobs/kubevirt/kubevirtci/kubevirtci-periodics.yaml
rhrazdil: Updated, although in a slightly different way than you suggested, let me know what you think

fgimenez: lgtm :+1:

fgimenez (commented):
Nice thanks! 

I've tried to create and configure the `kubevirtci-crio-mirror` bucket so that it is writable with the credentials that will be used in the job and the contents readable publicly but looks like the bucket is already created, did you create it?

rhrazdil:
@fgimenez I'm not aware of creating it, afaik I don't have acces to do such thing. Unless the job has somehow been executed? But I don't se rehearse job running here.
Can you see when it was created?

fgimenez:
> Can you see when it was created?

Nope, the bucket doesn't belong to the kubevirt org in GCP, it has been created under a different owner (bucket names are unique across all the GCP infrastructure). The name is specific enough, it is weird that it is already created, maybe you did some local tests with your credentials and it is available under your user?

In any case we can use a different name, like `kubevirt-crio-mirror` wdyt?

rhrazdil:
> The name is specific enough, it is weird that it is already created, maybe you did some local tests with your credentials and it is available under your user?

@fgimenez Oh I didn't realize that it's unique across the entire GCP...   Yes that's exactly what happened, I have removed it

fgimenez:
@rhrazdil awesome thanks, I've already created and configured it to be publicly readable :+1:

fgimenez (approved):
/lgtm
/approve

github/ci/prow/files/jobs/kubevirt/kubevirtci/kubevirtci-periodics.yaml
```diff
@@ -0,0 +1,69 @@
...
+          - name: BASE_REPOID
+            value: devel_kubic_libcontainers_stable
+          - name: CRIO_VERSIONS
+            value: "1.20"
+        command:
+          - "/usr/local/bin/runner.sh"
+          - "/bin/sh"
+          - "-c"
+          - |
+            set -e;
```
rmohr, line 34: The job looks great, just one general note for the future: I think it is preferrable to keep all jobs, no matter if golang, bash, or whatever defined in the normal source code and not inside the job definitions.

The pattern is adding `project-infra` as `extraRef` to the job and defining it as the "current directory". Then one can simply invoke the script here and fixing scripts is still just a change on the PR and the next job run picks up the new version.

That is better maintain- and testable.

rhrazdil: I agree, thanks. I'll refactor the job as you suggest

PR #942, opened 2021-02-12
Fix missing reposync in kubevirtci crio sync periodic

rhrazdil:
bootstrap image doesn't have reposync, let's
fix that by installing it.


fgimenez (approved):
/lgtm
/approve

PR #947, opened 2021-02-12
Several grafana fixes

fgimenez:
c&p error, i noticed that the service was not accessible from https://grafana.ci.kubevirt.io/

/cc @dhiller @rmohr 

### DIFF
--- a/github/ci/services/prometheus-stack/manifests/grafana.yaml
+++ b/github/ci/services/prometheus-stack/manifests/grafana.yaml
@@ -141,7 +141,7 @@ data:
     provisioning = /etc/grafana/provisioning
     [auth.anonymous]
     enabled = true
-    org_name = KubeVirt community
+    org_name = Main Org.
 
   datasources.yaml: |
     apiVersion: 1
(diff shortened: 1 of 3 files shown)


fgimenez:
Several things are addressed in this PR:

* fixed deployment code to remove new lines from secrets
* fixed prometheus service URL in grafana data source definition
* fixed org_name entry in grafana.ini for anonymous auth 
* c&p error in the ingress service selector, i noticed that the service was not accessible from https://grafana.ci.kubevirt.io/

github/ci/services/prometheus-stack/manifests/grafana.yaml
```diff
@@ -141,7 +141,7 @@ data:
     provisioning = /etc/grafana/provisioning
     [auth.anonymous]
     enabled = true
-    org_name = KubeVirt community
+    org_name = Main Org.
```
dhiller, line 144: This looks odd?

dhiller (approved):
/lgtm
/approve
/hold

Besides the org_name looking odd?

github/ci/services/prometheus-stack/manifests/grafana.yaml
fgimenez: Yep, but it is the value of the default org created, without it there's an "unknown org" error in the logs and people not already logged get to the logging screen (not direct access given that they can be assigned to a default org). This is the same value that we use in the existing deployment https://github.com/kubevirt/project-infra/blob/master/github/ci/prometheus/templates/grafana-config.yaml#L16

PR #948, opened 2021-02-12
Use fedora image instead of debian and download metadata for crio mirror

rhrazdil:
The reposync version from debian yum-utils package
doesn't download metadata even when `--download-metadata`
parameter is specified.

Switching to fedora image instead and using dnf-utils.


fgimenez (approved):
/lgtm
/approve

PR #951, opened 2021-02-16
Periodically Check for a new Fedora version and create a new CDI

shwetaap:
Periodically Check for a new Fedora version and create a new CDI for it in the registry

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/common-templates/common-templates-periodics.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/common-templates/common-templates-periodics.yaml
@@ -1,0 +1,139 @@
+periodics:
+- name: periodic-common-templates-nightly-image-builder
+  cron: "30 2 * * *"
+  annotations:
+    testgrid-create-test-group: "false"
+  decorate: true
+  decoration_config:
+    timeout: 1h
+    grace_period: 5m
+  max_concurrency: 1
+  labels:
+    preset-dind-enabled: "true"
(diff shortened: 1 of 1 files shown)


fgimenez (approved):
/lgtm
/approve

github/ci/prow-deploy/files/jobs/kubevirt/common-templates/common-templates-periodics.yaml
```diff
@@ -0,0 +1,139 @@
...
+    - org: kubevirt
+      repo: common-templates
+      base_ref: containerdisk-automation
+  spec:
+    nodeSelector:
+      type: bare-metal-external
+    containers:
+    - image: kubevirtci/bootstrap:v20201119-a5880e0
+      command:
+        - "/usr/local/bin/runner.sh"
```
rmohr, line 29: @shwetaap @fgimenez sorry that I did not see that before, but I think the same applies like for https://github.com/kubevirt/project-infra/pull/1021#pullrequestreview-610127031.

@shwetaap I may have not been able to explain it clear enough, but this logic needs a place, a repo where it lives. The prowjob would only execute an entrypoint to that repo. Otherwise Maintenance is very hard.

Would you mind moving this logic to a repo and simplify this job?

rmohr: It should for instance be possible to simply run a script locally to test changes before pushing to CI. Yes we have CI to do stuff for us, but this pattern locks one into a very unproductive pattern.

PR #952, opened 2021-02-16
prow-deploy: add SRIOV node patch action

fgimenez:
prow-deploy task for patching SRIOV nodes using the `hack/patch_node.sh` script, includes check in smoke tests.  

For now it only defines SRIOV node entries in the testing configuration, when we use the deployment code in the new control plane we need to add the actual node names and capacities in the prod config.

/cc @oshoval @dhiller  

### DIFF
--- a/github/ci/prow-deploy/molecule/default/vars/kubevirtci-testing.yml
+++ b/github/ci/prow-deploy/molecule/default/vars/kubevirtci-testing.yml
@@ -6,3 +6,7 @@ kubeconfig_path: '/root/.kube/config'
 secrets_dir: '{{ project_infra_root }}/github/ci/prow-deploy/kustom/overlays/kubevirtci-testing/secrets/'
 
 kind_cluster_config: '/workspace/cluster.yaml'
+
+SRIOVNodes:
+- name: node01
+  capacity: 2
--- a/github/ci/prow-deploy/molecule/default/vars/kubevirtci-testing.yml
+++ b/github/ci/prow-deploy/molecule/default/vars/kubevirtci-testing.yml
@@ -7,6 +7,8 @@ secrets_dir: '{{ project_infra_root }}/github/ci/prow-deploy/kustom/overlays/kub
 
 kind_cluster_config: '/workspace/cluster.yaml'
 
+node_name: node01
+
 SRIOVNodes:
-- name: node01
+- name: '{{ node_name }}'
   capacity: 2
--- a/github/ci/prow-deploy/molecule/default/prepare.yml
+++ b/github/ci/prow-deploy/molecule/default/prepare.yml
@@ -48,7 +48,7 @@
             - |
               kind: JoinConfiguration
               nodeRegistration:
-                name: node01
+                name: '{{ node_name }}'
                 kubeletExtraArgs:
                   system-reserved: memory=16Gi
     - name: Launch cluster
--- a/github/ci/prow-deploy/molecule/default/smoke_tests.yml
+++ b/github/ci/prow-deploy/molecule/default/smoke_tests.yml
@@ -39,3 +39,10 @@
       --address="http://{{ deckUrl }}/hook" \
       --hmac {{ prowHmac }}
   changed_when: false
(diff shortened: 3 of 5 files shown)


github/ci/prow-deploy/molecule/default/vars/kubevirtci-testing.yml
```diff
@@ -6,3 +6,7 @@ kubeconfig_path: '/root/.kube/config'
 secrets_dir: '{{ project_infra_root }}/github/ci/prow-deploy/kustom/overlays/kubevirtci-testing/secrets/'
 
 kind_cluster_config: '/workspace/cluster.yaml'
+
+SRIOVNodes:
+- name: node01
```
oshoval, line 11: nit (not mandatory),
maybe better to call it node_name?

oshoval (approved):
Very nice, thanks

github/ci/prow-deploy/molecule/default/vars/kubevirtci-testing.yml
oshoval, line 11: can this be auto detected ? 
just in case we use the deploy with other provider, i.e kind-sriov
where the nodes have other name

Not crucial atm, up to you 
maybe can be also be a ENV var with a default value
so its easy to override

fgimenez: Given that the top-level element is called `SRIOVNodes` calling one of the child fields `node_*` would be redundant as I see it (given that all the children are nodes), wdyt?

oshoval: Np, understood,
just cause when it refereed its item.name
but all is good, thanks

fgimenez: > can this be auto detected ?

Not sure, can we determine from the node attributes if it should be patched? If so we would not need the static definition of vars at all.

> maybe can be also be a ENV var with a default value

But we need to define an array of nodes to be patched right? I would avoid env vars for that,let me know wdyt

oshoval: > > can this be auto detected ?
> 
> Not sure, can we determine from the node attributes if it should be patched? If so we would not need the static definition of vars at all.

Talking only on the smoke tests, and testing deploy, not about the IBM real one, which must have specific name + capacity tuples as you did.
Meant that we just need to patch one of the nodes, no matter which, for smoke tests.
If if someone will use the deploy script with other provider, this would fail, because the node name isn't node01.

> 
> > maybe can be also be a ENV var with a default value
> 
> But we need to define an array of nodes to be patched right? I would avoid env vars for that,let me know wdyt

On the IBM totally true, and there we dont need the above changes,
for the IBM you did totally fine, and it will have indeed specific node name + capacity,
as for example we have now srv17 - 4, srv20 - 2 and so on.

dhiller (approved):
/lgtm
/approve
/hold

Just remove when ready :)

github/ci/prow-deploy/molecule/default/vars/kubevirtci-testing.yml
oshoval: Feel free to disregard, in case someone will update the provider in the deploy (for testing locally) to other provider, he will be able to remove the check or adapt it for the right node name, thanks

oshoval (approved):
Thanks :)

/hold cancel

github/ci/prow-deploy/molecule/default/vars/kubevirtci-testing.yml
fgimenez: I see what you mean, in the testing environment we are creating the test cluster and we are actually defining the node name (we need to label it in a specific way so that some workloads are scheduled to it). I've modified the code a bit to have the node name as a variable and use it both for naming the node and for defining the name of the SRIOV node, PTAL.

PR #946, opened 2021-02-12
Add kuberhealthy deployment

fgimenez:
Test and deployment code for a basic [Kuberhealthy](https://github.com/Comcast/kuberhealthy) installation. Includes 3 preconfigured synthetic checks for damonset, deployment and dns status. Includes ingress for accessing the status page at https://status.ci.kubevirt.io

I've taken the opportunity to refactor the HTTP service checks in the `github/ci/services` deployments e2e tests.

/cc @rmohr @dhiller 

Signed-off-by: Federico Gimenez <fgimenez@redhat.com>

PR #955, opened 2021-02-16
Refactor secret management scripts to separate file

fgimenez:
Follow up of https://github.com/kubevirt/project-infra/pull/946#pullrequestreview-591107237 

/cc @dhiller 

### DIFF
--- a/hack/manage-secrets.sh
+++ b/hack/manage-secrets.sh
@@ -3,10 +3,8 @@
 set -euo pipefail
 
 decrypt_secrets(){
-    local github_token_path="${1}"
-
     target_dir=$(mktemp -d)
-    git clone https://kubevirt-bot:$(cat ${github_token_path})@github.com/kubevirt/secrets ${target_dir}
+    git clone https://kubevirt-bot@github.com/kubevirt/secrets ${target_dir}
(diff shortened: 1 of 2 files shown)


hack/manage-secrets.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+
+set -euo pipefail
+
+decrypt_secrets(){
+    local github_token_path="${1}"
+
+    target_dir=$(mktemp -d)
+    git clone https://kubevirt-bot:$(cat ${github_token_path})@github.com/kubevirt/secrets ${target_dir}
```
dhiller, line 9: Not sure if that makes sense here, but maybe we should use GIT_ASKPASS here instead of catting the token. In case one accidentally calls the function with `-x`, we then might leak the token here, might we not?

fgimenez: Absolutely, makes total sense. Already pushed the changes to use `GIT_ASKPASS`, this successful execution was run with them https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/logs/post-project-infra-kuberhealthy-deployment/1362077653179305984 PTAL

dhiller (approved):
/lgtm
/approve

PR #972, opened 2021-02-22
Reorg prow files and bump prow utility images

fgimenez:
The reorg aims to put the prow files under `prow-deploy` instead of current `prow` given the closer automated deployment. It consists of:
* Move `github/ci/prow/files` to `github/ci/prow-deploy/files` (replace original location with soft link)
* Link configuration files to `github/ci/prow-deploy/kustom/base/configs/current` 

The change of the utility images only affects prow-deploy code, current manual deployment works the same with the old configuration.

/cc @dhiller


fgimenez:
The changes are mostly about moving things around, not sure why gh things we are removing a lot of lines...

fgimenez:
> The changes are mostly about moving things around, not sure why gh thinks we are removing a lot of lines...

Now it is clear, there were duplicated config files under github/ci/prow-deploy/kustom/base/configs/20201021/ that had the same content as the config files in github/ci/prow/files now we keep a single copy.

dhiller (commented):
Looks good to me! Thanks for reducing the duplication. One question: I figure that the jobs in  `github/ci/prow/files/jobs/kubevirt/project-infra` that point to files in `github/ci/prow/files` will still work due to symlinks. It may make sense to point the jobs to the new location though. WDYT?

fgimenez:
> I figure that the jobs in `github/ci/prow/files/jobs/kubevirt/project-infra` that point to files in `github/ci/prow/files` will still work due to symlinks. It may make sense to point the jobs to the new location though. WDYT?

Currently all the actual job files are under `github/ci/prow-deploy/files/jobs`, the root folder being `prow-deploy` instead of `prow` given that we will soon start using the automated deployment code instead of the old one. We have a symlink from `github/ci/prow/files` to `github/ci/prow-deploy/files` and everything is in a single place, is this ok?

To keep the old deployment working (in case we need to deploy before the new clusters are in place) the new deployment code patches the old configuration as needed. For instance in this PR we add patch code to update the tags of the prow utility images to the latest version, this way the old deployment keeps working. Let me know if this makes sense or I'm missing your point :)

dhiller:
> > I figure that the jobs in `github/ci/prow/files/jobs/kubevirt/project-infra` that point to files in `github/ci/prow/files` will still work due to symlinks. It may make sense to point the jobs to the new location though. WDYT?
> 
> Currently all the actual job files are under `github/ci/prow-deploy/files/jobs`, the root folder being `prow-deploy` instead of `prow` given that we will soon start using the automated deployment code instead of the old one. We have a symlink from `github/ci/prow/files` to `github/ci/prow-deploy/files` and everything is in a single place, is this ok?
> 
> To keep the old deployment working (in case we need to deploy before the new clusters are in place) the new deployment code patches the old configuration as needed. For instance in this PR we add patch code to update the tags of the prow utility images to the latest version, this way the old deployment keeps working. Let me know if this makes sense or I'm missing your point :)

All this makes total sense, I figure I didn't explain correctly what I meant, i.e. [this config path](https://github.com/kubevirt/project-infra/blob/62f7f4488f0867cf5de88f8d4db98304d9c22211/github/ci/prow-deploy/files/jobs/kubevirt/project-infra/project-infra-presubmits.yaml#L15) points to the old path `prow` instead of `prow-deploy` and I was just thinking that we might change all the pathes that point to the old place in the project-infra jobs to the new `prow-deploy` location instead. Or am I overlooking something and this will not work?

fgimenez:
> I was just thinking we might change all the pathes in the project-infra jobs that point to the old place to the new `prow-deploy` location instead. Or am I overlooking something and this will not work?

Absolutely, makes total sense, thanks for bringing this up. I've already pushed the changes PTAL

dhiller (approved):
/lgtm

PR #977, opened 2021-02-24
Add sriov lane requirements

rmohr:
SRIOV lane installs `gettext` via `apt` and it fails on the bootstrap image due to the missing `ip` binary which is part of `iproute`.


dhiller (approved):
/lgtm
/approve

PR #979, opened 2021-02-25
Fix adding the latest kubevirtci provider

rmohr:
Ensure that we properly detect if the latest k8s version provider
already exists, and in that case only bump to the latest minor version
if necessary, instead of copying the n-1th release into a subdirectory
of the latest existing provider.

Fixes a wrong copy operation like in this PR: https://github.com/kubevirt/kubevirtci/pull/547


fgimenez (approved):
/lgtm
/approve

PR #985, opened 2021-03-01
Allow mixins for layering funcitonality on top of the bootstrap image

rmohr:
These mixin entrypoints allow layered containers to bring in additional setup and teardown code, avoiding the need to wrap the `runner.sh` script.

One example is for instance a layered golang container:

```
FROM quay.io/kubevirtci/bootstrap:latest

# Cache latest stable golang version
RUN curl -sL -o /bin/gimme https://raw.githubusercontent.com/travis-ci/gimme/master/gimme && chmod +x /bin/gimme
RUN gimme 1.15.8

RUN echo 'eval $(gimme 1.15.8)' > /etc/setup.mixin.d/golang.sh
```

Similar things can very often be achived by putting code into profile files or similar places, but they very often depend on the execution shell or other assumptions on how one enters a container. This way injection can be done in a well-defined way.

### DIFF
--- a/images/bootstrap/Dockerfile
+++ b/images/bootstrap/Dockerfile
@@ -99,6 +99,9 @@ RUN  curl -Lo ./bazelisk https://github.com/bazelbuild/bazelisk/releases/downloa
 RUN USE_BAZEL_VERSION=3.4.1 bazel version && \
   USE_BAZEL_VERSION=4.0.0 bazel version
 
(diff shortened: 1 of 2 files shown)


fgimenez (approved):
/lgtm
/approve
/hold

nice, feel free to unhold when ready

rmohr:
/hold cancel

verified with #986

PR #987, opened 2021-03-01
Pick up docker setup from CI

rmohr:
The default debian location is not by default picked up on fedora.
Changing that by explicitly sourcing it and taking DOCKER_OPTS into
account.


rmohr:
/approve

I verified that the  setup is picked up correctly

rmohr:
Jobs are green force-merge to restore CI with the container published in the postsubmit.

PR #989, opened 2021-03-02
New bootstrap with nested shell for http proxies

rmohr:
Our docker setup uses a http proxy. The http proxy is picked up via environment variables. Use a subshell to contain the proxy settings to the docker startup.


fgimenez (approved):
/lgtm
/approve

PR #991, opened 2021-03-02
Set real paths on config_updater settings and update symlinks

fgimenez:
/cc @rmohr @dhiller 

### DIFF
--- a/config
+++ b/config
@@ -1,0 +1,1 @@
+github/ci/prow-deploy/files/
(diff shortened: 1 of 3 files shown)


config
```diff
@@ -0,0 +1 @@
+github/ci/prow-deploy/files/
```
rmohr, line 1: I think that we should flip the order and move the real content to `config` and linke the other places. the same for `jobs`.

dhiller (approved):

PR #992, opened 2021-03-02
Request more memory on e2e jobs

rmohr:
Limit the amount of parallel runs per node a little bit more. We have too much pressure on the nodes. Limit the amount on parallel jobs runs per node slightly by requesting more memory.


davidvossel (commented):
/lgtm

fgimenez (approved):
/lgtm
/approve

dhiller (approved):
/approve

rmohr:

Nothing which should influence this jobs and the job is in the queue due to lack of resources (which we want to fix with this PR)

PR #994, opened 2021-03-03
Publish conformance yaml on developer builds of kubevirt

rmohr:
Publish conformance yaml to allow for instance for kubevirtci during cluster provision running latest conformance tests from kubevirt.


ormergi (approved):
Thank you!

PR #997, opened 2021-03-03
Remove travis branch protection from CDI

awels:
travis-ci.org no longer exists so the branch protection is no longer valid. The
tests performed by travis will be replaced with matching PROW jobs.


dhiller (approved):
/approve

PR #1000, opened 2021-03-04
Update dictionary file

mazzystr:
I don't know what happened with last nights pull req.  It was like I cntl-z'd and undid some stuff and didn't realize it.

Nevertheless.  Here is the rest of the changes to clean up user-guide `make check_spelling`


dhiller (approved):
/approve

PR #1002, opened 2021-03-05
Rollback to previous bootstrap image

fgimenez:
Aiming to remove the fedora image as a potential source of instability.

/cc @rmohr @dhiller 

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
@@ -24,7 +24,7 @@ presubmits:
       nodeSelector:
         type: bare-metal-external
       containers:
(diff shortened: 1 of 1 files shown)


dhiller (commented):
Could we maybe copy the image to quay to not run into the rate limit again?

fgimenez:
> Could we maybe copy the image to quay to not run into the rate limit again?

Makes total sense, but I'm not sure if rolling back the image is the right approach. We see now a lot of failures but nothing seems to point to the fedora image as the cause. I've just converted it to draft, will continue with it if we determine that this is going to help.

rmohr:
@fgimenez I think we can close this one now. We merged 6 PRs today: https://storage.googleapis.com/kubevirt-prow/reports/flakefinder/kubevirt/kubevirt/flakefinder-2021-03-08-024h.html until now.

What do you think?

fgimenez:
/close

makes sense :+1:

PR #851, opened 2021-01-17
WIP sriov, POC 1

oshoval:
Signed-off-by: Or Shoval <oshoval@redhat.com>

PR #1005, opened 2021-03-07
sriov, Update the style of the sriov CNI annotation

oshoval:
In order to support running the CNI twice once needed,
update the style of the undercloud sriov CNI so it would
be able to indicate two CNI instances should be run,
and as a result allocate two PFs (for migration).
Otherwise, calling the annotation twice using the old style,
would result in dedupe of the annotation and the
undercloud CNI would run only once.

There is no logical effect beside having it suitable for
running the CNI twice.
Tested on CI with one and two CNI calls.

See cni-plugins/sriov-passthrough-cni/README.md
for more info.

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits-0.30.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits-0.30.yaml
@@ -279,7 +279,7 @@ presubmits:
         type: bare-metal-external
   - always_run: false
     annotations:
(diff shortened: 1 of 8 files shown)


oshoval:
Tested here with 2 CNI calls
https://github.com/kubevirt/project-infra/pull/862

Tested here with 1 CNI and multi jobs
https://github.com/kubevirt/project-infra/pull/851

oshoval:
The annotation style of https://github.com/kubevirt/project-infra/pull/1013 also works
i will update the docs

not related to that, sometimes all the test fails, i believe its due to a bug with sriov-operator 4.4
once we ditch it (or in case it will be bumped in the future), it will solve it.

PR #1007, opened 2021-03-08
kubevirt-presubmits: Add sig-network k8s 1.19 job

oshoval:
This PR adds a new job `pull-kubevirt-e2e-k8s-1.19-sig-network` 

The purpose of this job is to allow running a more
focused group of tests, in isolation from the other tests,
owned by the sig-network.
    
It will run only the tests that are marked with sig-network,
including MacVtap and Multus (which require CNAO).

At the end of the effort, this new lane `pull-kubevirt-e2e-k8s-1.19-sig-network`
would always run, but won't be a gating.

The differences between this new lane and the official k8s-1.19
```
always_run: false
optional: true
```
The TARGET has suffix `-sig-network`

After testing we will make it `always_run` true,
since there is no rehearsal on k8s-1.19 we need this lane first merged.

Kubevirt counterpart https://github.com/kubevirt/kubevirt/pull/5167

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
@@ -113,6 +113,42 @@ presubmits:
           resources:
             requests:
               memory: "34Gi"
+  - name: pull-kubevirt-e2e-k8s-1.19-sig-network
+    skip_branches:
+      - release-\d+\.\d+
+    annotations:
+      fork-per-release: "true"
+      testgrid-dashboards: kubevirt-presubmits
+    always_run: false
+    optional: true
+    skip_report: false
+    decorate: true
+    decoration_config:
+      timeout: 7h
+      grace_period: 5m
+    max_concurrency: 11
+    labels:
+      preset-dind-enabled: "true"
+      preset-docker-mirror-proxy: "true"
+      preset-shared-images: "true"
+      preset-bazel-cache: "true"
+      preset-bazel-unnested: "true"
+    spec:
+      nodeSelector:
+        type: bare-metal-external
+      containers:
+        - image: quay.io/kubevirtci/bootstrap:v20210302-4c35b4e
+          command:
+            - "/usr/local/bin/runner.sh"
+            - "/bin/sh"
+            - "-c"
+            - "export TARGET=k8s-1.19-sig-network && automation/test.sh"
+          # docker-in-docker needs privileged mode
+          securityContext:
+            privileged: true
+          resources:
+            requests:
+              memory: "34Gi"
   - name: pull-kubevirt-e2e-k8s-1.18
     skip_branches:
       - release-\d+\.\d+


oshoval:
/cc @phoracek 
/cc @fgimenez
Can you take a look please?
merging this lane will allow me to test manually the sig network filter on CI
the new lane first will run only manually, and after i verify the logic we will make it 
always run but not mandatory

fgimenez:
Looks good but let's clear the merge queue first before putting more pressure on CI resources, even with manually triggered builds.

oshoval:
Just updated commit message

EdDev (changes requested):
>  sig-network, Add sig-network k8s 1.19 lane

kubevirt-presubmits: Add sig-network k8s 1.19 job

> 
> This lane would always run, but won't be a gating.
> It will run only the tests that are marked with sig-network
> (not including MacVtap and Multus since they need CNAO).

They should run everything, so I suggest you copy the CNAO job which deploys it as well.

> 
> The purpose of this lane is to check the flakiness of network,

This is not the reasoning for the sig-network lane.
It is proposed in order to allow a more focused group of tests to run in isolation from all the others, owned by the sig-network.

> and allow to run only the network tests, as part of the
> effort to divide CI to separate lanes according their
> topic.

There is not such formal effort defined yet.

Please place the reasoning first in the commit message.

oshoval:
> > This lane would always run, but won't be a gating.
> > It will run only the tests that are marked with sig-network
> > (not including MacVtap and Multus since they need CNAO).
> 
> They should run everything, so I suggest you copy the CNAO job which deploys it as well.

Updated https://github.com/kubevirt/kubevirt/pull/5167 to install CNAO in case `-sig-network` is included,
marked all the families that CNAO runs "Multus|Networking|VMIlifecycle|Expose|Macvtap" with Sig-Network as well,
this way `sig-network`job would run all what CNAO runs + the Sig-Network that exists and are not CNAO exclusive (since marked them as well with `Sig-Network`)
let me know please what you think, the PR is still draft because i am testing it locally (and will need to test it on CI as well),
but we can continue this discussio either here or there, up to you
Thanks

oshoval:
Addressed comments

github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
```diff
@@ -113,6 +113,42 @@ presubmits:
...
+      - release-\d+\.\d+
+    annotations:
+      fork-per-release: "true"
+      testgrid-dashboards: kubevirt-presubmits
+    always_run: false
+    optional: true
+    skip_report: false
+    decorate: true
+    decoration_config:
+      timeout: 7h
```
EdDev, line 127: How about changing this to 4h? I would even go lower, but we need to collect data for that.

EdDev (commented):
Please consider reducing the timeout.
Other than that, it LGTM.

github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
oshoval: i would suggest to first collect the data and then lets reduce ?

EdDev: Fine, but please track this one.

EdDev (approved):
Thank you, looks correct to me.

I think we first need https://github.com/kubevirt/kubevirt/pull/5167 to be merged and only then take this in.

oshoval:
> Thank you, looks correct to me.
> 
> I think we first need [kubevirt/kubevirt#5167](https://github.com/kubevirt/kubevirt/pull/5167) to be merged and only then take this in.

Thanks

The opposite would allow me to test 5167 on CI, i think we should do the oppsite?
Since its manual running, i don't see why we should block this PR due to the queue,
i will run it just once, and can even run it with dry run in the first runs which is much faster

fgimenez (approved):
/lgtm
/approve

I think we need to wait until we have migrated to the new infra and have more resources in place for actually enabling it, good for manually triggering and testing now that the queue have started moving.

rmohr:
> Can you take a look please?
> merging this lane will allow me to test manually the sig network filter on CI
> the new lane first will run only manually, and after i verify the logic we will make it
> always run but not mandatory

PR #1009, opened 2021-03-08
Reduce the amount of required jobs as emergency measure

rmohr:
The following lanes are affected:

 * k8s-1.17-rook is now set to optional and not always_run
 * k8s-1.17 is now set to optional and not always_run
 * k8s-1.20 is now set to not always_run

Most notably the following two main lanes stay enabled (together with
other smaller lanes):

 * k8s-1.18
 * k8s-1.19

This step should be revised and/or reverted within one week (Mar 15th).

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
@@ -6,7 +6,7 @@ presubmits:
     annotations:
       fork-per-release: "true"
       testgrid-dashboards: kubevirt-presubmits
-    always_run: true
+    always_run: false
     optional: true
     skip_report: true
     decorate: true
@@ -196,8 +196,8 @@ presubmits:
     annotations:
       fork-per-release: "true"
       testgrid-dashboards: kubevirt-presubmits
-    always_run: true
-    optional: false
+    always_run: false
+    optional: true
     skip_report: false
     decorate: true
     decoration_config:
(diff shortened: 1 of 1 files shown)


rmohr:
/hold

Let's collect a few lgtm

fgimenez (approved):
/lgtm

dhiller (approved):

github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
```diff
@@ -591,8 +591,8 @@ presubmits:
     annotations:
       fork-per-release: "true"
       testgrid-dashboards: kubevirt-presubmits
-    always_run: true
-    optional: false
+    always_run: false
```
xpivarc, line 594: Do we have another ceph lane?

rmohr: No, not at the moment.

fgimenez (commented):
/approve

EdDev (approved):

fgimenez:
I think we should also put `always_run: false` on `pull-kubevirt-e2e-kind-1.17-sriov-quarantine` and `pull-kubevirt-e2e-1.17-quarantine`

fgimenez:
> @fgimenez definitely. Would you mind disabling all of them? They create a lot of red entries without benefits right now. In the current situation I would prefer to reduce the noise as much as possible.

Sure proposed here https://github.com/kubevirt/project-infra/pull/1011

PR #1010, opened 2021-03-08
For HCO - skip if code/build/test wasn't changed

nunnatsa:
Don't run HCO test if only doc changed.


dhiller (approved):
/approve

PR #1011, opened 2021-03-09
Disable quarantine lanes

fgimenez:


rmohr:
With other words, they were added and enabled too fast.

fgimenez:
Sure, the automatic execution is disabled until the queue is empty, the worker nodes have a lot of pressure. Once the queue is empty the quarantine lanes will be enabled again, we can trigger them manually meanwhile.

fgimenez:
We also have a bare metal down since yesterday, which makes the pending job queue longer.

EdDev:
OK, I think I get it.
Well, they do something: Show us the problems.
But we can just move them to a periodic run, which seems more appropriate.

enp0s3 (approved):
/approve

PR #1012, opened 2021-03-09
Revert disabling rook-ceph

rmohr:
Re-enavbling ceph due to mailing list feedback.

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
@@ -591,8 +591,8 @@ presubmits:
     annotations:
       fork-per-release: "true"
       testgrid-dashboards: kubevirt-presubmits
-    always_run: false
-    optional: true
+    always_run: true
+    optional: false
     skip_report: false
     decorate: true
     decoration_config:


rmohr:
/approve
/hold

@awels feel free to lgtm and cancel the hold whenever you see fit.

davidvossel (approved):
/lgtm
/approve

davidvossel (commented):
/hold

unhold when you all are ready

EdDev:
But this is a 1.17 lane, we know it is problematic and will fail frequently. Has something changed?
Was this job less prone to failures?

awels:
On Tue, Mar 9, 2021 at 8:29 AM Edward Haas <notifications@github.com> wrote:

> But this is a 1.17 lane, we know it is problematic and will fail
> frequently. Has something changed?
>
Was this job less prone to failures?
>

Its the 1.17 rook-ceph lane, this is a special lane that runs a limited
number of tests, in particular related to storage operations available in
KubeVirt like snapshot and block device operations. It is actually very
stable except for one test we have identified.

awels:
If someone can ack https://github.com/kubevirt/kubevirt/pull/5178 and guide it in, then we can re-enable to rook-ceph lane. I have had 3 continuous runs without failures with the fix applied.

awels:
/hold cancel
Fix was merged, lane should be relatively stable at this point.

PR #1017, opened 2021-03-10
Trying again to require HCO sanity test on push to master

nunnatsa:
Trying to use the exact same text as apears in the lane:

`Sanity Checks / Sanity Checks (pull_request)`

### DIFF
--- a/github/ci/prow-deploy/kustom/base/configs/current/config/config.yaml
+++ b/github/ci/prow-deploy/kustom/base/configs/current/config/config.yaml
@@ -277,7 +277,6 @@ branch-protection:
               required_status_checks:
                 contexts:
                   - coverage/coveralls
-                  - "Sanity Checks / Sanity Checks (pull_request)"
             release-1.2:
               protect: true
               required_status_checks:
--- a/github/ci/prow-deploy/kustom/base/configs/current/config/config.yaml
+++ b/github/ci/prow-deploy/kustom/base/configs/current/config/config.yaml
@@ -277,7 +277,7 @@ branch-protection:
               required_status_checks:
                 contexts:
                   - coverage/coveralls
-                  - "Sanity Checks / Sanity Checks"
+                  - "Sanity Checks / Sanity Checks (pull_request)"
             release-1.2:
               protect: true
               required_status_checks:


tiraboschi (approved):

github/ci/prow-deploy/kustom/base/configs/current/config/config.yaml
```diff
@@ -277,7 +277,7 @@ branch-protection:
               required_status_checks:
                 contexts:
                   - coverage/coveralls
-                  - "Sanity Checks / Sanity Checks"
+                  - "Sanity Checks / Sanity Checks (pull_request)"
```
dhiller, line 280: I don't think this will work. I had a look at the status json that is returned by the api (I use gh cli for that). I figure in order to make a status required the context name needs to match. This is the case for coveralls (see line 279)

However, for the action result no status is attached to the commit directly:

```
$ gh api repos/kubevirt/hyperconverged-cluster-operator/commits/26f76a42a953bac5fcf539d63f87f97df1ece9b2/status | jq '.statuses[].context'
"coverage/coveralls"
"branch-ci-kubevirt-hyperconverged-cluster-operator-master-okd-images"
"branch-ci-kubevirt-hyperconverged-cluster-operator-master-images"
"branch-ci-kubevirt-hyperconverged-cluster-operator-release-4.9-images"
"branch-ci-kubevirt-hyperconverged-cluster-operator-release-4.8-images"
```

I figure to make this one required you need to attach a new status via api to the commit and then by adding that context name here you can make it required. Makes sense?

dhiller: Maybe this one can help: https://github.com/andarist/create-commit-status-action

tiraboschi (approved):

github/ci/prow-deploy/kustom/base/configs/current/config/config.yaml
nunnatsa: Checked that and other similar implementations. This approach can't be used when creating PRs from forks, as we do. The secrects are not available from fork.

PR #1020, opened 2021-03-10
Set the bazel output base directory to a well known location

rmohr:
Ensure that we know which directory bazel will use as a cache
directories, independent of the user used on the prow job.

Further mounting an emptyDir at this well known location if jobs decide
to run bazel unnested.

This way bazel will not write massively into the container overlay,
which can fill up the container space pretty fast and which can also
perform bad.

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presets.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presets.yaml
@@ -15,3 +15,9 @@ presets:
   env:
     - name: KUBEVIRT_RUN_UNNESTED
       value: "false"
+  volumes:
+  - name: bazeluserroot
+    emptyDir: {}
+  volumeMounts:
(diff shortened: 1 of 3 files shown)


rmohr:
bazelrc file when the cache is enabled:

```
# cat /etc/bazel.bazelrc 
startup --output_user_root=/tmp/cache/bazel
startup --host_jvm_args=-Dbazel.DigestFunction=sha256
build --remote_local_fallback
build --remote_http_cache=http://bazel-cache.kubevirt-prow.svc.cluster.local:8080///workspace,f107c644ec3d43330ad677d00bcd326f
```

bazelrc when the cache is not requested:

```
# cat /etc/bazel.bazelrc 
startup --output_user_root=/tmp/cache/bazel
```

fgimenez (approved):
/lgtm
/approve
/hold

Great thx, feel free to unhold when ready.

rmohr:
@fgimenez improved the weird wording on one of the comments, should be good otherwise. Would need another lgtm.

PR #1021, opened 2021-03-11
Add kubevirt/kubevirt job to kubevirtci

qinqon:
Is very difficult to test kubevirtci PRs at kubevirt, this change add a
job for k8s-1.18 to test kubevirt at a kubevirtci PR.

From PR https://github.com/kubevirt/kubevirtci/pull/565 tested at job https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirtci/565/check-kubevirt-e2e-k8s-1.18/1370022818195968000 

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/kubevirtci/kubevirtci-presubmits.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/kubevirtci/kubevirtci-presubmits.yaml
@@ -106,7 +106,55 @@ presubmits:
         resources:
           requests:
             memory: "8Gi"
-
+  - name: check-kubevirt-e2e-k8s-1.18
+    always_run: false
+    optional: true
+    decorate: true
+    skip_report: false
+    extra_refs:
+      - org: kubevirt
+        repo: kubevirt
+        base_ref: master
+    decoration_config:
+      timeout: 7h
+      grace_period: 5m
+    max_concurrency: 1
+    labels:
+      preset-dind-enabled: "true"
+      preset-docker-mirror-proxy: "true"
+      preset-shared-images: "true"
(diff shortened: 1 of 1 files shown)


fgimenez (approved):
/lgtm

github/ci/prow-deploy/files/jobs/kubevirt/kubevirtci/kubevirtci-presubmits.yaml
```diff
@@ -106,7 +106,55 @@ presubmits:
...
+      preset-shared-images: "true"
+      preset-bazel-cache: "true"
+      preset-bazel-unnested: "true"
+    spec:
+      nodeSelector:
+        type: bare-metal-external
+      containers:
+        - image: gcr.io/k8s-testimages/kubekins-e2e@sha256:8c80c98d035a0f285ad49b964e5d0b62004844f3340407367841618e80e1503c
+          command:
+            - "/usr/local/bin/runner.sh"
```
rmohr, line 134: We can not accept jobs anymore which do their stuff in the command section here. Jobs need a repo where the logic is placed, so that all changes can be evolved together.

rmohr (commented):
@qinqon would you mind reverting this PR?

I think what you actually want is a script in kubevirtci which does something like this, which you would just call like conformance tests as part of the job, if you like.

This would be another job which we have to somehow maintain and automate updates.

rmohr:
@dhiller would you mind assisting Felix on how to run the full set of e2e tests via the developer builds, like on openshift ci?

rmohr:
> @qinqon would you mind reverting this PR?

At least keep it only temporary, since I can understand why you need it right now. :)

PR #1016, opened 2021-03-09
Use bazel to build rehearse image

fgimenez:
Will fix failures like https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/logs/publish-rehearse-image/1368982337546620929

/cc @dhiller

Signed-off-by: Federico Gimenez <fgimenez@redhat.com>

PR #1022, opened 2021-03-11
Pull in the unversioned python binary explicitly

rmohr:
Some dependencies either changed in the bootstrap image or something
changed in fedora.

This ensures that we always have the python binary present.

Without this explicit `/usr/bin/python` binary present, bazel can not detect python and fails with something like this:

```
ERROR: no such package '@pip_deps//': pip_import failed:  (src/main/tools/process-wrapper-legacy.cc:75: "execvp(python, ...)": No such file or directory
```


fgimenez (approved):
/lgtm
/approve

thanks! i'll update #1016 as soon as the image is built to use it :)

PR #1024, opened 2021-03-11
Move all presubmits to config.yaml

rmohr:
Moves all presubmits to the main confuguration file. Overlaps are not fixed.

Nothing prevents jobs from repo X/Y to use presets which are defined in the directories of repo A/B. This makes it hard to trace what exactly is happening.

Moving all presets to a common location for later cleanup.


fgimenez (approved):
/lgtm
/approve

PR #1027, opened 2021-03-12
Reenable unnested bazel builds for e2e tests in kubevirt

rmohr:
Last time bazel was writing into the container overlay, which cause unexpected disk pressure on nodes.
This time bazel writes to a well define emptyDir which is mounted via the bazel preset.

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/containerized-data-importer/containerized-data-importer-presubmits.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/containerized-data-importer/containerized-data-importer-presubmits.yaml
@@ -243,7 +243,7 @@ presubmits:
       preset-bazel-cache: "true"
     spec:
       containers:
-        - image: quay.io/kubevirtci/bootstrap:v20210302-4c35b4e
+        - image: quay.io/kubevirtci/bootstrap:v20210311-09ebaa2
           command:
             - "/usr/local/bin/runner.sh"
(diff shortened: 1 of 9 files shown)


rmohr:
A successful job run: https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/4914/pull-kubevirt-e2e-k8s-1.19-custom/1370055531024093184, executed at ibm-srv04.ovirt.org.

The graphana dashboard shows that the amount of free storage for container images goes down by roughly two GB because of the new bootstrap image from 84Gi to 82Gi. Then during the build, the consumption never fell below 80Gi, while the additional drop does not seem to becaused by the build itself.

Enterying the pod shows this (dropping directories with size 0 for readability):

```
# du -sh /*
54G	./docker-graph
13M	./etc
1.4G	./google-cloud-sdk
649M	./home
1.3M	./logs
96M	./root
388K	./run
19G	./tmp
2.5M	./tools
1.3G	./usr
70G	./var
4.0K	./workspace
```
docker-graph is an emptyDir, inside `/var`, we see that the whole 70G block is coming from this mount:

```
/dev/sda3              930G   80G  850G   9% /var/lib/stdci/shared/kubevirt-images
```

Finally `/tmp` has a size of `19G`, which is coming from the new bazel output root directory which is an emptyDir now.

fgimenez (approved):
/lgtm
/approve

PR #1030, opened 2021-03-15
Define periodics for quarantine runs

fgimenez:
* Removes old presubmit quarantine lanes
* Adds environment variable for the existing periodic jobs to work with the new logic.
* Adds periodics for CNAO, SRIOV and rook-ceph lanes

/cc @dhiller @rmohr 


rmohr (commented):
This will make the periodics fail pretty often by default. I am not sure I like that.

fgimenez:
Indeed, we can create separate periodic lanes to run the stable suite, the ones in this PR will allow us to define when a test should get in and out of quarantine.

rmohr (approved):
/lgtm

I don't like it, but it looks ok.

PR #1032, opened 2021-03-15
kubevirt-presubmits: Add sig-storage k8s 1.19 job

brybacki:
The purpose of this job is to allow running a more
focused group of tests, in isolation from the other tests,
owned by the sig-storage.

This lane would always run, but won't be a gating.
It will run only the tests that are marked with sig-network

The differences between this new lane and the official k8s-1.19
```
always_run: false
optional: true
```
The TARGET has suffix `-sig-storage`

After testing we will make it always_run true,
since there is no rehearsal on k8s-1.19 we need this lane first merged.
Corresponding Kubevirt PR https://github.com/kubevirt/kubevirt/pull/5212

I have based my work on https://github.com/kubevirt/project-infra/pull/1007 (including text for commit and PR)

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
@@ -149,6 +149,42 @@ presubmits:
           resources:
             requests:
               memory: "34Gi"
+  - name: pull-kubevirt-e2e-k8s-1.19-sig-storage
+    skip_branches:
+      - release-\d+\.\d+
+    annotations:
+      fork-per-release: "true"
+      testgrid-dashboards: kubevirt-presubmits
+    always_run: false
+    optional: true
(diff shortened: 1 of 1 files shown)


brybacki:
/cc @phoracek
/cc @fgimenez
Can you take a look?

fgimenez (commented):
Looks good but we won't be able to enable it (`always_run: true`) until we have more CI resources in place. Do you need to have now this job in place for testing? In a kubevirt/kubevirt PR you could modify `automation/test.sh` to set `KUBEVIRT_E2E_FOCUS` to focus on the tests labeled with `*-sig-storage`, wdyt?

brybacki (commented):
I would like to have this job to correctly test the integration in a target environment. 

Temporarily modifying automation/test.sh might also work, but then I am skipping the integration part, and the verification of KUBEVIRT_E2E_FOCUS.

brybacki:
@fgimenez can we do it like it was done for the sig-network

PR #1034, opened 2021-03-15
Bump rehearse image

fgimenez:
Already live in prow's control plane.

/cc @dhiller 


dhiller (approved):
/approve

PR #1036, opened 2021-03-16
Add k8s-1.20-cgroupsv2 check-provision job

vasiliy-ul:
Set always_run -> false, optional -> true so it can be triggered manually when needed.

Needed for testing cgroups v2 support in kubevirt:
https://github.com/kubevirt/kubevirt/pull/4907
https://github.com/kubevirt/kubevirtci/pull/567

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/kubevirtci/kubevirtci-presubmits.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/kubevirtci/kubevirtci-presubmits.yaml
@@ -161,6 +161,33 @@ presubmits:
           requests:
             memory: "8Gi"
 
+  - name: check-provision-k8s-1.20-cgroupsv2
+    always_run: false
+    optional: true
+    decorate: true
+    decoration_config:
+      timeout: 3h
+    max_concurrency: 1
+    labels:
+      preset-dind-enabled: "true"
+      preset-docker-mirror: "true"
+    spec:
+      nodeSelector:
+        type: bare-metal-external
+      containers:
+      - image: gcr.io/k8s-testimages/kubekins-e2e@sha256:8c80c98d035a0f285ad49b964e5d0b62004844f3340407367841618e80e1503c
+        command:
+        - "/usr/local/bin/runner.sh"
+        - "/bin/sh"
+        - "-c"
+        - "cd cluster-provision/k8s/1.20 && CGROUPV2=true ../provision.sh"
+        # docker-in-docker needs privileged mode
+        securityContext:
+          privileged: true
+        resources:
+          requests:
+            memory: "8Gi"
+
   - name: check-provision-k8s-1.17
     always_run: true
     optional: false


github/ci/prow-deploy/files/jobs/kubevirt/kubevirtci/kubevirtci-presubmits.yaml
```diff
@@ -161,6 +161,33 @@ presubmits:
...
+    spec:
+      nodeSelector:
+        type: bare-metal-external
+      containers:
+      - image: gcr.io/k8s-testimages/kubekins-e2e@sha256:8c80c98d035a0f285ad49b964e5d0b62004844f3340407367841618e80e1503c
+        command:
+        - "/usr/local/bin/runner.sh"
+        - "/bin/sh"
+        - "-c"
+        - "cd cluster-provision/k8s/1.20 && CONTAINER_SUFFIX=cgroupsv2 USE_ENV=1 ../provision.sh"
```
rmohr, line 183: The job looks good. Let's reiterate a little bit more on the kubevirtci PR.

github/ci/prow-deploy/files/jobs/kubevirt/kubevirtci/kubevirtci-presubmits.yaml
```diff
@@ -161,6 +161,33 @@ presubmits:
...
+    spec:
+      nodeSelector:
+        type: bare-metal-external
+      containers:
+      - image: gcr.io/k8s-testimages/kubekins-e2e@sha256:8c80c98d035a0f285ad49b964e5d0b62004844f3340407367841618e80e1503c
+        command:
+        - "/usr/local/bin/runner.sh"
+        - "/bin/sh"
+        - "-c"
+        - "cd cluster-provision/k8s/1.20 && CONTAINER_SUFFIX=cgroupsv2 CGROUPV2=1 ../provision.sh"
```
rmohr, line 183: Would it make sense to change `CGROUPV2` to `true` and for now also automatically in our scripting set the `CONTAINER_SUFFIX` to `cgroupsv2` if it `CGROUPV2` is set?

The less configuration knobs we expose here, the easier it is to change things.

vasiliy-ul: Yes, I can set `CONTAINER_SUFFIX` inside `provision.sh`. Just to clarify: are you suggesting to mkae `CGROUPV2=true` instead of `CGROUPV2=1`?

rmohr: Yes. Except if `1` has a different meaning than `true`.

vasiliy-ul: Ok, clear. Though CONTAINER_SUFFIX needs to match the suffix of the job's name here. Probably I can add that note as a comment in provision.sh then

rmohr: do you mean that they should match in the sense of a weak relationship, like identifying jobs which create this suffix?

rmohr:
Job is correctly configured as optional to allow testing.

/lgtm
/approve

github/ci/prow-deploy/files/jobs/kubevirt/kubevirtci/kubevirtci-presubmits.yaml
vasiliy-ul: ah, no. I probably confused it with smth else. nevermind, it should be fine :)

rmohr:

Error is unrelated.

PR #1052, opened 2021-03-18
Fixed providers for e2e periodics

fgimenez:
The `-latest` periodic is running on 1.20 now, even though we are not using this provider its directory is under `cluster-up/cluster`, and the code to determine the target in these jobs is blindly picking it. 

With these changes we will have fixed targets for each e2e periodic that need to be updated when we include a new provider. The job names keep the same so that we don't need to update testgrid config

/cc @dhiller @enp0s3 


dhiller (approved):
/approve

PR #1060, opened 2021-03-19
The artifact mirror job should use the right CDI main branch

rmohr:
Let the CDI artifact mirror job create PRs against the right main
branch, which is `main` now, and not `master`.


dhiller (approved):
/approve

PR #1050, opened 2021-03-18
Remove apt command usage from SRIOV check-up job

ormergi:
The bootstrap image this job uses is fedora based and
does not have 'apt' package manager.
The job report is finished successfully, but it actually had an error that the trap didn't catch 
causing the job to deploy the default cluster k8s-1.18 instead.
https://prow.apps.ovirt.org/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirtci/559/check-up-kind-1.17-sriov/1372225414985945088

This PR removes the usage of apt and adjust the trap to catch such errors.
Also make use of `GIMME_GO_VERSION` env var to get golang binaries

PR #1065, opened 2021-03-21
sriov, presubmits, kubevirt, Update bootstrap image

oshoval:
As part of the effort to migrate project-infra images
from `docker.io` to `quay.io`,
update the bootstrap image of kubevirt sriov presubmit job
to use a newer one, which is hosted on `quay.io`.

Use a quay.io/kubevirtci/golang
in order to align kubevirt job to kubevirtci job,
and to have golang easier.
    
Since the new image doesn't have apt, but already has
`gettext-base`, update the job run command accordingly.
Remove `KUBEVIRT_PROVIDER` because `TARGET` is enough,
`KUBEVIRT_PROVIDER` will be set accordingly
by automation/test.s according to `TARGET`.

Add `set -e` in order to fail fast in case the job command
has an error, and not mask those errors.

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
@@ -363,21 +363,22 @@ presubmits:
                   - "true"
               topologyKey: kubernetes.io/hostname
       containers:
-      - image: kubevirtci/bootstrap:v20201119-a5880e0
+      - image: quay.io/kubevirtci/golang:v20210316-d295087
+        env:
+          - name: "TARGET"
+            value: "kind-k8s-sriov-1.17.0"
+          - name: "GIMME_GO_VERSION"
+            value: "1.13.8"
+          - name: "KUBEVIRT_PROVIDER" # for cluster-down in the command below
+            value: "kind-k8s-sriov-1.17.0"
         command:
         - "/usr/local/bin/runner.sh"
         - "/bin/bash"
         - "-c"
-        - >
-            apt update &&
-            apt install gettext-base -y &&
-            wget https://dl.google.com/go/go1.13.8.linux-amd64.tar.gz -nv -S &&
-            tar -xzf go1.13.8.linux-amd64.tar.gz -C /usr/local/ &&
-            export PATH=/usr/local/go/bin:$PATH &&
-            export KUBEVIRT_PROVIDER=kind-k8s-sriov-1.17.0 &&
-            export TARGET=kind-k8s-sriov-1.17.0;
-            trap "echo teardown && make cluster-down" EXIT ERR SIGINT SIGTERM;
-            automation/test.sh;
+        - |
+            set -e
+            trap "echo teardown && make cluster-down" EXIT ERR SIGINT SIGTERM
+            automation/test.sh
         # docker-in-docker needs privileged mode
         securityContext:
           privileged: true


oshoval:
https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_project-infra/1065/rehearsal-pull-kubevirt-e2e-kind-1.17-sriov/1373545235023400960

Passed

EdDev (commented):
Please update the commit message with the reasoning and something that expresses what was updated.

oshoval:
Addressed comments

EdDev (approved):

ormergi (changes requested):
Please follow this PR example https://github.com/kubevirt/project-infra/pull/1050
and change the `command` and `env` section.
These jobs should be almost identical

oshoval:
> Please follow this PR example #1050
> and change the `command` and `env` section.
> These jobs should be almost identical

This PR is for changing the bootstrap, nothing more.

Its not a go image, so i am not sure its the same state where you could use `GO_GIMMIE`
moreover, i did it already for the env vars that have a fixed value.
The env var that was left is the PATH, since its doesn't have a fixed value, it might look weird with the setting on PATH that depends on PATH.

I can try those two changes, but at least the `GO_GIMMIE` might not work.

oshoval:
It fails
https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_project-infra/1065/rehearsal-pull-kubevirt-e2e-kind-1.17-sriov/1373586003956076544
`/bin/bash: go: command not found`
It's not a go image as the one in the example you refere, hence i can't do that.
In case there is something else, please be specific.

Reverting

oshoval:
The failure is for the run with GO_GIMMIE
`11:03:21: /home/prow/go/src/github.com/kubevirt/kubevirt/cluster-up//cluster/kind-k8s-sriov-1.17.0/config_sriov.sh: line 192: go: command not found`
https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_project-infra/1065/rehearsal-pull-kubevirt-e2e-kind-1.17-sriov/1373586003956076544

oshoval:
/rehearse

(reverted `GO_GIMMIE`, it passed before when it was without it, lets run again)

ormergi:
You need to use quay.io/kubevirtci/golang:v20210316-d295087 for the go-gimme to work
This is image that is used on the check-cluster-up lane

oshoval:
> You need to use quay.io/kubevirtci/golang:v20210316-d295087 for the go-gimme to work
> This is image that is used on the check-cluster-up lane

I know, and this is what i am saying, but i am not sure this is the image we want here, 
since almost all kubevirt uses a bootstrap image, i am not sure we want sriov to be different.

oshoval:
I think we shouldn't block about the go,
its not in the direct scope of this PR, we can open other PR if needed for that.
This PR intends to align the boostrap image to be on quay.io, and i find it better to have it as most of the other images.
Its true that it saves time, but we might consider on the other hand to install go in the bootstrap so all the jobs that need go will have the option ? (on other PR as well)

I would be glad if we can divide it to parts and focus on the current topic please.

oshoval:
@qinqon @fgimenez @ormergi 
Please suggest, is it better that we will use `quay.io/kubevirtci/bootstrap:v20210311-09ebaa2` on kubevirt, as most of the jobs do,
and install go ourself.
or use the `quay.io/kubevirtci/golang:v20210316-d295087` bootstrap image as kubevirtci sriov job does and then we can use `GO_GIMMIE`.

2 Points here:
1. Its indeed better that sriov kubevirtci and kubevirt will be the same, in one way or another.
2. Worth considering (if possible), that the `quay.io/kubevirtci/bootstrap:v20210311-09ebaa2` will have go installed, so every job that need go will be able to use `GO_GIMMIE`.

Thanks

oshoval:
Rebase
since point [1] here is good enough reason, i will update the image
https://github.com/kubevirt/project-infra/pull/1065#issuecomment-803834963

oshoval:
Updated to use `quay.io/kubevirtci/golang:v20210316-d295087`

github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
```diff
@@ -327,19 +327,17 @@ presubmits:
...
         - "-c"
         - >
-            apt update &&
-            apt install gettext-base -y &&
-            wget https://dl.google.com/go/go1.13.8.linux-amd64.tar.gz -nv -S &&
-            tar -xzf go1.13.8.linux-amd64.tar.gz -C /usr/local/ &&
-            export PATH=/usr/local/go/bin:$PATH &&
-            export KUBEVIRT_PROVIDER=kind-k8s-sriov-1.17.0 &&
-            export TARGET=kind-k8s-sriov-1.17.0;
             trap "echo teardown && make cluster-down" EXIT ERR SIGINT SIGTERM;
```
ormergi, line 341: nit: Add && between the trap and automation/test.sh
`trap "echo teardown && make cluster-down" EXIT ERR SIGINT SIGTERM && automation/test.sh`

oshoval: trap can't fail, so i don't see a reason to.
its also not scalable (to add commands) this way, its better to be a command list, than a '&&' list of one command.

ormergi: The problem is that If something goes wrong the job will continue silently.
For example, this is exactly what happen with stiov check cluster-up lane last week
I think you already familiar with #1050, we could prevent the false-positive report if there was && between the trap to other commands.
Not scalable? it dosent prevent one from adding more commands.

oshoval: Trap can't fail.
Its a bad practice to '&&' it to other commands, i am talking about it specifically.
one can add commands below it and '&&' them with themself if needed.

Yes i am familiar with it, and found it also, before knowing about that PR.

We are creating here a bad practice just because with other commands there is a possible to fail.
In case one adds a command, he will need to think if to '&&' it, instead just doing it, just because its there already.

oshoval: What about adding set -e ?
automation/test.sh already have set -e is it won't change the behavior.

```
            set -e;
            trap "echo teardown && make cluster-down" EXIT ERR SIGINT SIGTERM;
            automation/test.sh;
```

/cc @qinqon since you suggested this '&&' first
please advice, and lets wrap it

qinqon: I am ok with `set -e`

oshoval:
Added `set -e`

oshoval:
Rebase

github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
ormergi: Please dont add set -e, we already have it on the test.sh script and we put a trap on the command section
that another thing to worry about.

> Trap can't fail.
Its a bad practice to '&&' it to other commands, i am talking about trap specifically.
one can add commands below it and '&&' them with themself if needed.

In case of another bootstrap image change, the shell could also change and bring unexpected behavior.
e.g: trap is not supported, trap syntax change..

oshoval: Sorry, i don't see how trap is going to stop be supported,
and the set -e here is for the command of the job
i find it much better than using && especially when the trap can't fail.
One don't need to worry on this about, on the other hand, if anybody would need to worry about adding && its the not thing to have, this set -e solve it.

oshoval:
https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_project-infra/1065/rehearsal-pull-kubevirt-e2e-kind-1.17-sriov/1373935305215709184
Passed

github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
ormergi: We have `trap .. ERR` to catch errors, in case more commands should be added to this job it should be added after the trap line. 
Adding set -e is redundant

oshoval: `set -e` cause that a command with an error will exit the "script"
trap ... ERR on its own won't help
hence its not redundant
run this and see done is printed, add set -e and see it doesn't

```
#!/bin/bash

trap "echo teardown" ERR;

false

echo done
```

qinqon: As far as I know "-e" modifier and trap are different things, one make the script fail if one of the commands fail and the other i s use to "arm" the error signal so they complement each other.

I think we can add `set -e` and `trap` so we don't care if the commands are done with `&&` or  `;`

oshoval:
The error of rehearsal-pull-kubevirt-e2e-k8s-1.18
is unrelated, i stopped it by the rebase,
from my side, all is set to proceed please.

oshoval:
/cc @fgimenez 
can you take a look please?

github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
ormergi: > This way we can add lines without infinite '&&'.
Its a bad practice to '&&' commands, especially the trap, which can't fail.
The code should be nice and easy to maintain.

This is not the example I suggested, on your examples replace the ; with && on the trap with and you will see the script will exit.

> As far as I know "-e" modifier and trap are different things, one make the script fail if one of the commands fail and the other i s use to "arm" the error signal so they complement each other.

I think we should test it first

Deleting KIND infra cluster w/o using `kind` binary may cause resources leak (cluster-down use it)
thus its important to test SRIOV lane jobs before merging.
After we finish with this job we can continue with release branches jobs.
Please inject an error so we could see that it does fail and cluster-down is invoked once.

ormergi:
Deleting KIND infra cluster w/o using `kind` binary may cause resources leak 
cluster-down use kind binary to delete the cluster.
Since you change how errors are handheld its important to test it before merging.
Could you please inject an error so we could see that it does fail and cluster-down is invoked **once**.

After we finish with this job we can continue with release branches jobs.

github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
oshoval: I tested it here
https://github.com/kubevirt/project-infra/pull/1068
it runs the trap.

oshoval:
> Deleting KIND infra cluster w/o using `kind` binary may cause resources leak
> cluster-down use kind binary to delete the cluster.
> Since you change how errors are handheld its important to test it before merging.
> Could you please inject an error so we could see that it does fail and cluster-down is invoked **once**.
> 
> After we finish with this job we can continue with release branches jobs.

This PR should not touch the release branches, you are welcome to do so on other PR.
I tested it, see and lets please continue at https://github.com/kubevirt/project-infra/pull/1065#discussion_r598657887
if needed

ormergi:
> This PR should not touch the release branches, you are welcome to do so on other PR.
> Release branches should not be touched unless mandatory.
> 

I didnt asked you to do so.

> I tested it, see and lets please continue at [#1065 (comment)](https://github.com/kubevirt/project-infra/pull/1065#discussion_r598657887)
> if needed
> 
> It could run twice, or run without kind binary even before, if needed it can be fixed on a follow PR, it doesnt related to this PR (nor part of project-infra).

What is the point in merging somthing that we would need to fix later?
We dont want cluster-down to run twice it should run once and it must use kind binary.

oshoval:
> > This PR should not touch the release branches, you are welcome to do so on other PR.
> > Release branches should not be touched unless mandatory.
> 
> I didnt asked you to do so.
> 
> > I tested it, see and lets please continue at [#1065 (comment)](https://github.com/kubevirt/project-infra/pull/1065#discussion_r598657887)
> > if needed
> > It could run twice, or run without kind binary even before, if needed it can be fixed on a follow PR, it doesnt related to this PR (nor part of project-infra).
> 
> What is the point in merging something that we would need to fix later?
> We dont want cluster-down to run twice it should run once and it must use kind binary.

This PR doesnt present a regression, it could run twice the cluster-down as well as before,
this PR is intended to update the bootstrap image thats all.

oshoval:
Rebased

There is no valid reason so far to block this PR.
There is no regression created by this PR.
We had always potential double cluster-down, (one in the job, and one in automation/test.sh)
i can remove the one in the job if needed.

ormergi:
I ran few tests and there are two options using && between commands or adding `set -e`
I guess it matter of taste.
  
> I even not sure there is a problem with running it twice, or that it can run it without the binary,
> about running it twice - it wont find it in the 2nd time.
> about running it without the binary - it means it even didnt create it.

I verified it while back and it is recommended to use kind binary for teardown (to prevent resources leak due to leftovers, also we use kind seamlessly by cluster down)
It will run cluster-down twice if there is an error because we also specify EXIT flag (but this is out of this PR scope to change it)
I guess running cluster-down wont harm much.

/lgtm

oshoval:
Cool

@fgimenez 
can you take a look please?

github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
```diff
@@ -363,19 +363,18 @@ presubmits:
...
+        env:
+          - name: "TARGET"
+            value: "kind-k8s-sriov-1.17.0"
+          - name: "GIMME_GO_VERSION"
+            value: "1.13.8"
         command:
         - "/usr/local/bin/runner.sh"
         - "/bin/bash"
         - "-c"
         - >
```
ormergi, line 376: Since you mentioned that we need add `;` at the end of each command
if we would use `|-` instead, we wont need to add `;`
(Its just a suggestion I understand if you dont want to do it on this PR)

oshoval:
Rebased

github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
oshoval: Done, used '|' because we we don't care for 'strip'
its not that we have additional empty lines.

oshoval:
Updated to use '|' instead of '>' 
so we don't need ';' between new lines

github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
```diff
@@ -363,21 +363,20 @@ presubmits:
...
-            wget https://dl.google.com/go/go1.13.8.linux-amd64.tar.gz -nv -S &&
-            tar -xzf go1.13.8.linux-amd64.tar.gz -C /usr/local/ &&
-            export PATH=/usr/local/go/bin:$PATH &&
-            export KUBEVIRT_PROVIDER=kind-k8s-sriov-1.17.0 &&
-            export TARGET=kind-k8s-sriov-1.17.0;
-            trap "echo teardown && make cluster-down" EXIT ERR SIGINT SIGTERM;
-            automation/test.sh;
+        - |
+            set -e
+            trap "echo teardown && make cluster-down" EXIT ERR SIGINT SIGTERM
```
oshoval, line 378: since i removed `export KUBEVIRT_PROVIDER=kind-k8s-sriov-1.17.0`
the trap here tries to run the wrong cluster-down
we have already a cluster-down in automation/test.sh
so i will remove this one ok ?
alternatively i can readd the `export KUBEVIRT_PROVIDER=kind-k8s-sriov-1.17.0`

oshoval:
first push - rebase
second - add KUBEVIRT_PROVIDER so the cluster-down in the command will run the right target

fgimenez (approved):
/lgtm
/approve

PR #1072, opened 2021-03-22
Increase requested memory in e2e periodic

fgimenez:
The goal is to check results for a while with this job running on a node with less load.

/cc @rmohr @dhiller


github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml
```diff
@@ -420,7 +420,7 @@ periodics:
           privileged: true
         resources:
           requests:
-            memory: "29Gi"
+            memory: "200Gi"
```
rmohr, line 423: How about 60 or 90Gi?

rmohr:
/lgtm
/approve

:+1:

PR #1073, opened 2021-03-22
Make operator job required

rmohr:
I missed on my review that the job was not `required`. The operator job needs to be required. Otherwise we would not run them at all.


fgimenez (approved):
/lgtm
/approve

PR #1078, opened 2021-03-23
sig-storage, Create a periodic job

brybacki:
In order to test sig-storage job, create a periodic
lane that will run each 8h (similar to sig-network).
We will run it for a week, and if its stable,
we will consider making the sig-network job run on each PR.


fgimenez (approved):
/lgtm
/approve

PR #1079, opened 2021-03-23
Stop using docker proxy when pushing CDI images

awels:


dhiller (approved):
/approve

Side note: you can also just remove the line completely but this one will do also I think.

PR #1080, opened 2021-03-24
Increase the worker node in the cluster to accommodate all common-templates automation tests.

shwetaap:
Increase the worker node in the cluster to accommodate all common-templates automation tests.

The old kubevirt cluster had a master node and a worker node. This cluster also creates 10 pvs in each of the nodes. The common-templates automation test for linux has 12 test cases for the different VM flavors and sizes. Each of these tests need a pv. There was a recent addition in common-templates code that specifies that the high-performance VMs only need to be placed on worker nodes. If the worker node PVs get used up before the last few high-performance VM tests are run(the PVs are bound, even after the test finishes), then there is a VM scheduling error for the last few tests. Increasing the worker node, provides another 10 PVs, that allow the tests to pass.


akrejcir:
Can you add a short explanation why is this needed?

shwetaap:
@akrejcir : The old kubevirt cluster had a master node and a worker node. This cluster also creates 10 pvs in each of the nodes. The common-templates automation test for linux has 12 test cases for the different VM flavors and sizes. Each of these tests need a pv. There was a recent addition in common-templates code that  specifies that the high-performance VMs only need to be placed on worker nodes. If the worker node PVs get used up before the last few high-performance VM tests are run(the PVs are bound, even after the test finishes), then there is a VM scheduling error for the last few tests. Increasing the worker node, provides another 10 PVs, that allow the tests to pass.

akrejcir:
Thanks
/lgtm

fgimenez (approved):
/approve

PR #1082, opened 2021-03-24
Rehearsal discover head branch

qinqon:
If a repository has a different HEAD branch name than the one specified
at project-infra (which is master as this time) the rehearsal plugin
will rehearse the job but cloning "master" instead of the HEAD branch.
This change uses "git ls-remote" to discover the HEAD branch name.****


fgimenez (commented):
Thanks for working on this! :heart: 

The tests are great, but could we use the git client instead of calling the cmd binary? We would be able to properly interpret the results without the shell pipes, awk and so on, wdyt?

qinqon:
> Thanks for working on this! 
> 
> The tests are great, but could we use the git client instead of calling the cmd binary? We would be able to properly interpret the results without the shell pipes, awk and so on, wdyt?

Dind't find the right go api to do "ls-remote" propertly I have check at https://github.com/go-git/go-git without much luck.

fgimenez:
> Dind't find the right go api to do "ls-remote" propertly I have check at https://github.com/go-git/go-git without much luck.

Have you tried https://github.com/go-git/go-git/blob/master/_examples/ls-remote/main.go ? seems to work fine

qinqon:
> > Dind't find the right go api to do "ls-remote" propertly I have check at https://github.com/go-git/go-git without much luck.
> 
> Have you tried https://github.com/go-git/go-git/blob/master/_examples/ls-remote/main.go ? seems to work fine

You right, this one is working fine

```golang
package main                                                                    
                                                                                
import (                                                                        
    "log"                                                                       
    "strings"                                                                   
                                                                                
    "github.com/go-git/go-git/v5"                                               
    "github.com/go-git/go-git/v5/config"                                        
    "github.com/go-git/go-git/v5/plumbing"                                      
    "github.com/go-git/go-git/v5/storage/memory"                                
)                                                                               
                                                                                
func main() {                                                                   
                                                                                
    // Create the remote with repository URL                                    
    rem := git.NewRemote(memory.NewStorage(), &config.RemoteConfig{             
        Name: "origin",                                                         
        URLs: []string{"https://github.com/Zenika/MARCEL"},                     
    })                                                                          
                                                                                
    // We can then use every Remote functions to retrieve wanted information    
    refs, err := rem.List(&git.ListOptions{})                                   
    if err != nil {                                                             
        log.Fatal(err)                                                          
    }                                                                           
                                                                                
    var headBranch string                                                       
    for _, ref := range refs {                                                  
        if ref.Type() == plumbing.SymbolicReference && ref.Name().String() == "HEAD" {
            headBranch = ref.Target().String()                                  
        }                                                                       
    }                                                                           
                                                                                
    log.Printf("HEAD branch: %s", strings.Split(headBranch, "/")[2])            
}                                                                               
```

qinqon:
/hold 

Us go-git

fgimenez:
/override pull-project-infra-ci-search-deploy-test

@qinqon i thin the error is legit, looks like ci-search image is no longer available:
```
$ docker pull registry.svc.ci.openshift.org/ci-search:latest
Error response from daemon: received unexpected HTTP status: 503 Service Unavailable
```

fgimenez (approved):
/lgtm
/approve

awesome, thanks!!

PR #1083, opened 2021-03-24
sriov, presubmit, Add run_if_changed rule

oshoval:
In order to save resources where we can,
add the `run_if_changed` rule for sriov presubmit job,
it will run only if the relevant sriov provider folders are changed.

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/kubevirtci/kubevirtci-presubmits.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/kubevirtci/kubevirtci-presubmits.yaml
@@ -3,7 +3,8 @@ presubmits:
   - name: check-up-kind-1.17-sriov
     annotations:
       k8s.v1.cni.cncf.io/networks: multus-cni-ns/sriov-passthrough-cni
-    always_run: true
+    always_run: false # only because we have run_if_changed
+    run_if_changed: "(^cluster-up/cluster/kind-k8s-sriov-1.17.0/.*)|(^cluster-up/cluster/kind/.*)"
     optional: false
     decorate: true
     decoration_config:


oshoval:
/cc @EdDev @qinqon @fgimenez 

Later we can do it also for the VM based providers (but they have more folders), if desired.

oshoval:
`job check-up-kind-1.17-sriov is set to always run but also declares run_if_changed targets, which are mutually exclusive`
safe to remove always run i believe, since run_if_changed will be the conditional always run

github/ci/prow-deploy/files/jobs/kubevirt/kubevirtci/kubevirtci-presubmits.yaml
```diff
@@ -3,7 +3,8 @@ presubmits:
   - name: check-up-kind-1.17-sriov
     annotations:
       k8s.v1.cni.cncf.io/networks: multus-cni-ns/sriov-passthrough-cni
-    always_run: true
+    always_run: false # only because we have run_if_changed
+    run_if_changed: "cluster-up/cluster/kind-k8s-sriov-1.17.0/.*"
```
oshoval, line 7: just need to make sure the rule is recursive

github/ci/prow-deploy/files/jobs/kubevirt/kubevirtci/kubevirtci-presubmits.yaml
```diff
@@ -3,7 +3,8 @@ presubmits:
   - name: check-up-kind-1.17-sriov
     annotations:
       k8s.v1.cni.cncf.io/networks: multus-cni-ns/sriov-passthrough-cni
-    always_run: true
+    always_run: false # only because we have run_if_changed
+    run_if_changed: "^(cluster-up/cluster/kind-k8s-sriov-1.17.0/.*)|(cluster-up/cluster/kind-k8s-sriov-1.17.0/certcreator/.*)|(cluster-up/cluster/kind-k8s-sriov-1.17.0/certcreator/certlib/.*)|(cluster-up/cluster/kind-k8s-sriov-1.17.0/manifests/.*)$"
```
oshoval, line 7: according community
`"^cluster-up/cluster/kind-k8s-sriov-1.17.0/.*"` should be enough to also catch files and folder recursively
Updating to that in order to make it simpler and less to maintain.

github/ci/prow-deploy/files/jobs/kubevirt/kubevirtci/kubevirtci-presubmits.yaml
```diff
@@ -3,7 +3,8 @@ presubmits:
   - name: check-up-kind-1.17-sriov
     annotations:
       k8s.v1.cni.cncf.io/networks: multus-cni-ns/sriov-passthrough-cni
-    always_run: true
+    always_run: false # only because we have run_if_changed
+    run_if_changed: "^cluster-up/cluster/kind-k8s-sriov-1.17.0/.*"
```
phoracek, line 7: We depend on the base kind too: https://github.com/kubevirt/kubevirtci/blob/master/cluster-up/cluster/kind-k8s-sriov-1.17.0/provider.sh#L45

oshoval: Right, thanks
will add it

or should we untangle it (duplicate the file to the two kind providers we have as was done for VM based providers?)

oshoval:
Addressed comments,
Will delete all the other folders locally and see it runs, and also will test it by touching some files after its merged.

oshoval:
Since some files in cluster-up/.* itself (not recursive) and some files on cluster-up/hack/ are needed,
i am closing this PR, unless you think it worth to create a list and maintain it.
let me know please and i will open if so

Thanks

PR #1084, opened 2021-03-24
sriov, presubmit, Remove redundant trap

oshoval:
Kubevirt `automation/test.sh` already has a trap
to cluster-down the kind sriov cluster,
it is enough to run it once, and the signal
will propagate first to the script.

Hence remove it from the job command.

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
@@ -369,15 +369,11 @@ presubmits:
             value: "kind-k8s-sriov-1.17.0"
           - name: "GIMME_GO_VERSION"
             value: "1.13.8"
-          - name: "KUBEVIRT_PROVIDER" # for cluster-down in the command below
-            value: "kind-k8s-sriov-1.17.0"
         command:
         - "/usr/local/bin/runner.sh"
         - "/bin/bash"
-        - "-c"
+        - "-ce"
         - |
-            set -e
-            trap "echo teardown && make cluster-down" EXIT ERR SIGINT SIGTERM
             automation/test.sh
         # docker-in-docker needs privileged mode
         securityContext:


ormergi:
Actually there are three https://github.com/kubevirt/kubevirt/blob/master/automation/test.sh#L180-L198
automation/test.sh is complex, I added a trap from "outside" because I dont trust it.
Did you test that it actually catch errors it in case of job is terminated in the middle or an error during cluster up?

oshoval:
> Actually there are three https://github.com/kubevirt/kubevirt/blob/master/automation/test.sh#L180-L198
> automation/test.sh is complex, I added a trap from "outside" because I dont trust it.
> Did you test that it actually catch errors it in case of job is terminated in the middle or an error during cluster up?

The 3rd is not a cluster-down, but a dumper.
I do trust it, and you can see it works
look on every instance that fails 
search here trap, see its armed, and then see there is cluster-down, without the word teardown, which means its the one in the script, and not the one in the job.
https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/5300/pull-kubevirt-e2e-kind-1.17-sriov/1374489330537467904

Please provide an example why you don't trust it, in case it hangs and SIGKILL at the end, after the 5m grace, also the one of the job will be SIGKILL so it won't be stronger.

oshoval:
first rebase
2nd change to /bin/bash -ce and inject error

oshoval:
The failure was expected, injected an error.
Updated to use `/bin/bash -ce` instead of adding set -e directive.

github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
```diff
@@ -369,15 +369,11 @@ presubmits:
...
-          - name: "KUBEVIRT_PROVIDER" # for cluster-down in the command below
-            value: "kind-k8s-sriov-1.17.0"
         command:
         - "/usr/local/bin/runner.sh"
         - "/bin/bash"
-        - "-c"
+        - "-ce"
         - |
-            set -e
-            trap "echo teardown && make cluster-down" EXIT ERR SIGINT SIGTERM
```
alonSadan, line 380: Why did you delete also the `echo teardown` ?

oshoval: you don't need it if you are not doing the cluster-down,
else it will just print without doing anything, why do we need it?

alonSadan (approved):
Thanks Or. The example provided [here](https://github.com/kubevirt/project-infra/pull/1084#issuecomment-806417859) shows that it works as expected.

oshoval:
good flow example of rehearse
https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_project-infra/1084/rehearsal-pull-kubevirt-e2e-kind-1.17-sriov/1376884866922057728

```
+ make cluster-down
"Timestamp is enabled"
./cluster-up/down.sh
15:27:08: Deleting cluster "sriov" ...
```
passed and cleaned the cluster 

bad flow example in the comments above.

@phoracek FYI
thanks

oshoval:
/cc @fgimenez 
/cc @qinqon 

can you take a look please ?

fgimenez (approved):
/approve

PR #1087, opened 2021-03-25
hco, periodic, Update nightly build to use quay.io

oshoval:
In order to fix docker.io rate limit,
switch to use quay.io

Depends on 
https://github.com/kubevirt/hyperconverged-cluster-operator/pull/1206

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/hyperconverged-cluster-operator/hyperconverged-cluster-operator-periodics.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/hyperconverged-cluster-operator/hyperconverged-cluster-operator-periodics.yaml
@@ -12,7 +12,7 @@ periodics:
     preset-dind-enabled: "true"
     preset-docker-mirror-proxy: "true"
     preset-shared-images: "true"
-    preset-kubevirtci-docker-credential: "true"
+    preset-kubevirtci-quay-credential: "true"
     rehearsal.allowed: "true"
   extra_refs:
     - org: kubevirt


oshoval:
https://github.com/kubevirt/hyperconverged-cluster-operator/pull/1206 was merged,
so its safe to merge this one please
Thanks

orenc1:
Explanation:
The connection between the label of `preset-kubevirtci-quay-credential: "true"` to the usage of `${QUAY_USER}` and `${QUAY_PASSWORD}` environment variables in the prow jobs is being done here:
https://github.com/kubevirt/project-infra/blob/master/github/ci/prow-deploy/kustom/base/configs/current/config/config.yaml#L453

```
- labels:
    preset-kubevirtci-quay-credential: "true"
  env:
  - name: QUAY_USER
    value: /etc/kubevirtci-cred/username
  - name: QUAY_PASSWORD
    value: /etc/kubevirtci-cred/password
  volumes:
  - name: kubevirtci-cred
    secret:
      defaultMode: 0400
      secretName: kubevirtci-quay-credential
  volumeMounts:
  - name: kubevirtci-cred
    mountPath: /etc/kubevirtci-cred
    readOnly: true
```

orenc1 (approved):
/approve

oshoval:
Thanks Oren

@fgimenez 
once its merged, can you run please the periodic ?
in case it will pass
we need to make those two new quay.io public, once they are created please

quay.io/kubevirtnightlybuilds/hyperconverged-cluster-operator
quay.io/kubevirtnightlybuilds/hyperconverged-cluster-webhook

Thanks

oshoval:
@orenc1 @nunnatsa 
@fgimenez suggests if we can use kubevirt or kubevirtci repos on quay.io
instead opening a new `quay.io/kubevirtnightlybuilds`,
is it fine by you?

@fgimenez can you elaborate the reasons please?

fgimenez:
> @fgimenez can you elaborate the reasons please?

Sure, as I see it we should keep all the required images within the existing orgs, so that the management is easier. If they are related to testing/CI, they should belong to `kubevirtci`, otherwise to `kubevirt`.

oshoval:
It means that we would either use (existing, and for stable versions it seems)

https://quay.io/repository/kubevirt/hyperconverged-cluster-operator?tab=tags
https://quay.io/repository/kubevirt/hyperconverged-cluster-webhook?tab=tags

Or the kubevirtci ones (doesnt exists, so if we select this option it will be HCO nightly dedicated)

Please suggest

fgimenez:
If you want to mark one concrete image version as a nightly build it would be much better to indicate it in a tag instead of assigning it to a separate org to group them, let me know wdyt.

orenc1:
I'm in favor of using `kubevirtci` quay.io org to host the hco nightly builds, instead of a dedicated org or `kubevirtnightlybuilds`.
@yuvalturg do you know who created that org and why not using `quay.io/kubevirtci`?

oshoval:
It was on docker.io until now @orenc1

oshoval:
https://github.com/kubevirt/hyperconverged-cluster-operator/pull/1208
PR for that

PR #1094, opened 2021-03-26
Add sig-compute periodic

fgimenez:
I've also taken the opportunity to specify the execution time of the sig periodics so that we don't DoS ourselves triggering all of them at once.

/cc @dhiller @vatsalparekh 

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml
@@ -190,7 +190,7 @@ periodics:
   annotations:
     testgrid-dashboards: kubevirt-periodics
     testgrid-days-of-results: "60"
-  cron: "0 */8 * * *"
+  cron: "0 1,9,17 * * *"
   decorate: true
   decoration_config:
     timeout: 4h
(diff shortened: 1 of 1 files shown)


EdDev (changes requested):
The change of the periodic cron time seems to be unrelated to the sig-compute.
Could you please split this into two PR/s?

There is also a question on how can this me maintained and based on what algorithm one needs to set these values.
It will be nice if you explain what trigger such a need to protect against DoS here.

dhiller (approved):
/approve

dhiller:
First of all: sorry @EdDev , didn't want to go over you with approving the PR, I just missed your comment :-(

> The change of the periodic cron time seems to be unrelated to the sig-compute.
> Could you please split this into two PR/s?

@EdDev do you really think this additional noise and load on CI of creating another PR is worthwile for this correction (where everyone is already complaining that it's so slow)?

> There is also a question on how can this be maintained and based on what algorithm one needs to set these values.

I figure that Federico does know how the load is spread best on the cluster, although this is a fair ask (I'd also be curious to know TBH)

> It will be nice if you explain what trigger such a need to protect against DoS here.

IIUC the only need here is to make sure to spread the load over time so that not all periodics start at the same time and cause massive cluster load.

PR #1081, opened 2021-03-24
sig-network, Make sig-network job mandatory and always run

oshoval:
This PR makes the `sig-network` job mandatory and always run.
It also removes sig-network tests from
`pull-kubevirt-e2e-k8s-1.19` job,
and from k8s-1.19 periodic job (note its name is `periodic-kubevirt-e2e-k8s-latest`).

It removes CNAO jobs, since `sig-network` runs all the tests that are needed now anyway,
see https://github.com/kubevirt/kubevirt/pull/5354 which moved the `VMIlifecycle` networking tests
to `sig-network`.

No test run duplication needed, and it will ease running
`sig-network` group isolated.

The job was tested periodically for a week, each 8 hours [1],
23 runs, all passed, beside two which are not due to the `sig-network`
tests problems [2] [3].
It's stable enough for this change.

Reduce timeout from 7h to 4h for `pull-kubevirt-e2e-k8s-1.19-sig-network`.

[1] https://prow.apps.ovirt.org/job-history/gs/kubevirt-prow/logs/periodic-kubevirt-e2e-k8s-1.19-sig-network
[2] https://prow.apps.ovirt.org/view/gs/kubevirt-prow/logs/periodic-kubevirt-e2e-k8s-1.19-sig-network/1373786699879944192
Failed due to https://github.com/kubevirt/kubevirt/issues/5285
[3] https://prow.apps.ovirt.org/view/gs/kubevirt-prow/logs/periodic-kubevirt-e2e-k8s-1.19-sig-network/1374149087418388482#1:build-log.txt%3A49
Failed due to general quay connectivity error.

Signed-off-by: Or Shoval <oshoval@redhat.com>

PR #1101, opened 2021-03-29
cnao jobs, Remove CNAO jobs in favor of sig-network

oshoval:
Since we have now sig-network which already does
all what CNAO job does, we don't need both.
Sig-network will be the one we are maintaining
and provides an isolated environment to run
networking tests.
    
Remove the presubmit and the periodic jobs of CNAO.

Hold until 
https://github.com/kubevirt/project-infra/pull/1081
https://github.com/kubevirt/kubevirt/pull/5354
are merged.

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml
@@ -628,46 +628,6 @@ periodics:
         resources:
           requests:
             memory: "29Gi"
-- name: periodic-kubevirt-e2e-k8s-latest-cnao
-  annotations:
-    testgrid-dashboards: kubevirt-periodics
-    testgrid-days-of-results: "60"
-  cron: "0 22,10 * * *"
-  decorate: true
-  decoration_config:
-    timeout: 7h
-    grace_period: 5m
-  labels:
-    preset-dind-enabled: "true"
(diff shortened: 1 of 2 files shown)


oshoval:
/hold 

merge only after https://github.com/kubevirt/project-infra/pull/1081 is merged

oshoval:
Let me know please if `VMIlifecycle` which is the only group that we don't run as part of `[sig-network]`
is needed to be run as part of `sig-network` job (without labeling it because then it will remove it from k8s-1.19)
but just with adding it to the focus in case needed.
Currently i didn't add it to `sig-network` job.

oshoval:
As discussed with @phoracek 
I will move those 2 tests, under sig-network (to one of the sig-network files)
and then we will be able to remove the CNAO lane and wont need to test the whole VMIlifecycle,
as those 2 will be the sanity check for it, with CNAO deployed on the sig-network job.

```
It("VMIs with Bridge Networking shouldn't fail after the kubelet restarts", func() {
It("VMIs with Bridge Networking should work with Duplicate Address Detection (DAD)", func() {
```

oshoval:
Following previous comment
created https://github.com/kubevirt/kubevirt/pull/5354

oshoval:
Rebased and removed the jobs instead of just disabling them from auto run.

oshoval:
Moved to https://github.com/kubevirt/project-infra/pull/1081 upon Federico request

PR #1102, opened 2021-03-30
Fix issue with github api not returning consistent label filtered results

davidvossel:
For some reason, the issue filters by label don't appear to work in the github api request. I don't have an explanation for this and it breaks our ability to make releases now because it looks like all issues are a blocker.

To fix this, I filter on our end by the blocker label after retrieving the issues. I think this is compensating for a github api regression because this definitely worked properly at one point.


fgimenez (approved):
/lgtm
/approve

PR #1112, opened 2021-04-06
Run some quarantined tests with in-memory etcd

maya-r:
We suspect disk pressure might be causing apiserver panics in the
test runs: https://github.com/kubevirt/kubevirt/issues/5353

And this will reduce the disk pressure a bit.

Tunable mentioned in https://github.com/kubevirt/kubevirtci/pull/568


fgimenez (approved):
/lgtm
/approve

PR #1113, opened 2021-04-06
sriov, periodics, Remove redundant trap

oshoval:
Kubevirt automation/test.sh already has a trap
to cluster-down the kind sriov cluster,
it is enough to run it once, and the signal
will propogate first to the script.

Hence remove it from the job command.

As done on https://github.com/kubevirt/project-infra/pull/1084 for the presubmit.

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml
@@ -717,15 +717,11 @@ periodics:
             value: "kind-k8s-sriov-1.17.0"
           - name: "GIMME_GO_VERSION"
             value: "1.13.8"
-          - name: "KUBEVIRT_PROVIDER" # for cluster-down in the command below
-            value: "kind-k8s-sriov-1.17.0"
         command:
         - "/usr/local/bin/runner.sh"
         - "/bin/bash"
-        - "-c"
+        - "-ce"
         - |
-            set -e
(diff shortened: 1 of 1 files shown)


github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml
```diff
@@ -717,15 +717,11 @@ periodics:
...
           - name: "GIMME_GO_VERSION"
             value: "1.13.8"
-          - name: "KUBEVIRT_PROVIDER" # for cluster-down in the command below
-            value: "kind-k8s-sriov-1.17.0"
         command:
         - "/usr/local/bin/runner.sh"
         - "/bin/bash"
-        - "-c"
+        - "-ce"
         - |
```
fgimenez, line 724: I think we no longer need this array element to be a multiline item, it could just be `- "automation/test.sh"`, wdyt?

oshoval: It would work like that, i left it because i prefer that in case someone need to add a command (even for debug, as we need sometimes for rehearse)
he won't need to do '&&`, otherwise we can find ourself with `&&` again in the long run
wdyt?

fgimenez: ok sgtm

fgimenez (approved):
/lgtm
/approve

PR #1118, opened 2021-04-08
sig-network, k8s-1.19, Move presubmit job to new CI cluster

oshoval:
New CI cluster is now ready for k8s-1.19 sig-network.
The missing ipv6 docker gateway was added.

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
@@ -683,7 +683,7 @@ presubmits:
       preset-shared-images: "true"
       preset-bazel-cache: "true"
       preset-bazel-unnested: "true"
-    cluster: phx-prow
+    cluster: prow-workloads
     spec:
       nodeSelector:
         type: bare-metal-external


oshoval:
@fgimenez @EdDev 
Did you conclude today in the meeting that we want to move k8s-1.19-sig-network (which always runs now)
to the new cluster, and still XFAIL the new tests ?

Or we are still according the previous plan for this lane, that we are not moving to the new cluster,
until it fixes the docker mirror ? (I think its better, so we want skip those tests)

Thanks

EdDev:
> @fgimenez @EdDev
> Did you conclude today in the meeting that we want to move k8s-1.19-sig-network (which always runs now)
> to the new cluster, and still XFAIL the new tests ?
> 
> Or we are still according the previous plan for this lane, that we are not moving to the new cluster,
> until it fixes the docker mirror ? (I think its better, so we want skip those tests)
> 
> Thanks

We will wait for the new cluster to be in a condition where the problem is fixed and then move the 1.19 job.
What exactly changed that you are asking this now?

oshoval:
> > @fgimenez @EdDev
> > Did you conclude today in the meeting that we want to move k8s-1.19-sig-network (which always runs now)
> > to the new cluster, and still XFAIL the new tests ?
> > Or we are still according the previous plan for this lane, that we are not moving to the new cluster,
> > until it fixes the docker mirror ? (I think its better, so we want skip those tests)
> > Thanks
> 
> We will wait for the new cluster to be in a condition where the problem is fixed and then move the 1.19 job.
> What exactly changed that you are asking this now?

Just understood it offline from Federico that there was an agreement to XFAIL (i guess it was meant only for non k8s-1.19),
so made sure.
Thanks

oshoval:
@fgimenez Looks like you moved to a new cluster `cluster: phx-prow` (instead of `cluster: prow-workloads`, which i guess is the same?)

I am closing this PR, i do see some failures here and there on 
https://prow.ci.kubevirt.io/?job=*sig-network&state=failure
Didn't investigate if its due to the docker-mirror ipv6

fgimenez:
Nope, the job is still executing in the old cluster, phx-prow. We still
need to migrate it to prow-workloads.

oshoval:
Thanks reopened

oshoval:
@fgimenez is it official that the new cluster have now docker proxy support, hence configures the docker ipv6 gateway address ?

fgimenez:
> @fgimenez is it official that the new cluster have now docker proxy support, hence configures the docker ipv6 gateway address ?

Yes, we have docker proxy including docker ipv6 gateway address in prow-workloads. We also have functional bazel cache running there, so in this PR we should only change the `cluster` property. This is a successful execution of the 1.19-sig-network periodic in `prow-workloads` https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/logs/periodic-kubevirt-e2e-k8s-1.19-sig-network/1397500483559493632

oshoval:
https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_project-infra/1118/rehearsal-pull-kubevirt-e2e-k8s-1.19-sig-network/1397916020462587904
Passed

oshoval:
Thanks Federico

@fgimenez @EdDev 
Please take a look

maiqueb (approved):
Thanks.

fgimenez (approved):
/approve

Thanks!

oshoval:
Thanks, we might want to move the periodic as well (unless already done)
right ?

fgimenez:
Sure, all the jobs with nodeSelector `type: bare-metal-external` should be moved to `prow-workloads`. We can't move all of them at once given that we are still waiting for 2 bare metal machines, but updating some of them is ok. If this periodic is not yet running on prow-workloads let's move it now.

PR #1119, opened 2021-04-08
remove unused overlay

fgimenez:
This overlay is not used in the prometheus-stack component.


fgimenez:
Small change and tests passing, merging

PR #1122, opened 2021-04-11
sig-network, presubmit, Update env var style

oshoval:
Update job to use the Env section, instead exporting
the needed TARGET as part of the command.

It is nicer and aligned to the recommended style.


EdDev (approved):
Thanks

phoracek (commented):
/approve

PR #1127, opened 2021-04-13
add stable e2e periodics and prevent slack reports from quarantined lanes

fgimenez:
/cc @rmohr 

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml
@@ -705,6 +705,90 @@ periodics:
         resources:
           requests:
             memory: "29Gi"
+- name: periodic-kubevirt-e2e-k8s-latest-stable
+  annotations:
+    testgrid-dashboards: kubevirt-periodics
+    testgrid-days-of-results: "60"
+  cron: "0 22,10 * * *"
+  cluster: prow-workloads
+  decorate: true
+  decoration_config:
+    timeout: 7h
+    grace_period: 5m
+  labels:
+    preset-dind-enabled: "true"
+    preset-docker-mirror-proxy: "false"
+    preset-shared-images: "true"
+  extra_refs:
+  - org: kubevirt
+    repo: kubevirt
+    base_ref: master
+    work_dir: true
+  skip_report: true
+  spec:
+    nodeSelector:
+      type: bare-metal-external
+    containers:
+      - image: quay.io/kubevirtci/bootstrap:v20210311-09ebaa2
+        env:
+          - name: TARGET
+            value: k8s-1.19
+          - name: KUBEVIRT_E2E_SKIP
+            value: "SRIOV|GPU|Operator|\\[sig-network\\]|\\[sig-storage\\]"
+          - name: KUBEVIRT_WITH_ETC_IN_MEMORY
+            value: "true"
+        command:
+          - "/usr/local/bin/runner.sh"
+          - "/bin/sh"
+          - "-c"
+          - "automation/test.sh"
+        # docker-in-docker needs privileged mode
+        securityContext:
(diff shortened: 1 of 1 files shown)


github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml
```diff
@@ -502,6 +502,90 @@ periodics:
...
+    testgrid-days-of-results: "60"
+  cron: "0 22,10 * * *"
+  cluster: prow-workloads
+  decorate: true
+  decoration_config:
+    timeout: 7h
+    grace_period: 5m
+  labels:
+    preset-dind-enabled: "true"
+    preset-docker-mirror-proxy: "false"
```
rmohr, line 517: There is no docker mirror yet?

fgimenez: nope, we don't have yet dynamically provisioned persistent volumes in the new cluster

rmohr (commented):
The PR goes against my expectation. My expectation would be that the main periodics run all tests except quarantained tests and that we add lanes which ONLY run the quarantined tests.

fgimenez:
As I see it we need to keep running the whole suite to check if quarantined tests are already fixed, running them in isolation would not tell us if they are good to join the stable suite again, wdyt? These lanes (stable + quarantined) tell us what would look like the merge blocking jobs PRs if the quarantine tests would be back to the stable suite.

github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml
rmohr: I think it is critical to have a dedicated cache-node (otherwise the caches interfere with workloads, we can not mix them), like i the current cluster. You don't need PVCs for that. For caches it is even preferable to use host paths.

rmohr: On e.g. managed clusters partition layouts can not be influenced by us, disks size may not be big enough, and one can normally request fast storage, or storage which has independent networking, which makes PVCs a good option there too.

fgimenez: Ok sounds good to me to have a dedicated cache node. We only have the new SR-IOV machines in this cluster, I think it would be better to use more regular bare metals as the cache node, maybe even the same srv05 we use on phx-prow, which already has in place the partitions, wdyt?

I still think we would need some kind of dynamically/statically provisioned PVs in this cluster, besides the cache workloads. prometheus will need them, and maybe other workloads too. I opened a support ticket in IBM UI and they proposed a few options, all of them from third-parties or ibm cloud operator, which I think would not work for our use case.

rmohr: > I still think we would need some kind of dynamically/statically provisioned PVs in this cluster, besides the cache workloads. prometheus will need them, and maybe other workloads too. I opened a support ticket in IBM UI and they proposed a few options, all of them from third-parties or ibm cloud operator, which I think would not work for our use case.

I would suggest to follow an architecture where prometheus in this setup is non-persistent, but it is scraped from another prometheus in a managed cluster with persistent storage. The prometheus in the managed cluster would then also fuel all dashboards.

fgimenez: Indeed, i was thinking about using thanos for persisting and centralizing the metrics, but in any case I think we would need a small temporary storage for each of the prometheus instances that are going to be aggregated. Will check in any case.

fgimenez:
/close

closing this one in favor of #1306

PR #1131, opened 2021-04-14
Revert running etcd in-memory

maya-r:
The original background was a blind attempt at reducing I/O
load, we now have a better theory for the problem, and we
want to understand why this lane has more failures than the
-prev lane by reducing differences.


fgimenez (approved):
/lgtm
/approve

PR #1136, opened 2021-04-15
Remove virtblocks job

fgimenez:
virtblocks is no longer active and the job can be removed.

/cc @rmohr @dhiller 


dhiller (approved):
/approve

PR #1143, opened 2021-04-16
Make the arm cross compile job always run

rmohr:
With https://github.com/kubevirt/kubevirt/pull/5456 merged we can start running the crosscompile check on every PR. Still keeping it optional for the next 2-3 days.


dhiller (approved):
/approve
/hold

PR #1144, opened 2021-04-16
Align skipped tests for 1.19 and 1.18 periodic e2e from master

fgimenez:
/cc @brybacki @dhiller


brybacki:
I am not sure about this. For 1.19 we have two additional separate lanes - sig-network and sig-storage, that is why we can skip the network and storage tests in main 1.19 lane. There are no such lanes for 1.18.

fgimenez:
@brybacki you are totally right, better keep the same skips we have in the presubmits, closing.

PR #1145, opened 2021-04-19
Promote k6t 1.20 sig-net presubmit lane

qinqon:
Looks like the periodic jobs for kubevirt sig-network k8s-1.20 are fine
[1], let's promote the presubmit so it's always run but not voting and
not reporting to github.

[1] https://grafana-kubevirt-prow.apps.ovirt.org/d/Cz6zgEEGz/e2e-jobs-overview?orgId=1&refresh=2h&from=now-5d&to=now&var-job_name=periodic-kubevirt-e2e-k8s-1.20-sig-network

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
@@ -48,10 +48,9 @@ presubmits:
     annotations:
       fork-per-release: "true"
       testgrid-dashboards: kubevirt-presubmits
-    always_run: false
-    optional: true
+    always_run: true
+    optional: false
     cluster: prow-workloads
-    skip_report: true
     decorate: true
     decoration_config:
       timeout: 4h


EdDev:
> let's promote the presubmit so it's always run but not voting and not reporting to github.

Why not voting and not reporting?

qinqon:
> > let's promote the presubmit so it's always run but not voting and not reporting to github.
> 
> Why not voting and not reporting?

Let's add voting and reporting and do some "rehearse" here.

qinqon:
/hold cancel 

rehearse is ok

EdDev (approved):
Thanks you!

fgimenez (commented):
As I understand it this lane should replace 1.19-sig-network lane, right? I mean, with the current resources we should not have 2 provider lanes per sig.

Also, I think we didn't have enough executions to mark this job as mandatory, wdyt?

EdDev:
> As I understand it this lane should replace 1.19-sig-network lane, right? I mean, with the current resources we should not have 2 provider lanes per sig.

1.22-sig-network  future job will replace 1.19, not this one.
 
> Also, I think we didn't have enough executions to mark this job as mandatory, wdyt?

There were 15 successful runs, I think it is enough.

qinqon:
sig-network lane is much smaller so it has less flakies so it need less re-runs I woudl say also I suppose that we want to test sig-network and friends at latest three k8s versions.

EdDev:
> sig-network lane is much smaller so it has less flakies so it need less re-runs, and I suppose that we want to test sig-network and friends at latest three k8s versions.

We want to break the one-big job into smaller ones which has been already proven to help.
We need to do it on all lanes, although I care less about 1.18 ones.

fgimenez:
> We want to break the one-big job into smaller ones which has been already proven to help.
> We need to do it on all lanes, although I care less about 1.18 ones.

No question about that, but with the current resources this is not a good idea from my PoV, we would need to trigger 4 jobs more per PR plus periodics.

fgimenez:
> There were 15 successful runs, I think it is enough.
> EDIT: https://prow.apps.ovirt.org/job-history/gs/kubevirt-prow/logs/periodic-kubevirt-e2e-k8s-1.20-sig-network

Looks good :+1:

qinqon:
> > We want to break the one-big job into smaller ones which has been already proven to help.
> > We need to do it on all lanes, although I care less about 1.18 ones.
> 
> No question about that, but with the current resources this is not a good idea from my PoV, we would need to trigger 4 jobs more per PR plus periodics.

We are suppose to stop the sig-network k8s 1.20 periodic when we merge this, also is not something that we can decide we need to test SIG lanes per latest three k8s versions, if we don't have resources we have to ask for it.

EdDev:
> > No question about that, but with the current resources this is not a good idea from my PoV, we would need to trigger 4 jobs more per PR plus periodics.
> 
> We are suppose to stop the sig-network k8s 1.20 periodic when we merge this, also is not something that we can decide we need to test SIG lanes per latest three k8s versions, if we don't have resources we have to ask for it.

We have provided 2 new HW machines to the pool, I do not understand what resources are missing.
We showed that the added workload is 30min and I do not see how it will present a problem to the existing load.

It is also impractical to continue running only the existing jobs, we are supposed to support 3 versions. If we cannot, lets drop 1.18 and bring in 1.20, as it is more relevant to our work.

phoracek:
Kubernetes is on 1.21 already, we are failing to test with the recent 3 releases. Unless our rule is to support 3 releases back, without saying what is the newest one... Sticking to the former, we should drop 1.18 and use 1.20 instead. This would resolve this specific PR.

However, back to our promise, we are late with our 1.21 testing. @fgimenez once we have 1.21 jobs ready, will we have HW to run them on?

fgimenez:
> Sticking to the former, we should drop 1.18 and use 1.20 instead. This would resolve this specific PR.

@phoracek yes, this sounds good to me :+1: 

> However, back to our promise, we are late with our 1.21 testing. @fgimenez once we have 1.21 jobs ready, will we have HW to run them on?

We need to be able to lock an additional 4 * 34Gb  of memory per code push and per periodic run (4 * 3 times per day), I'll check if we can have a new BM to handle this increased load.

fgimenez:
> We have provided 2 new HW machines to the pool, I do not understand what resources are missing.
> We showed that the added workload is 30min and I do not see how it will present a problem to the existing load.

The problem is triggering 4 new jobs per code push. The new machines help to lower the existing load and with the increased number of e2e tests we already run per code push. These changes duplicate both.

EdDev:
> > We have provided 2 new HW machines to the pool, I do not understand what resources are missing.
> > We showed that the added workload is 30min and I do not see how it will present a problem to the existing load.
> 
> @EdDev the main problem is triggering 4 new jobs per code push. The new machines help to lower the existing load and with the increased number of e2e tests we already run per code push and the new e2e periodics. These changes imply we should duplicate both.

If I'm not mistaken, this PR looks to add a single job.
In the past we had 1.17 and 1.20 running and for the stabilization effort we took them out. Are now under load?

fgimenez:
> If I'm not mistaken, this PR looks to add a single job.

For sure, but why we would have only one sig lane on two providers, the other sigs should have these jobs too if we merge this one. Wjhich is absolutely fine from my PoV once the we make sure we have enough resources.

> In the past we had 1.17 and 1.20 running and for the stabilization effort we took them out. Are now under load?

Not so much, but that doesn't mean we can be very quickly if we lock a lot of resources per code push. 

Do we need to keep 1.18 at this point? As Petr suggested I think we should drop it and then start adding the new sig jobs for 1.20, wdyt?

qinqon:
> > If I'm not mistaken, this PR looks to add a single job.
> 
> For sure, but why we would have only one sig lane on two providers, the other sigs should have these jobs too if we merge this one. This is absolutely fine from my PoV once we make sure we have enough resources.

Every SIG is the owner of their lanes they have to decide if a sig lane is prometed from periodic to presubmit, in this case is sig networking the one doing this step.

> 
> > In the past we had 1.17 and 1.20 running and for the stabilization effort we took them out. Are now under load?
> 
> Not so much, but that doesn't mean we can be very quickly if we lock a lot of resources per code push.
> 
> Do we need to keep 1.18 at this point? As Petr suggested I think we should drop it and then start adding the new sig jobs for 1.20, wdyt?

If we are already covering all the tests at 1.19 let's remove 1.18 and start activating sigs at 1.20.

fgimenez:
/lgtm
/approve

> If we are already covering all the tests at 1.19 let's remove 1.18 and start activating sigs at 1.20.

ok sgtm, I'll propose a PR for removing it

PR #1149, opened 2021-04-20
build multiarch bootstrap image

zhlhahaha:
build bootstrap image for amd64 and arm64
New Dockerfile for bootstrap has `ARG ARCH` which is not acquired in the build-bootstrap-image test, so the test would always fail.

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/project-infra/project-infra-postsubmits.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/project-infra/project-infra-postsubmits.yaml
@@ -83,12 +83,15 @@ postsubmits:
                 memory: "1Gi"
               limits:
                 memory: "1Gi"
-    - name: publish-bootstrap-image
+    - name: publish-multiarch-bootstrap-image
       always_run: false
       run_if_changed: "images/golang/.*|images/bootstrap/.*"
       annotations:
         testgrid-create-test-group: "false"
       decorate: true
+      decoration_config:
+        grace_period: 5m0s
+        timeout: 4h0m0s
       max_concurrency: 1
       labels:
         preset-dind-enabled: "true"
@@ -104,16 +107,16 @@ postsubmits:
               - |
                 cat "$QUAY_PASSWORD" | docker login --username $(cat "$QUAY_USER") --password-stdin=true quay.io
                 cd images
-                ./publish_image.sh bootstrap quay.io kubevirtci
-                ./publish_image.sh golang quay.io kubevirtci
+                ./publish_multiarch_image.sh bootstrap quay.io kubevirtci
+                ./publish_multiarch_image.sh -l golang quay.io kubevirtci
             # docker-in-docker needs privileged mode
             securityContext:
               privileged: true
             resources:
               requests:
-                memory: "1Gi"
+                memory: "8Gi"
               limits:
-                memory: "1Gi"
+                memory: "8Gi"
     - name: publish-kubekins-e2e-image
       always_run: false
       run_if_changed: "images/kubekins-e2e/.*"
--- a/github/ci/prow-deploy/files/jobs/kubevirt/project-infra/project-infra-presubmits.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/project-infra/project-infra-presubmits.yaml
@@ -85,10 +85,13 @@ presubmits:
               memory: "1Gi"
             limits:
               memory: "1Gi"
-  - name: build-bootstrap-image
+  - name: build-multiarch-bootstrap-image
     always_run: false
     run_if_changed: "images/bootstrap/.*|images/golang/.*"
     decorate: true
+    decoration_config:
+      grace_period: 5m0s
+      timeout: 4h0m0s
     labels:
       preset-dind-enabled: "true"
       preset-docker-mirror-proxy: "true"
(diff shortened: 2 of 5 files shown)


zhlhahaha:
cc: @rmohr 
cc: @fgimenez

zhlhahaha:
> @zhlhahaha do we still need multi-arch bootstrap images? If possible it would be great if we can stay with the cross-build (and 

Yes, once we have multi-arch bootstrap images, we can run unit-tests and kind-e2e tests parallelly on the ARM64 server. 

The prow utility images have already supported multi-arch. https://github.com/kubernetes/test-infra/pull/22362
For the Bootstrap image, there is a PR but no maintainer review it. https://github.com/kubernetes/test-infra/pull/23044

images/golang/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM bootstrap
+FROM bootstrap-amd64
```
dhiller, line 1: Don't we need the `ARCH` here? If we don't we would just create a regular amd64 golang image and could thus replace the `publish_multiarch_image.sh` for golang with the regular `publish_image.sh` no?

dhiller (commented):
Looks very good already, besides the IMHO missing ARCH inside the golang Dockerfile.

zhlhahaha:
Hi @dhiller the PR has updated, would you like to take a look?

images/golang/Dockerfile
zhlhahaha: We need multi-arch golang image, I have update the PR.

dhiller (approved):

dhiller (commented):
/approve

/hold @brianmcarey mind taking another look? If you don't find anything, then let's remove the hold

zhlhahaha:
Hi @dhiller 
I see the two required tests `build-bootstrap-image` and `build-golang-image` have failed. As you can see in this PR, the build command for this two images has changed to following command, but in the tests, it still used old version build command. So the tests failed.
```
./publish_multiarch_image.sh bootstrap quay.io kubevirtci
./publish_multiarch_image.sh -l golang quay.io kubevirtci
```

github/ci/prow-deploy/files/jobs/kubevirt/project-infra/project-infra-presubmits.yaml
```diff
@@ -85,7 +85,7 @@ presubmits:
               memory: "1Gi"
             limits:
               memory: "1Gi"
-  - name: build-bootstrap-image
+  - name: build-multiarch-bootstrap-image
     always_run: false
     run_if_changed: "images/bootstrap/.*|images/golang/.*"
     decorate: true
```
brianmcarey, line 91: As we are doubling the number of images that we are building - we may need to update the default 2 hours timeout for this job. I tried running this locally and it timed out - I updated the timeout to 3 hours and the job passed. Here is an example of how to configure the timeout:
https://github.com/kubevirt/project-infra/blob/9ce9f79571657feabd0847fce5ea058936698022/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml#L110-L112

Once we have the timeout increased for this and the publish job - I think this looks good.

images/publish_multiarch_image.sh
```diff
@@ -0,0 +1,121 @@
...
+
+publish_manifest() {
+    export DOCKER_CLI_EXPERIMENTAL="enabled"
+    local amend
+    local full_image_name="${1:?}"
+    amend=""
+    for arch in ${archs[*]};do
+        amend+=" --amend ${full_image_name}-${arch}"
+    done
+    docker manifest create ${full_image_name} ${amend}
```
brianmcarey, line 108: It would be cool to try this with a tool like buildah but we can try doing that later. Other kubevirt projects are starting to adopt buildah for multi-arch builds - https://github.com/kubevirt/hostpath-provisioner/pull/115

brianmcarey (commented):
/lgtm 

/unhold

brianmcarey:
/override build-bootstrap-image

Local run of job completed successfully.

xpivarc (commented):
Looks good. I think we should have just one `publish-image` in the end but this will keep us going.

github/ci/prow-deploy/files/jobs/kubevirt/project-infra/project-infra-postsubmits.yaml
```diff
@@ -104,16 +107,16 @@ postsubmits:
               - |
                 cat "$QUAY_PASSWORD" | docker login --username $(cat "$QUAY_USER") --password-stdin=true quay.io
                 cd images
-                ./publish_image.sh bootstrap quay.io kubevirtci
-                ./publish_image.sh golang quay.io kubevirtci
+                ./publish_multiarch_image.sh bootstrap quay.io kubevirtci
+                ./publish_multiarch_image.sh -l golang quay.io kubevirtci
```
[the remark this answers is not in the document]
brianmcarey: Good point - we probably don't need this specific local image flag.

images/publish_multiarch_image.sh
```diff
@@ -0,0 +1,121 @@
...
+build_image() {
+    local local_base_image=${1:?}
+    local build_target="${2:?}"
+    local image_name="${3:?}"
+    local base_image="${4:?}"
+    # add qemu-user-static
+    docker run --rm --privileged multiarch/qemu-user-static --reset -p yes
+    # build multi-arch images
+    for arch in ${archs[*]};do
+        if [[ $local_base_image == false ]]; then
```
brianmcarey, line 86: If the base image is remote - docker will pull this image as part of the build automatically - @zhlhahaha do we need this special handling for local base image?

zhlhahaha: Yes, otherwise, docker would not pull correct CPU arch baseimage.

github/ci/prow-deploy/files/jobs/kubevirt/project-infra/project-infra-postsubmits.yaml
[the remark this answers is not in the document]
zhlhahaha: We need this flag. For Dockerfile whose baseimage is from remote repository, we need manually pull correct CPU arch base image, otherwise it would use local image even if the local image has different CPU arch with `--platform`.
For these Dockerfile we use local built image, like the `golang` image, we do not need to pull base image.

images/publish_multiarch_image.sh
brianmcarey: Ok thanks @zhlhahaha

PR #1151, opened 2021-04-21
Added kubeconfig to crier options

fgimenez:
We started getting errors in testgrid after a recent update, the test passed but the job was shown as failed with a message `podinfo.json not found please install prow's gcs reporter`, probably this has been failing silently for a while. 

After some investigation turns out that crier requires to have access to the kubeconfig with all the contexts when executing jobs in remote clusters in order to upload podinfo.json to the job artifacts.Tested successfully. 

/cc @brybacki 
/cc @dhiller 


brybacki:
Nice, looks like it works, the problematic jobs are now green!
but I do not know the prow and its yaml files, I do not feel confidence to comment on it

dhiller (commented):
/approve
/hold

Thanks! Remove whenever ready!

dhiller (approved):

fgimenez:
/unhold

Thanks! This is ready to go, all is working fine and the new manifests already include it https://github.com/kubevirt/project-infra/blob/master/github/ci/prow-deploy/kustom/base/manifests/test_infra/1736ab42da959a5be7de22586b495120000bc4ef/crier_deployment.yaml

PR #1154, opened 2021-04-21
Enable 1.20-sig-sorage presubmit

fgimenez:
Once it is enabled we need to watch for 24h the number of pending jobs in https://grafana-kubevirt-prow.apps.ovirt.org/d/Cz6zgEEGz/e2e-jobs-overview?orgId=1&refresh=2h to check if they increase too much. 

/cc @brybacki 
/cc @dhiller 


dhiller (approved):
/approve

PR #1171, opened 2021-04-28
Revert memory increase for e2e periodic

fgimenez:
No improvements were seen in the test suite stability, reverting the memory request to its previous value.

/cc @rmohr @dhiller


oshoval:
Maybe worth to revert also the changes of sriov memory increase ?
Not sure it helps there as well, there was 29Gb already which is quite a lot, 
the VMs use only 512Mi, and there are at max around 2 VMs at the same time on sriov lane.
the node them self also dont consume that much i think

dhiller (approved):
/approve

PR #1173, opened 2021-04-28
Update default Prow config

fgimenez:
On deployment we had a script that updated the default (old) config to the desired values. But other tools like config_updater are relying on the actual config file, we need to update it to the latest values. With these changes there's no need to update the config on deployment, the yq script is removed.

/cc @rmohr @dhiller  


dhiller (approved):
/approve
/hold

Looks good to me, unhold whenever ready :)

fgimenez:
/unhold

thanks!

PR #1175, opened 2021-04-28
Fix rehearse deployment

fgimenez:
Required by rehearse plugin, without it the rehearse replicaset complains with:
```
Error creating: pods "rehearse-5cc657d6f6-" is forbidden: error looking up service account kubevirt-prow/prow-rehearse: serviceaccount "prow-rehearse" not found
```

/cc @oshoval @dhiller 

### DIFF
--- a/github/ci/prow-deploy/kustom/base/manifests/local/prow-rehearse-deployment.yaml
+++ b/github/ci/prow-deploy/kustom/base/manifests/local/prow-rehearse-deployment.yaml
@@ -21,7 +21,7 @@ spec:
       terminationGracePeriodSeconds: 180
       containers:
         - name: rehearse
-          image: quay.io/kubevirtci/rehearse:v20210312-b3a5e0f2
+          image: docker.io/kubevirtci/rehearse:v20200923-8930f83
           imagePullPolicy: Always
           args:
             - --dry-run=false
--- a/github/ci/prow-deploy/kustom/base/configs/current/plugins/plugins.yaml
+++ b/github/ci/prow-deploy/kustom/base/configs/current/plugins/plugins.yaml
@@ -320,6 +320,7 @@ external_plugins:
   - name: cherrypicker
   kubevirt/project-infra:
   - name: rehearse
+    endpoint: http://prow-rehearse:9900
     events:
       - pull_request
       - issue_comment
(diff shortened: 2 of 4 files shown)


github/ci/prow-deploy/kustom/base/manifests/local/prow-rehearse-deployment.yaml
```diff
@@ -21,7 +21,7 @@ spec:
       terminationGracePeriodSeconds: 180
       containers:
         - name: rehearse
-          image: quay.io/kubevirtci/rehearse:v20210312-b3a5e0f2
+          image: docker.io/kubevirtci/rehearse:v20200923-8930f83
```
dhiller, line 24: Why do we use docker again? Also this image seems very old?

dhiller (commented):
One question using the old rehearse image from docker inline.

github/ci/prow-deploy/kustom/base/manifests/local/prow-rehearse-deployment.yaml
fgimenez: Indeed, this is the one we were using in the old prow control plane. 

The latest on quay doesn't trigger the jobs for some reason, see https://github.com/kubevirt/project-infra/pull/851#issuecomment-828498877 

IIRC we tried to bump to the latest rehearse image on quay when we fixed the tests and changed the build process to bazel but it had also this issue of not triggering jobs, then we reverted to the older one.

dhiller (approved):
/approve

github/ci/prow-deploy/kustom/base/manifests/local/prow-rehearse-deployment.yaml
oshoval: Lets mirror it to quay ?
since we don't have proxy atm on the new cluster, so we might hit docker rate limit
(even in the future when we do have, a cache miss is prone for that)

oshoval:
Thanks

github/ci/prow-deploy/kustom/base/manifests/local/prow-rehearse-deployment.yaml
fgimenez: rehearse plugin runs with prow control plane, we have docker proxy there. I think we should instead try to fix the latest version.

PR #1187, opened 2021-05-02
Add optional k8s-1.20 ceph lane

maya-r:
The ceph lane seems like it uses an old version of kubernetes, let's add an optional lane to check if it's stable, and later switch.


dhiller (commented):
/approve

dhiller (approved):

PR #1190, opened 2021-05-03
Set K8s 1.20 sig-compute lane to run always

enp0s3:
This is a part of an effort to put 1.20 lane as mandatory.
we would like to observe the stability of the lane
and as part of it we will enable the reporting and make
it always run.


dhiller (approved):
/approve

PR #1197, opened 2021-05-05
Include 1.19-sig-compute in reports

fgimenez:
1.19-sig-compute is being executed https://prow.ci.kubevirt.io/?job=periodic-kubevirt-e2e-k8s-1.19-sig-compute but the results are not being reported.

/cc @vatsalparekh @enp0s3 @dhiller 


dhiller (approved):
/approve

PR #1159, opened 2021-04-24
Fixes for prow-deploy

fgimenez:
Signed-off-by: Federico Gimenez <fgimenez@redhat.com>

PR #1201, opened 2021-05-07
Re-enable prow deploy verify step

fgimenez:
With the changes in https://github.com/kubevirt/project-infra/pull/1159, we started using the same manifests for both production and tests, so that the tests actually exercise the infrastructure that is going to be deployed. This broke part of the verification step because we had specific patches for ingress in the testing overlay that were removed and also because on production we don't need to deploy an ingress controller (this is handled by IBM cloud services, previously the testing overlay took care of setting up the ingress controller).

These changes:
* Patch the production ingress to be used in tests
* Include the most recent nginx-ingress controller adapted to kind in molecule tests, installs, and waits for it to be ready
* Reenable the verification step in molecule tests 
* Not related to fixes in tests, reverts a change in `project_infra_root` variable for `prow-workloads` to be able to run on CI env.

/cc @dhiller


github/ci/prow-deploy/kustom/overlays/kubevirtci-testing/resources/ingress-nginx-controller.yaml
```diff
@@ -400,8 +400,15 @@ spec:
...
+      nodeSelector:
+        ingress-ready: 'true'
+        kubernetes.io/os: linux
+      tolerations:
+        - effect: NoSchedule
+          key: node-role.kubernetes.io/master
+          operator: Equal
       serviceAccountName: ingress-nginx
-      terminationGracePeriodSeconds: 300
+      terminationGracePeriodSeconds: 0
```
dhiller, line 411: I figure it's by intent that we are doing a hard stop here?

dhiller (commented):
/lgtm
/approve

One question in line though, being curious  :slightly_smiling_face:

dhiller (approved):

PR #1218, opened 2021-05-13
Deploy prometheus-operator on prow-workloads cluster

fgimenez:
For now prometheus storage is held in memory, once we have the cache node available we can use it instead.

/cc @rmohr @dhiller 


rmohr:
/lgtm
/approve
/hold

fgimenez:
/unhold

All good thanks!

PR #1221, opened 2021-05-14
Adapt TestGrid config managment to Config Merger

fgimenez:
Updates to use Config Merger instead of transfigure for TestGrid config management. 

Until now with transfigure every time we merged changes in our jobs config, the new testgrid config was generated and PR was issued in k8s/test-infra with the relevant changes. With Config Merger, we first instruct TestGrid to read our config from a GCS bucket and then we can manage that config without requiring merging changes in k8s/test-infra. Config Merger configuration on test-infra side is described here https://github.com/GoogleCloudPlatform/testgrid/tree/master/cmd/config_merger and this is an example of the steps required to migrate to use Config Merger https://github.com/GoogleCloudPlatform/oss-test-infra/issues/832

These changes remove the old transfigure jobs and add 3 new jobs:
* testgrid config check presubmit: when there are changes on job configs we generate the new testgrid config and check that it is ok.
* testgrid update postsumit: after merging changes in jobs configs we we generate the new testgrid config, check that it is ok and propose a PR to update it in project-infra.
* testgrid upload postsubmit: after merging changes in testgrid config, the new testgrid config file is uploaded to GCS.

/cc @dhiller

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/project-infra/project-infra-postsubmits.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/project-infra/project-infra-postsubmits.yaml
@@ -614,7 +614,7 @@ postsubmits:
           secret:
             secretName: pgp-bot-key
     - name: post-project-infra-update-testgrid-config
-      run_if_changed: '^github/ci/prow-deploy/files/orgs\.yaml$'
+      run_if_changed: '^github/ci/prow-deploy/files/jobs/.*$'
       branches:
       - master
       annotations:
--- a/github/ci/prow-deploy/files/jobs/kubevirt/project-infra/project-infra-postsubmits.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/project-infra/project-infra-postsubmits.yaml
@@ -613,8 +613,8 @@ postsubmits:
         - name: pgp-bot-key
           secret:
             secretName: pgp-bot-key
-    - name: post-project-infra-upload-testgrid-config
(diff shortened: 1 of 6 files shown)


github/ci/prow-deploy/files/jobs/kubevirt/project-infra/project-infra-postsubmits.yaml
```diff
@@ -613,8 +613,8 @@ postsubmits:
         - name: pgp-bot-key
           secret:
             secretName: pgp-bot-key
-    - name: post-project-infra-upload-testgrid-config
-      run_if_changed: '^(github/ci/prow-deploy/files/jobs/.*)|(github/ci/testgrid/gen-config\.yaml)$'
+    - name: post-project-infra-update-testgrid-config
+      run_if_changed: '^github/ci/prow-deploy/files/orgs\.yaml$'
```
dhiller, line 617: This change for `run_if_changed` doesn't look right? Shouldn't it be like this?
```suggestion
      run_if_changed: '^github/ci/prow-deploy/files/jobs/.*$'
```

dhiller (changes requested):
Looks like a great improvement!

One thing I found as commented.

github/ci/prow-deploy/files/jobs/kubevirt/project-infra/project-infra-postsubmits.yaml
fgimenez: Absolutely, great catch, thanks!

dhiller (approved):
/approve

PR #1222, opened 2021-05-14
Dockerfile for website

mazzystr:
Dockerfile for website

```
$ podman image ls | grep kubevirt.github.io
localhost/kubevirt-kubevirt.github.io  latest  6c3d5350fc07  About a minute ago  842 MB
```

Image is parred down to 840MiB.  It could probably be parred down further if we break dnf and systemd


mazzystr:
@dhiller Need rubber stamp for this

PR #1243, opened 2021-05-20
Fix testgrid update job

fgimenez:
Changes in this PR:
* Use bazel cache, reduces execution time from 33m here https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/logs/post-project-infra-update-testgrid-config/1395311825699999744 to 3m here https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/logs/post-project-infra-update-testgrid-config/1395331223252897792
* Include missing kubernetes/test-infra extra_ref
* Use `GIT_ASKPASS` instead of plain token
* Reduce verbosity of scripts
* A couple of fixes related to the GitHub username

Successful execution here https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/logs/post-project-infra-update-testgrid-config/1395331223252897792 

/cc @dhiller @rmohr

### DIFF
--- a/github/ci/testgrid/hack/common.sh
+++ b/github/ci/testgrid/hack/common.sh
@@ -1,12 +1,12 @@
 #!/bin/bash
 
-set -euxo pipefail
+set -euo pipefail
 
 BASEDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
 PROJECT_INFRA_ROOT=$(readlink -f --canonicalize ${BASEDIR}/../../../..)
 TEST_INFRA_ROOT=$(readlink -f --canonicalize ${PROJECT_INFRA_ROOT}/../../kubernetes/test-infra)
 TESTGRID_CONFIG=$(readlink -f --canonicalize ${BASEDIR}/../gen-config.yaml)
-USER=kubevirtbot
+USER=kubevirt-bot
 EMAIL=kubevirtbot@redhat.com
 
 generate_config(){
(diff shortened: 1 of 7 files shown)


github/ci/testgrid/hack/common.sh
```diff
@@ -30,6 +30,7 @@ generate_config(){
 
 run_tests(){
     (
+        ${PROJECT_INFRA_ROOT}/hack/create_bazel_cache_rcs.sh
```
rmohr, line 33: Can't you just transparently enable the cache by simple setting the bazel preset label on the job?

Something like:

```
preset-bazel-cache: "true"
```

this way we would stay in sync with what we use as bootstrap image, and we don't lookup a non-existing cache if things are called locally.

rmohr (changes requested):
Looks great. One suggestion.

github/ci/testgrid/hack/common.sh
fgimenez: Good point, added the preset label to set the proper env vars. However we still need to call the `create_bazel_cache_rcs.sh` script to set up bazelrc settings, for the testgrid config management jobs we need to use the transfigure image `gcr.io/k8s-prow/transfigure` (which ships configurator and other required tools) instead of our bootstrap image (on which we execute `create_bazel_cache_rcs.sh` during image build).

rmohr: ah :+1:

rmohr (commented):
/lgtm
/approve
/hold

dhiller (approved):

PR #1244, opened 2021-05-20
kubevirt: Do not run e2e tests on documentation changes

akrejcir:
Added `run_if_changed` fields to e2e presubmit tests, that match all files, except documentation.

Specifically, these file patterns are ignored:
- *.md
- *.txt
- docs/*

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits-0.40.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits-0.40.yaml
@@ -18,6 +18,8 @@ presubmits:
       preset-shared-images: "true"
     max_concurrency: 11
     name: pull-kubevirt-e2e-k8s-1.20-0.40
+    # The regex matches all files except these patterns: *.md, *.txt, docs/*
+    run_if_changed: '^(?:[^d]|d[^o]|do[^c]|doc[^s]|docs[^/]).*(?:[^.]txt|[^t]xt|[^x]t|[^.]md|[^m]d|[^dt])$'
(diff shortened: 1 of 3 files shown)


fgimenez:
Thanks for the PR @akrejcir 

> It seems regex lookarounds are not implemented in the golang regex library.

Indeed, negative lookahead won't work... Could we use something like `(^[^firstChar].*|firstChar[^secondChar].*|.....)`? Example here https://play.golang.org/p/SWuSDVhyMXT

akrejcir:
> Indeed, negative lookahead won't work... Could we use something like `(^[^firstChar].*|firstChar[^secondChar].*|.....)`? Example here https://play.golang.org/p/SWuSDVhyMXT

Good idea. I have changed the regex.

fgimenez:
@akrejcir looks good, can you please add a golang playground link with some examples? I don't see a clear way for us to test otherwise.

akrejcir:
Here it is: https://play.golang.org/p/-YoSs7PPia0
I have added all test cases I could think of.

fgimenez:
Thanks looks very good! 

In the diff I see several changes in the `always_run` fields of some jobs, we should remove them. 

For testing I think we could create a separate PR that adds a new presubmit with the pattern in the `run_if_changed` field and  these settings:
```
always_run: false
optional: true
skip_report: true
```
We should also change the command to something like "echo lane was executed" or something like that and reduce the memory requests to "10Mi" for instance. We can copy the presubmit from one of the existing ones (maybe the unit tests one) and perform the changes on the copied job, wdyt?

akrejcir:
I have created a PR with the test lane: https://github.com/kubevirt/project-infra/pull/1284

PR #1252, opened 2021-05-25
Prow workloads bootstrap automation

fgimenez:
Adds an ansible role with all the customizations required to setup e2e worker nodes. Also creates a prow job to run the job on tag push.

/cc @dhiller @rmohr


fgimenez:
/hold

Waiting to be able to run test job.

rmohr:
I guess for the cache nodes we will need an extra role. Looks good on a quick glance. Do we want to separate images and emptyDir space again like we did in the past?

fgimenez:
> Do we want to separate images and emptyDir space again like we did in the past?

Makes sense to me, I'll prepare a separate role for the cache node.

dhiller (approved):
/approve

fgimenez:
/retest

imagePullBackoff, maybe we need to increase the retries for pods ready

PR #1267, opened 2021-05-31
libguestfs-appliance: periodic: notify by email on failure

victortoso:
It'll send an email if the job fails or it stales for an hour.
Note that the last successful job took less than 10 minutes.

Signed-off-by: Victor Toso <victortoso@redhat.com>

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/libguestfs-appliance/libguestfs-appliance-periodics.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/libguestfs-appliance/libguestfs-appliance-periodics.yaml
@@ -5,7 +5,10 @@ periodics:
   # Run once a week at midnight on Sunday morning
   cron: "0 0 * * 0"
   annotations:
-    testgrid-create-test-group: "false"
+    testgrid-dashboards: kubevirt-libguestfs-appliance-periodics
+    testgrid-alert-email: afrosi@redhat.com, victortoso@redhat.com
+    testgrid-num-failures-to-alert: "1"
+    testgrid-alert-stale-results-hours: "1"
   decorate: true
   decoration_config:
     timeout: 3h
--- a/github/ci/testgrid/gen-config.yaml
+++ b/github/ci/testgrid/gen-config.yaml
@@ -1,9 +1,11 @@
 dashboards:
 - name: kubevirt-presubmits
 - name: kubevirt-periodics
+- name: kubevirt-libguestfs-appliance-periodics
 
 dashboard_groups:
 - name: kubevirt
   dashboard_names:
   - kubevirt-presubmits
   - kubevirt-periodics
+  - kubevirt-libguestfs-appliance-periodics


github/ci/prow-deploy/files/jobs/kubevirt/libguestfs-appliance/libguestfs-appliance-periodics.yaml
```diff
@@ -6,6 +6,9 @@ periodics:
   cron: "0 0 * * 0"
   annotations:
     testgrid-create-test-group: "false"
```
fgimenez, line 8: We need to create a test-group if we want to use testgrid's email capqabilities, the changes required for that are:
* add a dashboard `libguestfs-appliance-periodics` to the base config here https://github.com/kubevirt/project-infra/blob/master/github/ci/testgrid/gen-config.yaml
* add an annotation referring to that test-group in the job like here https://github.com/kubevirt/project-infra/blob/master/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml#L535 (instead of the current `testgrid-create-test-group: "false"`)

victortoso: Thanks for the review and the clarification. I'll update it Tomorrow morning (CEST time).

victortoso:
Changed dashboard name from `libguestfs-appliance-periodics` to `kubevirt-libguestfs-appliance-periodics` due previous failure in [pull-project-infra-check-testgrid-config](https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_project-infra/1267/pull-project-infra-check-testgrid-config/1399804591071039488) (I think this chage is enough, let's see)

```
config_test.go:196: Dashboard libguestfs-appliance-periodics: must prefix with one of: [[conformance kops presubmits sig wg provider kubernetes-clients] [canonical cos cri-o istio googleoss google knative kopeio redhat vmware gardener jetstack kyma kubevirt]]
```

victortoso:
@fgimenez, let me know if there is some other changes needed.

github/ci/testgrid/gen-config.yaml
```diff
@@ -1,9 +1,11 @@
 dashboards:
 - name: kubevirt-presubmits
 - name: kubevirt-periodics
+- name: kubevirt-libguestfs-appliance-periodics
```
fgimenez, line 4:
```suggestion
- name: libguestfs-appliance-periodics
```

github/ci/testgrid/gen-config.yaml
```diff
@@ -1,9 +1,11 @@
...
 
 dashboard_groups:
 - name: kubevirt
   dashboard_names:
   - kubevirt-presubmits
   - kubevirt-periodics
+  - kubevirt-libguestfs-appliance-periodics
```
fgimenez, line 11:
```suggestion
  - libguestfs-appliance-periodics
```

github/ci/prow-deploy/files/jobs/kubevirt/libguestfs-appliance/libguestfs-appliance-periodics.yaml
```diff
@@ -5,7 +5,10 @@ periodics:
   # Run once a week at midnight on Sunday morning
   cron: "0 0 * * 0"
   annotations:
-    testgrid-create-test-group: "false"
+    testgrid-dashboards: kubevirt-libguestfs-appliance-periodics
```
fgimenez, line 8:
```suggestion
    testgrid-dashboards: libguestfs-appliance-periodics
```

fgimenez (commented):
Looks good, but I think it would be better to keep the hierarchy that we have in github and prow jobs directories, we have a dashboard group called `kubevirt` (equivalent to github org) and under it dashboards following the format  `<gh-project-name>-<job-type>`, like `kubevirt-periodics`. Following this format I think the new dashboard should be named `libguestfs-appliance-periodics` wdyt?

github/ci/testgrid/gen-config.yaml
victortoso: I had an issue with that, something about needing a proper prefix which libguestfs doesn't fit, see this [comment](https://github.com/kubevirt/project-infra/pull/1267#issuecomment-852406685)

fgimenez: Sorry, didn't see your previous comment :) And didn't know about this requirement, then probably we need to change the name of the existing dashboards to something like `kubevirt-kubevirt-periodics`. In any case the changes in this PR are ok, thanks!

fgimenez (approved):
/approve

PR #1277, opened 2021-06-01
Fix testgrid config upload and simplify update workflow

fgimenez:
We were uploading a yaml file for testgrid config, a proto file is required instead. configurator generates the config in this format and is able to upload it too. These changes modify the uploader job to use the configurator image and also include a defaults file for the config, it is required by configurator and will allow us to fully customize the config. 

Given that we need to upload the config to GCS, instead of proposing a PR to have it in project-infra as we did before we directly upload it after the changes have been tested and merged. 

Successful execution here https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/logs/post-project-infra-update-testgrid-config/1399744356453191680, config proto file now available here gs://kubevirt-prow/testgrid/config and changes to test-infra to read our config from there proposed here https://github.com/kubernetes/test-infra/pull/22382

/cc @dhiller


fgimenez:
`pull-project-infra-check-testgrid-config` with the changes in this PR successful execution here https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_project-infra/1277/pull-project-infra-check-testgrid-config/1399756734888677376

github/ci/testgrid/hack/check.sh
```diff
@@ -3,16 +3,42 @@
...
+    mkdir -p "${testgrid_dir}/${testgrid_subdir}"
+
+    /configurator \
+        --prow-config "${prow_config}" \
+        --prow-job-config "${job_config}" \
+        --output-yaml \
+        --yaml "${testgrid_gen_config}" \
+        --default "${testgrid_default}" \
+        --oneshot \
+        --output "${testgrid_dir}/${testgrid_subdir}/gen-config.yaml"
```
dhiller, line 28: Maybe add `--verify-config-file` here as described in the [README](https://github.com/kubernetes/test-infra/blob/master/testgrid/cmd/configurator/README.md#basic-usage)? Or would that break the existing workflow?

dhiller (approved):
/approve
/hold

One nit I have commented on. Otherwise feel free to unhold.

github/ci/testgrid/hack/check.sh
fgimenez: It would not upload the file, according to the command help:
```
-validate-config-file
    	validate that the given config files are syntactically correct and exit (proto is not written anywhere)
```
In any case, we already have a job that checks the configuration before merging https://github.com/kubevirt/project-infra/blob/master/github/ci/prow-deploy/files/jobs/kubevirt/project-infra/project-infra-presubmits.yaml#L500 at this point all should be already verified.

PR #1282, opened 2021-06-02
Fix project_infra_root for workloads-production overlay

fgimenez:
Probably a leftover of local testing, it will prevent errors like https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/logs/post-project-infra-prow-workloads-cluster-deployment/1400104330970796032

/cc @dhiller 


dhiller (approved):
/approve

PR #1290, opened 2021-06-07
Add ingress controller to prow-workloads

fgimenez:
Required for accessing services in the prow-workloads cluster. 

The manifests come from upstream verbatim except for a change in the service type, from `LoadBalancer` to `NodePort` (with 30080 and 30443 as the `nodePorts` for HTTP and HTTPS respectively). This is needed for being able to receive incoming requests in the self-managed cluster.

/cc @rmohr 


rmohr:
/lgtm
/approve
/hold

fgimenez:
/unhold

Thanks, already deployed and tested, all working fine.

PR #1296, opened 2021-06-07
Set pushgateway resources

fgimenez:
We were losing some prow-related metrics during periods of high load, pushwateway was requesting more CPU than available and was evicted in some cases.

These changes set resources for pushgateway above the highest usage it had during the last 2 days https://grafana.ci.kubevirt.io/d/GlXkUBGiz/kubernetes-pod-overview?orgId=1&refresh=10s&var-namespace=kubevirt-prow&var-pod=pushgateway-64f6547bb7-gx9tv&var-container=All&from=now-2d&to=now It also fixes the QoS to Guaranteed (same memory and CPU requests and limits) to mitigate further issues.

/cc @dhiller 


dhiller (approved):
/approve

PR #1301, opened 2021-06-08
Add kubevirt/containerized-data-importer presubmits to testgrid

maya-r:


github/ci/testgrid/gen-config.yaml
```diff
@@ -1,11 +1,13 @@
 dashboards:
+- name: kubevirt-containerized-data-importer
```
dhiller, line 2: So you don't want to split dashboards for cdi as we do with kubevirt?

maya-r: if you mean separate between periodics and presubmits - cdi doesn't seem to have periodics

awels: Not right now, but we might in the future, I just added a post submit yesterday, and was considering a daily periodic while I did that. So it makes sense to prepare for that.

fgimenez (approved):
/approve

Looks good, thanks! 

Not sure how long will take config merger to update our config and show the new dashboard but the process is completely automated, will keep an eye on https://testgrid.k8s.io/kubevirt

PR #1335, opened 2021-06-16
Move jobs running on bare metals to prow-workloads

fgimenez:
Continues the migration to the new workloads cluster, sets all the jobs running on bare metals to execute on `prow-workloads`.

/cc @dhiller 


dhiller (approved):
/approve

PR #1341, opened 2021-06-17
Fix prow-deploy test

fgimenez:
Changes to prevent errors when running in `prow-workloads` cluster include:
* Use a more recent kubevirtci/prow-deploy image with kind v0.11.1 as suggested here https://github.com/kubernetes-sigs/kind/issues/2316#issuecomment-864182778, built with the changes in #1350 
* Bump kindest/node image used by kind (pointing to our quay mirror).
* Collect kind logs in case of cluster creation failure.

Additionally, all the executions with the new kind have taken less than 7min, previously it was taking ~10min.

Successful execution here https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_project-infra/1341/rehearsal-pull-project-infra-prow-deploy-test/1405963755489071104 

/cc @dhiller 

### DIFF
--- a/github/ci/prow-deploy/molecule/default/cleanup.yml
+++ b/github/ci/prow-deploy/molecule/default/cleanup.yml
@@ -14,3 +14,4 @@
       vars:
         main_actions:
           - cleanup
+      ignore_errors: yes
--- a/github/ci/prow-deploy/files/jobs/kubevirt/project-infra/project-infra-presubmits.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/project-infra/project-infra-presubmits.yaml
@@ -291,7 +291,7 @@ presubmits:
     labels:
       preset-dind-enabled: "true"
       preset-docker-mirror-proxy: "true"
-    cluster: phx-prow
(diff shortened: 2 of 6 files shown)


github/ci/prow-deploy/molecule/default/cleanup.yml
```diff
@@ -14,3 +14,4 @@
       vars:
         main_actions:
           - cleanup
+      ignore_errors: yes
```
dhiller, line 17: Why do we need that here and at other places?

fgimenez: Yes, these are added to cleanup and collect actions, we need them to allow all the actions in these tasks to complete, we can't collect all the logs when the cluster fails to spin up otherwise. As I understand it they won't mask any potential problem being tasks related to cleaning up and log collection after the tests, wdyt?

dhiller: OK, thanks for the explanation.

dhiller (approved):
/approve

PR #1346, opened 2021-06-18
kind-1.19-sriov jobs: Cancel always_run

ormergi:
This PR chance 1.19 sriov jobs to **not** always run in order to prevent extra pressure on CI.

We will change it back once we set 1.19 sriov jobs as voting lanes.


fgimenez (approved):
/approve

PR #1352, opened 2021-06-21
ovs-cni, Add postsubmits

RamLavi:
This commit adds 2 jobs:
- release-ovs-cni: image-push of tags to qauy
- main-ovs-cni: image-push of main branch to quay, using virtual-
  tags as the image tag.

### DIFF
--- a/github/ci/prow-deploy/files/jobs/k8snetworkplumbingwg/ovs-cni/ovs-cni-postsubmits.yaml
+++ b/github/ci/prow-deploy/files/jobs/k8snetworkplumbingwg/ovs-cni/ovs-cni-postsubmits.yaml
@@ -1,0 +1,52 @@
+postsubmits:
+  k8snetworkplumbingwg/ovs-cni:
+    - name: main-ovs-cni
+      annotations:
+        testgrid-create-test-group: "false"
+      decorate: true
+      max_concurrency: 1
+      labels:
+        preset-dind-enabled: "true"
+        preset-docker-mirror-proxy: "true"
+        preset-kubevirtci-quay-credential: "true"
+      branches:
+        - main
+      cluster: prow-workloads
+      spec:
+        containers:
+          - image: quay.io/kubevirtci/bootstrap:v20201119-a5880e0
+            command:
+              - "/usr/local/bin/runner.sh"
+              - "/bin/bash"
+              - "-c"
+              - "cat $QUAY_PASSWORD | docker login --username $(cat $QUAY_USER) --password-stdin=true quay.io && make docker-build docker-push"
+            # docker-in-docker needs privileged mode
+            securityContext:
+              privileged: true
+    - name: release-ovs-cni
+      annotations:
+        testgrid-create-test-group: "false"
+      decorate: true
+      max_concurrency: 1
+      labels:
+        preset-dind-enabled: "true"
+        preset-docker-mirror-proxy: "true"
+        preset-kubevirtci-quay-credential: "true"
+      branches:
+        - ^v\d+\.\d+\.\d+$
+      cluster: prow-workloads
+      spec:
+        containers:
+          - image: quay.io/kubevirtci/bootstrap:v20201119-a5880e0
+            command:
+              - "/usr/local/bin/runner.sh"
+              - "/bin/bash"
+              - "-ce"
+              - |
+                cat $QUAY_PASSWORD | docker login --username $(cat $QUAY_USER) --password-stdin=true quay.io
+                # Only push images on tags
+                COMMIT_TAG=$(git tag --points-at HEAD | head -1)
+                [ -z "$COMMIT_TAG" ] || make docker-build docker-push IMAGE_TAG=$COMMIT_TAG
+            # docker-in-docker needs privileged mode
+            securityContext:
+              privileged: true


RamLavi:
/hold
until a PR to add virtual-tag docker-tag on ovs-cni is in place

github/ci/prow-deploy/files/jobs/k8snetworkplumbingwg/ovs-cni/ovs-cni-postsubmits.yaml
```diff
@@ -0,0 +1,53 @@
...
+      cluster: prow-workloads
+      spec:
+        containers:
+          - image: quay.io/kubevirtci/bootstrap:v20201119-a5880e0
+            command:
+              - "/usr/local/bin/runner.sh"
+              - "/bin/bash"
+              - "-c"
+              - |
+                cat $QUAY_PASSWORD | docker login --username $(cat $QUAY_USER) --password-stdin=true quay.io &&
```
oshoval, line 46: you dont need && since you have `- |`
but it would be better to add "-ce" on line 41
else if this line fail , following lines might not fail , and the job would look like its done fine but it isn't

github/ci/prow-deploy/files/jobs/k8snetworkplumbingwg/ovs-cni/ovs-cni-postsubmits.yaml
```diff
@@ -0,0 +1,53 @@
...
+                # Only push images on tags
+                COMMIT_TAG=$(git tag --points-at HEAD | head -1)
+                [ -z "$COMMIT_TAG" ] ||
```
oshoval, line 49: please make those one liner ( no need enter after `||` )

oshoval (changes requested):

github/ci/prow-deploy/files/jobs/k8snetworkplumbingwg/ovs-cni/ovs-cni-postsubmits.yaml
RamLavi: Done

github/ci/prow-deploy/files/jobs/k8snetworkplumbingwg/ovs-cni/ovs-cni-postsubmits.yaml
RamLavi: Done

RamLavi:
@oshoval fixes [here](https://github.com/kubevirt/project-infra/compare/c54ea57205d6ea8741f3a2979cd2556b80b2736a..3804a80544c054b54e500673587bcca803564582)

github/ci/prow-deploy/files/jobs/k8snetworkplumbingwg/ovs-cni/ovs-cni-postsubmits.yaml
```diff
@@ -0,0 +1,52 @@
...
+          - image: quay.io/kubevirtci/bootstrap:v20201119-a5880e0
+            command:
+              - "/usr/local/bin/runner.sh"
+              - "/bin/bash"
+              - "-ce"
+              - |
+                cat $QUAY_PASSWORD | docker login --username $(cat $QUAY_USER) --password-stdin=true quay.io
+                # Only push images on tags
+                COMMIT_TAG=$(git tag --points-at HEAD | head -1)
+                [ -z "$COMMIT_TAG" ] || make docker-build docker-push IMAGE_TAG=$COMMIT_TAG
```
oshoval, line 49: lets think that i didn't suggest noise / that the logic is fine

what happen if `[ -z "$COMMIT_TAG" ]` is false now ?
is it legit that it happen ? and if so what is the returned value and behavior

RamLavi: well if the COMMIT_TAG is empty string then we still want to push the image, no?
Let's consult with @qinqon

qinqon: well if it's not tagged the job will not kick in since it only kicks in on tags.

oshoval: so logic should be changed here and /or on
https://github.com/k8snetworkplumbingwg/ovs-cni/pull/166/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R73

accordingly, 
for example remove here `[ -z "$COMMIT_TAG" ] ||`
and on the Makefile, do the first push only if `IMAGE_TAG` is not empty

RamLavi: the job won't kick in ..
the `branches` field is also referring to tags, so only commits with matching tag regex `^v\d+\.\d+\.\d+$` will initiate the job
But I agree that the logic is perhaps flawed still 
I'm leaning to changing it to `&&`  or removing the `[ -z "$COMMIT_TAG" ] ||` at all

oshoval: At the end as we discussed f2f all is fine, since this job is the one for the tags / branches only,
the regex is just for tags actually vXXX
so the COMMIT_TAG won't be empty (and even if so theoretically its just a guard)

oshoval (approved):

RamLavi:
Hey @fgimenez 
can you review the PR?

RamLavi:
/hold cancel
https://github.com/k8snetworkplumbingwg/ovs-cni/pull/166 got merged. 
We're ready to merge this PR

fgimenez (approved):
/approve

oshoval:
Little question
can it be that the two jobs run for the same time for the same commit, and because the new code of https://github.com/k8snetworkplumbingwg/ovs-cni/pull/166
they both push the same hash at the same time, and docker will block one (since the other is in the middle)
and we will see a failing job ?

PR #1364, opened 2021-06-24
[DRAFT] add periodics cluster sync test for ARM

zhlhahaha:
Hi @fgimenez @rmohr 

I add a periodic task for ARM, please take a look when you have time.
There are two additional requests to make this job work.
1. The kubeconfig of ARM server needs to be upload and sorted.
2. I am not sure if it is appropriate to use quay.io/kubevirt to store test images. If not which repository we can use to store test images.

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml
@@ -1131,3 +1131,93 @@ periodics:
       resources:
         requests:
           memory: "8Gi"
+- name: periodic-kubevirt-cluster-sync-test-ARM
+  annotations:
+    testgrid-dashboards: kubevirt-periodics
+    testgrid-days-of-results: "60"
+  cron: "40 4,12,20 * * *"
+  decorate: true
+  cluster: prow-workloads
+  decoration_config:
+    timeout: 4h
+    grace_period: 5m
+  labels:
+    preset-dind-enabled: "true"
+    preset-docker-mirror-proxy: "true"
+    preset-shared-images: "true"
+    preset-bazel-cache: "true"
+    preset-bazel-unnested: "true"
+    preset-kubevirtci-quay-credential: "true"
+  extra_refs:
+  - org: kubevirt
+    repo: kubevirt
+    base_ref: main
+    work_dir: true
+  - org: kubevirt
+    repo: project-infra
+    base_ref: main
+  reporter_config:
+    slack:
+      job_states_to_report: []
+  spec:
+    nodeSelector:
+      type: bare-metal-external
+    containers:
+      - image: quay.io/kubevirtci/bootstrap:v20210311-09ebaa2
+        env:
+          - name: TARGET
+            value: "external"
+          - name: KUBEVIRT_PROVIDER
+            value: "external"
+          - name: DOCKER_PREFIX
+            value: quay.io/kubevirtci
+          - name: DOCKER_TAG
+            value: "aarch64_test"
+          - name: KUBECONFIG
+            value: "/kubeconfig"
+          - name: IMAGE_PULL_POLICY
+            value: "Always"
+          - name: BUILD_ARCH
+            value: "crossbuild-aarch64"
+          - name: KUBEVIRT_E2E_FOCUS
+            value: "test_id:1464"
+          - name: GIT_ASKPASS
+            value: "../project-infra/hack/git-askpass.sh"
+        command:
+          - "/usr/local/bin/runner.sh"
+          - "/bin/sh"
+          - "-c"
+          - |
+            # install yq
+            curl -Lo ./yq https://github.com/mikefarah/yq/releases/download/3.4.1/yq_linux_amd64
+            chmod +x ./yq && mv ./yq /usr/local/bin/yq
+
+            # get kubeconfig
+            source ../project-infra/hack/manage-secrets.sh
+            decrypt_secrets
+            extract_secret 'kubeconfigARM' /kubeconfig
+
+            # login quay.io
+            cat "$QUAY_PASSWORD" | docker login --username $(cat "$QUAY_USER") --password-stdin=true quay.io
+
+            # run the test
+            automation/test.sh
+        # docker-in-docker needs privileged mode
+        securityContext:
+          privileged: true
+        resources:
+          requests:
+            memory: "29Gi"
+        volumeMounts:
+          - name: token
+            mountPath: /etc/github
+          - name: pgp-bot-key
+            mountPath: /etc/pgp
+            readOnly: true
+    volumes:
+      - name: token
+        secret:
+          secretName: oauth-token
+      - name: pgp-bot-key
+        secret:
+          secretName: pgp-bot-key


fgimenez:
> 1. The kubeconfig of ARM server needs to be upload and sorted.

Sure, I'll add the secret to our config shortly, in order to read it you need to:
* Mount the pgp-bot-key secret like here https://github.com/kubevirt/project-infra/blob/main/github/ci/prow-deploy/files/jobs/kubevirt/project-infra/project-infra-postsubmits.yaml#L484-L486 and https://github.com/kubevirt/project-infra/blob/main/github/ci/prow-deploy/files/jobs/kubevirt/project-infra/project-infra-postsubmits.yaml#L477-L479 
* Extract the kubeconfig secret like here https://github.com/kubevirt/project-infra/blob/main/github/ci/prow-deploy/files/jobs/kubevirt/project-infra/project-infra-postsubmits.yaml#L463-L465 instead of `kubeconfig` the key should be `kubeconfigARM`

> 2. I am not sure if it is appropriate to use quay.io/kubevirt to store test images. If not which repository we can use to store test images.

I think quay.io/kubevirtci is more appropriate for test images, which repositories do you need and from where are the images pushed?

zhlhahaha:
> I think quay.io/kubevirtci is more appropriate for test images, which repositories do you need and from where are the images pushed?

Thanks, Federio, Roman suggests I create a registry inside the arm cluster. I will give it a try.

zhlhahaha:
> Thanks, Federio, Roman suggests I create a registry inside the arm cluster. I will give it a try.

Hi, @rmohr I give it a try. To make other servers can push images to the registry, I need to put the registry address into docker deamon.json, like this:
`insecure-registries":["192.168.99.100:5000"]`

However, it seems that bazel rules_docker has not supported "insecure-registries" yet, and here is a PR for it. https://github.com/bazelbuild/rules_docker/pull/1403

Can we use repository `quay.io/kubevirtci`, or you have any other suggestions?

zhlhahaha:
Hi @rmohr @fgimenez 
I find a way that can successfully push images to a private repository.
1. generate private SSL certification
```
openssl req \
  -newkey rsa:4096 -nodes -sha256 -keyout /certs/domain.key \
  -addext "subjectAltName = IP:server ip" \
  -x509 -days 365 -out /certs/domain.crt
```
2. create a private repository with SSL certification:
```
 docker run -d \
  --restart=always \
  --name registry \
  -v /certs:/certs \
  -e REGISTRY_HTTP_ADDR=0.0.0.0:443 \
  -e REGISTRY_HTTP_TLS_CERTIFICATE=/certs/domain.crt \
  -e REGISTRY_HTTP_TLS_KEY=/certs/domain.key \
  -p 5000:443 \
  registry:2
```
3. pass the `domain.crt` to environmental variable `SSL_CERT_FILE` before using `./hack/bazel-push-images.sh`
```
./hack/dockerized "export SSL_CERT_FILE=/root/go/src/kubevirt.io/kubevirt/certs/domain.crt && ./hack/bazel-push-images.sh"
...
2021/06/28 05:51:03 Successfully pushed Docker image to 192.168.100.101:5000/virt-handler:aarch64
...
```

But here are some concerns:
1. SSL certification.
    The private SSL certification is supposed to exist in the ARM cluster and the x86 cluster. Rather than generate it every time the test run, it would be good to generate one and put it into secrets like kubeconfig.

2. IP address of arm cluster.
It would be better to hide the IP address of the ARM cluster, now it would be showed in the test log.

rmohr:
> I find a way that can successfully push images to a private repository.

Using a private certificate sounds good to me.

> 1. SSL certification.
>    The private SSL certification is supposed to exist in the ARM cluster and the x86 cluster. Rather than generate it every time the test run, it would be good to generate one and put it into secrets like kubeconfig.

Sounds good. Did you also check https://docs.docker.com/registry/deploying/#native-basic-auth? We can put that in an extra secret.

> 2\. IP address of arm cluster.
>     It would be better to hide the IP address of the ARM cluster, now it would be showed in the test log.

If we talk about a public accessible port, let's consider securing that properly. I am not sure if I feel comfortable by just hiding the IP.  

Maybe after all we should follow the suggestion from @fgimenez and use the `kubevirtci` org. Just a lot of images are generated there and it  does not feel correct to push the images there just for a test-run.

zhlhahaha:
> Maybe after all we should follow the suggestion from @fgimenez and use the `kubevirtci` org. Just a lot of images are generated there and it does not feel correct to push the images there just for a test-run.

Ok, I will use `kubevirtci` for now.

> I think quay.io/kubevirtci is more appropriate for test images, which repositories do you need and from where are the images pushed?

Hi @fgimenez, I need the quay.io/kubevirtci repositories. The images is pushed from the periodic jobs.
I will set the following environmental variables:
```
TARGET=external
DOCKER_PREFIX=quay.io/kubevirtci
DOCKER_TAG=aarch64_test
IMAGE_PULL_POLICY=Always
BUILD_ARCH=crossbuild-aarch64
KUBEVIRT_E2E_FOCUS="test_id:1464"
```
Then run `automatic/test.sh`
`make cluster-sync` will automatically push the images to quay.io/kubevirtci.

fgimenez:
> Hi @fgimenez, I need the quay.io/kubevirtci repositories

Sure, will ping you when ready.

fgimenez:
@zhlhahaha the repositories are ready, let me know if you find any issue with them.

zhlhahaha:
> @zhlhahaha the repositories are ready, let me know if you find any issue with them.

Thanks, Federico. Would you mind reviewing this PR? I have updated the periodic job.

github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml
```diff
@@ -1131,3 +1131,84 @@ periodics:
...
+          - "/usr/local/bin/runner.sh"
+          - "/bin/sh"
+          - "-c"
+          - |
+            # install yq
+            curl -Lo ./yq https://github.com/mikefarah/yq/releases/download/3.4.1/yq_linux_amd64
+            chmod +x ./yq && mv ./yq /usr/local/bin/yq
+
+            # get kubeconfig
+            git clone https://github.com/kubevirt/project-infra.git
```
fgimenez, line 1191: Instead of cloning I think we should add an entry in `extra_refs` for `kubevirt/project-infra`, and refer to `../project-infra/...`

github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml
```diff
@@ -1131,3 +1131,84 @@ periodics:
...
           memory: "8Gi"
+- name: periodic-kubevirt-cluster-sync-test-ARM
+  annotations:
+    testgrid-dashboards: kubevirt-periodics
+    testgrid-days-of-results: "60"
+  cron: "40 4,12,20 * * *"
+  decorate: true
+  cluster: prow-workloads
+  decoration_config:
+    timeout: 24h
```
fgimenez, line 1142: This is too big, maybe better 4h?

fgimenez (commented):
Thanks, we are going to need changes in order to be able to decrypt the secrets:
* Declare github oauth token volume like here https://github.com/kubevirt/project-infra/blob/main/github/ci/prow-deploy/files/jobs/kubevirt/project-infra/project-infra-postsubmits.yaml#L270-L272
* Mount Github oauth token like here https://github.com/kubevirt/project-infra/blob/main/github/ci/prow-deploy/files/jobs/kubevirt/project-infra/project-infra-postsubmits.yaml#L261-L262
* Declare `GIT_ASKPASS` like here https://github.com/kubevirt/project-infra/blob/main/github/ci/prow-deploy/files/jobs/kubevirt/project-infra/project-infra-postsubmits.yaml#L246-L247

fgimenez:
Thanks looks good, I've triggered the job here https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/logs/periodic-kubevirt-cluster-sync-test-ARM/1410524703319134208

fgimenez:
Failed here https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/logs/periodic-kubevirt-cluster-sync-test-ARM/1410524703319134208#1:build-log.txt%3A1771 with:
```
cp: cannot stat 'bazel-out/k8-fastbuild/bin/tools/manifest-templator/templator_/templator': No such file or directory
```
@rmohr wdyt?

rmohr:
> @rmohr wdyt?

I think that the issu are the build targets in this file: https://github.com/kubevirt/kubevirt/blob/main/hack/build-func-tests.sh
They all compile the tools for `${ARCHITECTURE}`. This is not the right architecture when cross-compiling and testing on a cluster with a different architecture. They all need to use `${HOST_ARCHITECTURE}` instead.

fgimenez:
Cool thx, proposed here https://github.com/kubevirt/kubevirt/pull/5961

fgimenez:
/approve
/hold

The job is passing here https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/logs/periodic-kubevirt-cluster-sync-test-ARM/1410592538611421184 with the changes from https://github.com/kubevirt/kubevirt/pull/5961 holding until it is merged.

fgimenez:
/unhold

https://github.com/kubevirt/kubevirt/pull/5961 is merged

PR #1373, opened 2021-06-24
Run sig-storage periodics with rook-ceph-default storage.

maya-r:
After the removal of the ceph periodic lane in #1370,
no periodics run ceph-dependent tests.

That's mostly a storage concern, let's add those to the
sig-storage lane.


maya-r:
It's not clear for me how it happens but it already uses rook-ceph.

PR #1381, opened 2021-06-25
Make sure to use Python3 in prow-deploy

fgimenez:
prow-deploy image already ships kubernetes and openshift modules, but installed with the python3 interpreter. We needed to install them again in the test execution because the interpreter used was python2.7. Now this python version is deprecated and the pip installation during the test started failing like here https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_project-infra/1380/pull-project-infra-prow-deploy-test/1408335642642878464

With these changes we make sure that python3 is used both for testing and deployment. The pip installation during testing is no longer needed, we use the modules installed in the image.

/cc @dhiller 


dhiller (approved):
/approve

PR #1382, opened 2021-06-27
periodics, 1.21-sig-network, remove QUARANTINE env var

oshoval:
Removing QUARANTINE env var results in skipping QUARANTINE tests.

Since those tests are flaky and quarantined we don't
need to run them in the periodic, so it will reflect the true
state of the job.

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml
@@ -581,8 +581,6 @@ periodics:
     containers:
       - image: quay.io/kubevirtci/bootstrap:v20210311-09ebaa2
         env:
-          - name: KUBEVIRT_QUARANTINE
-            value: "true"
           - name: TARGET
             value: "k8s-1.21-sig-network"
         command:


fgimenez:
/hold

We can't do this, we need to run quarantined tests in the periodic lanes to know when they are fixed.

oshoval:
@fgimenez 
In my opinion it is prone for causing other failures, since 1.21 doesnt run yet on each PR, we need to have a quick status of the lane.
Its not that we can / should review the lane every day and see if new tests failed / what was fixed.

One thing we can do in the middle is XFAIL, so we will still see the lane green unless something else happened,
and also see the test there XFAILed instead of FAIL
wdyt ?

fgimenez:
> In my opinion it is prone for causing other failures, since 1.21 doesnt run yet on each PR, we need to have a quick status of the lane.

You can know the status of the lane if it also executed the quarantined tests, just reading the results, not only checking top level green/red

> Its not that we can / should review the lane every day and see if new tests failed / what was fixed.

Totally, I do that now. And, as soon as 1.21 lanes start to run on each PR, everyone will do.

> Its network team responsibility to fix the Quarantine tests and to know the status of them.

No, it is network team responsibility to fix the Quarantine tests that are marked as sig-network. The status should be open to everyone, it is network team responsibility to know the status to check if they are fixed but it is also my responsibility.

> One thing we can do in the middle is XFAIL, so we will still see the lane green unless something else happened,
and also see the test there XFAILed instead of FAIL

With XFAIL you would be executing them anyway, what is the difference? As I said before, there's no problem with having the quarantined tests failing and the whole lane in red. You, as the rest of teams do, just need to check which tests failed, if they are quarantine tests you just need to take them into account when fixing them.

oshoval:
> You can know the status of the lane if it also executed the quarantined tests, just reading the results, not only checking top level green/red

I think we should strive to let the tools tells us as much info as we need in glance, 
only in the times that we need to know more details we will dive in, 
the current state enforce us to dive in even if we just want to know that there is no regression.
The lane was broken for month+ due to that.

> With XFAIL you would be executing them anyway, what is the difference? 

They will be still Quarantine and XFAIL, but the lane will be green unless something else happened.
So they won't run on presubmit (as now), and they will still run on periodic as now, but won't cause the whole job to be red if 
only Q tests failed.

fgimenez:
> The lane was broken for month+ due to that.

Well, one of the reasons why the lane was broken is broken quarantine tests, as soon as they are fixed the lane will be green. But again, it is just a click away and a bit of reading to know why the lane is failing at any time.  

> They will be still Quarantine and XFAIL, but the lane will be green unless something else happened.

I don't understand why it is so important to have this concrete lane green, it doesn't mean anything. As I see it, It is not a good reason for changing the way we manage quarantined tests.

oshoval:
> > The lane was broken for month+ due to that.
> 
> Well, one of the reasons why the lane was broken is broken quarantine tests, as soon as they are fixed the lane will be green. But again, it is just a click away and a bit of reading to know why the lane is failing at any time.
> 
> > They will be still Quarantine and XFAIL, but the lane will be green unless something else happened.
> 
> I don't understand why it is so important to have this concrete lane green, it doesn't mean anything. As I see it, It is not a good reason for changing the way we manage quarantined tests.

The lane was broken because a real bug, that was "masked" and postphoned due to the fact that the lane was less checked, 
imho one of the things that are prone to cause a lane to be less checked, is if it need extra clicks, especially when we can give it with the top view.
This is the answer for both quotes above imho.

fgimenez:
> The lane was broken because a real bug, that was "masked" and postphoned due to the fact that the lane was less checked,
> imho one of the things that are prone to cause a lane to be less checked, is if it need extra clicks, especially when we can give it with the top view.
> This is the answer for both quotes above imho.

No, I don't think so. The goal of periodics from master was from the beginning executing the whole suite, including quarantined tests. This is the main difference between quarantined and skipped tests, we keep executing quarantined tests to still get feedback from them and, most importantly, for knowing when they are fixed, but in a way that they don't interfere with the critical path to merge PRs. If we stop running quarantined tests in periodic lanes, we can't know when they are fixed, they would be the same as skipped tests and the test suite would erode over time. Please take a look at the quarantine document we agreed to handle this process https://github.com/kubevirt/kubevirt/blob/main/docs/quarantine.md

I think that checking less a lane because it is red is a real problem. Lanes that report errors should be checked more in my opinion, maybe we are too used to `/retest` and this is very bad as I see it.

oshoval:
> > The lane was broken because a real bug, that was "masked" and postphoned due to the fact that the lane was less checked,
> > imho one of the things that are prone to cause a lane to be less checked, is if it need extra clicks, especially when we can give it with the top view.
> > This is the answer for both quotes above imho.
> 
> No, I don't think so. The goal of periodics from master was from the beginning executing the whole suite, including quarantined tests. This is the main difference between quarantined and skipped tests, we keep executing quarantined tests to still get feedback from them and, most importantly, for knowing when they are fixed, but in a way that they don't interfere with the critical path to merge PRs. If we stop running quarantined tests in periodic lanes we can't know when they are fixed, they would be the same as skipped tests and the test suite would erode over time. Please take a look at the quarantine document we agreed to handle this process https://github.com/kubevirt/kubevirt/blob/main/docs/quarantine.md

Xfail will let us benefit from both of the worlds

 
> I think that checking less a lane because it is red is a real problem. Lanes that report errors should be checked more in my opinion, maybe we are too used to `/retest` and this is very bad as I see it.

Again as said above, its not that imho, its that we should make our life easier when possible, by making the indications more friendly.
Xfail will allow the ones that want to dive into all the details still able to, while not enforcing the ones that want to know in glance the status, to dive into the lane.
Current mode can mask real failures that happen less frequently, unless instance by instance is checked on the periodic, it has nothing to do with the retest that we run on the non periodics (beside if those retest break stuff, but this not what i meant).

fgimenez:
> Xfail will let us benefit from both of the worlds

Not sure what you mean, sorry. Every team is free to handle their tests as they wish, you can mark whatever test as XFail if that is what your team decides.

Currently the errors in the periodics from master are meaningful. When we have an error we know that there is something to fix, being it quarantined tests, bugs, etc. If we mark the quarantined tests as XFail we would get greens that are no longer meaningful, there are still things to fix (at least the quarantined tests), but the lane success conveys a false message saying that all is ok. This is why I think it would be a bad idea, just my opinion in any case.

oshoval:
Thanks, closing

PR #1397, opened 2021-06-30
Add presubmit jobs for cgroups v2 CI lanes

vasiliy-ul:
The PR introduces two presubmit jobs for kubevirtci. They check cluster-up for the case when cgroupv2 needs to be configured.

**(!)** Should be merged after https://github.com/kubevirt/kubevirtci/pull/629


vasiliy-ul:
@rmohr , FYI
Probably it makes sense to check the cgroup lanes in kubevirtci presubmit. What do you think?

oshoval (commented):
Maybe better to adapt check-provision-k8s-1.20-cgroupsv2 instead ?

vasiliy-ul:
In fact `check-provision-k8s-1.20-cgroupsv2` will not be needed anymore. Now its optional, I will probably remove it then.

oshoval:
This why i suggested to adapt it instead, so the actual check will be in line with current presubmit jobs
i.e - check-provision which are the ones that are running and mandatory

wdyt ?

vasiliy-ul:
Hm... I just found that there was e.g. `check-up-kind-1.19-sriov` job. So decided to make cgroupsv2 ones in the same manner. So are you suggesting to just keep the `check-provision-k8s-1.20-cgroupsv2` name and make it call `cluster-up` instead?

oshoval:
> Hm... I just found that there was e.g. `check-up-kind-1.19-sriov` job. So decided to make cgroupsv2 ones in the same manner. So are you suggesting to just keep the `check-provision-k8s-1.20-cgroupsv2` name and make it call `cluster-up` instead?

It already calls cluster-up at the end of it
it provisions and then check-up the cluster it creates, you just need please to adapt it to your cgroups change IIUC

vasiliy-ul:
Right, cluster-up is called when the provision is checked. But that is the point here: cgroupsv2 lanes do not need to be re-provisioned separately. They just reuse the already provisioned cluster and configure it when cluster-up is called. Doing one more provisoon here will be redundant IMHO.

oshoval:
You right, you can update the cluster up maybe then to do check both variants ?

vasiliy-ul:
do you mean modifying existing `check-provision-k8s-1.20` job and add a call to `cluster-up` which will check cgroups after the provision?

oshoval:
You just need to modify this file cluster-provision/k8s/check-cluster-up.sh on kubevirtci , wdyt?
This file is the one the cluster-up after provision,
so the lanes will provision the cluster and check it up twice, with or without cgroupsv2, IIUC

vasiliy-ul:
hm... `cluster-provision/k8s/check-cluster-up.sh` is a generic script and it is called for all the lanes. even for those that do not support cgroups v2. IMHO adding smth like
```bash
if "${CGROUPV2}" = "true"; then 
    # do some cgroupv2 stuff
fi
```
is not a good option in terms of maintenability since it may become messy when more similar cases are added.

oshoval:
Kubevirtci should check CGROUPV2 after a new provider is created and before the PR is merged,
so its either to do what you just wrote, or to add if according k8s-X, X >= 1.20 (we do this on couple places)

3rd option if this what you meant, is to add another line on the job that will know to cluster-up and resue the freshly temporally provided cluster, 
but imho its less maintainable, because all the logic already exists on `cluster-provision/k8s/check-cluster-up.sh`

vasiliy-ul:
ah... right, we need the new not-yet-merged cluster in order to check cluster-up. I probably missed that point... sorry for the confusion from my side. That definitely needs to be considered. I will take one more look on how to handle that.

vasiliy-ul:
Probably this PR does not make much sense then. Closing it for now. The cluster-up check should be done within existing provision jobs. @oshoval , thanks for pointing to that.

PR #1400, opened 2021-07-01
Deploy Prow exporter

fgimenez:
Prow exporter from https://github.com/kubernetes/test-infra/tree/master/prow/cmd/exporter is not included in Prow's manifests https://github.com/kubernetes/test-infra/tree/master/config/prow/cluster but the image is being built with the rest of the components. 

The metrics exposed by Prow exporter are very useful for us, especially the ones related to job durations. With these changes we:
* Create manifests for Prow's exporter deployment, service, and service monitor, including them in our Continuous Delivery pipeline.
* Modify the script to bump Prow version to include the exporter so that the version is kept in sync with the rest of Prow components.
* Add the ability to deploy different service monitors to each environment. For instance, greenhouse service monitor should be deployed in both control plane and workloads cluster, but Prow's pushgateway and exporter service monitor should only be deployed on control plane cluster. This will be more relevant when we enable alerts.

/cc @rmohr @dhiller 


dhiller (approved):
/approve

dhiller:
Remove hold whenever ready :)

fgimenez:
/unhold

All good thanks!

PR #1414, opened 2021-07-05
Fix cgroupv2 jobs

vasiliy-ul:
The PR fixes the periodic and presubmit cgroupsv2 jobs considering the approach from https://github.com/kubevirt/kubevirtci/pull/638

**NOTE**: Also (if no objections) it will mark the presubmit jobs in kubevirtci as mandatory.

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml
@@ -551,7 +551,9 @@ periodics:
       - image: quay.io/kubevirtci/bootstrap:v20210311-09ebaa2
         env:
           - name: TARGET
-            value: "k8s-1.20-cgroupsv2"
+            value: "k8s-1.20"
+          - name: KUBEVIRT_CGROUPV2
+            value: "true"
           - name: KUBEVIRT_E2E_SKIP
             value: "Multus|SRIOV|GPU|Macvtap|\\[sig-operator\\]"
         command:
--- a/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
@@ -382,7 +382,9 @@ presubmits:
         - automation/test.sh
         env:
         - name: TARGET
-          value: k8s-1.20-cgroupsv2
+          value: k8s-1.20
+        - name: KUBEVIRT_CGROUPV2
+          value: "true"
         - name: KUBEVIRT_E2E_SKIP
           value: Multus|SRIOV|GPU|Macvtap|\[sig-operator\]
         image: quay.io/kubevirtci/bootstrap:v20210311-09ebaa2
--- a/github/ci/prow-deploy/files/jobs/kubevirt/kubevirtci/kubevirtci-presubmits.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/kubevirtci/kubevirtci-presubmits.yaml
@@ -164,8 +164,8 @@ presubmits:
           requests:
             memory: "8Gi"
   - name: check-provision-k8s-1.20-cgroupsv2
-    always_run: false
-    optional: true
+    always_run: true
+    optional: false
     decorate: true
     decoration_config:
       timeout: 3h


oshoval (commented):
> NOTE: Also (if no objections) it will mark the presubmit jobs in kubevirtci as mandatory.
I don't think we want to double provision, which is what it does right ?

the conformance is pretty long,

instead maybe worth that the post provision will check both variants ?
or maybe we can check just one of them, no need to check every matrix of configuration

vasiliy-ul:
Sure, I do not insit on having both. I can enable only one. Just to have some checking that its not broken.

oshoval:
you can use rehearse in order to run it (at least the presubmit one)

update the spec , for example add `echo bla` 
and then write `/rehearse` as comment here

vasiliy-ul:
Sure, thanks for the hints. I think the presubmits in kubevirtci are fine. At least I tried to run them manually in https://github.com/kubevirt/kubevirtci/pull/638 (both suceeded).

rmohr:
/lgtm
/approve
/hold

Can go in once https://github.com/kubevirt/kubevirtci/pull/638 is propagated to kubevirt/kubevirt

rmohr:
> and then write `/rehearse` as comment here

that does sadly right now not work for periodics.

oshoval:
> > and then write `/rehearse` as comment here
> 
> that does sadly right now not work for periodics.

sure, meant just for the presubmits (see please comment first line)

oshoval:
> > NOTE: Also (if no objections) it will mark the presubmit jobs in kubevirtci as mandatory.
> > I don't think we want to double provision, which is what it does right ?
> 
> the conformance is pretty long,
> 
> instead maybe worth that the post provision will check both variants ?
> or maybe we can check just one of them, no need to check every matrix of configuration

@rmohr wdyt about this?

vasiliy-ul:
@rmohr , @oshoval regarding the presubmit checks in `kubevirtci`: I think makes sense to have at least one `1.20` as mandatory. Currently the cgroupsv2 periodic is based on 1.20 and probably it would be beneficial to ensure that it does not get broken accidentaly. The other one `1.21` - we can probably enable it later when it makes more sense.

oshoval:
> @rmohr , @oshoval regarding the presubmit checks in `kubevirtci`: I think makes sense to have at least one `1.20` as mandatory. Currently the cgroupsv2 periodic is based on 1.20 and probably it will be beneficial to ensure that it does not get broken accidentaly. The other one `1.21` - we can probably enable it later when it makes more sense.

of course, just wonder if it should be on its own lane if i understand you correctly, or that the same provision will spin up the two clusters variants (much faster and this is what i think we should do).
the cluster-up doesnt consume a lot of time, the stuff that consume a lot of time is the conformance and the provision which are common between the two variants

vasiliy-ul:
In this case we have a common provision, right. Then we have two different cluster-up's: a regular one and one for cgroupsv2. In my understanding since we have two different cluster-up's we need to do conformance for each of them. But frankly speaking I do not know what exactly the conformance step is checking. Therefore maybe a stupid question, but... are you telling that conformance testing does not depend on the cgroup version used within the cluster and we can run it only once?

oshoval:
> In this case we have a common provision, right. Then we have two different cluster-up's: a regular one and one for cgroupsv2. In my understanding since we have two different cluster-up's we need to do conformance for each of them. But frankly speaking I do not know what exactly the conformance step is checking. Therefore maybe a stupid question, but... are you telling that conformance testing does not depend on the cgroup version used within the cluster and we can run it only once?

basically you are right that it depends on the cluster-up instance, so forget what i say
maybe just a by product is that it wont actually affect but we cant know for sure

vasiliy-ul:
Meanwhile just removed the changes for 1.21 cgroups job.

vasiliy-ul:
> Can go in once kubevirt/kubevirtci#638 is propagated to kubevirt/kubevirt

Looks like the needed changes have landed in kubevirt. I just wanted to confirm that we are on the same page with this PR. I enabled only one kubevirtci presubmit job (for 1.20) just to be on the safe side and ensure it does not get broken accidentally.
@oshoval , @rmohr , are we okay with that for now? What do you think?

> Please ask @fgimenez help with checking the periodic

Hi @fgimenez . Could you help with the periodic job? Is there a way to check the changes before merging? Unfortunately the job got broken and this PR is supposed to fix it.

fgimenez:
> Hi @fgimenez . Could you help with the periodic job? Is there a way to check the changes before merging? Unfortunately the job got broken and this PR is supposed to fix it.

hi sure, it is running here https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/logs/periodic-kubevirt-e2e-k8s-1.20-cgroupsv2/1414490544733687808, pls check that the job definition is correct https://prow.ci.kubevirt.io/prowjob?prowjob=cccc34b1-e2e4-11eb-bd3f-f875a47e7fb8

vasiliy-ul:
Seems correct. Lets see how it goes. Thanks a lot :+1:

vasiliy-ul:
The job passed successfully: https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/logs/periodic-kubevirt-e2e-k8s-1.20-cgroupsv2/1414490544733687808. I also checked the logs and cgroups v2 seems configured as needed:

```
07:47:54: Transient hostname: node01
07:47:54: + '[' -f /sys/fs/cgroup/cgroup.controllers ']'
07:47:54: Configuring cgroup v2
...
07:48:29: Transient hostname: node02
07:48:29: Configuring cgroup v2
07:48:29: + '[' -f /sys/fs/cgroup/cgroup.controllers ']'
07:48:29: + echo 'Configuring cgroup v2'
```

oshoval:
> Looks like the needed changes have landed in kubevirt. I just wanted to confirm that we are on the same page with this PR. I enabled only one kubevirtci presubmit job (for 1.20) just to be on the safe side and ensure it does not get broken accidentally.
> @oshoval , @rmohr , are we okay with that for now? What do you think?

If so we need to remember to enable 1.21-cgroupsv2 once 1.20 will become outdated

vasiliy-ul:
> If so we need to remember to enable 1.21-cgroupsv2 once 1.20 will become outdated

Right. For now there are no kubevirt periodic/presubmit jobs for 1.21 (only 1.20). I think when those are added then it can be enabled.

oshoval (approved):
Thanks

PR #1418, opened 2021-07-06
Update Prometheus settings

fgimenez:
We are ingesting more metrics now, Prometheus pod was being OOMKilled with the old settings. Also, now that Thanos is in place we don't need that much local retention (we also had issues with the PV being filled).

/cc @dhiller 


dhiller (approved):
/approve

PR #1422, opened 2021-07-07
Add experimental lane to prevent builds on docs changes

fgimenez:
According to https://github.com/kubernetes/test-infra/blob/master/prow/jobs.md#requiring-jobs-for-auto-merge-through-tide looks like this combination of options should work for having required jobs that are not triggered on doc changes:

* always_run: false
* optional: false
* skip_if_only_changed: <docs_regexp>

This way the job should keep being required when run, ie. when the regexp doesn't match at least for one of the changed files. The job in this PR is an experimental lane to check if it works as expected.

/cc @dhiller @rmohr 

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
@@ -1402,3 +1402,25 @@ presubmits:
             memory: 4Gi
         securityContext:
           privileged: true
+  - name: pull-kubevirt-test-doc-change-ignore
+    skip_branches:
+    - release-\d+\.\d+
+    always_run: false
+    optional: false
+    skip_if_only_changed: "^docs/|.md$|^LICENSE$"
+    skip_report: false
+    cluster: prow-workloads
+    decorate: true
+    decoration_config:
+      timeout: 1h
+      grace_period: 5m
+    max_concurrency: 11
+    spec:
+      containers:
+      - image: quay.io/kubevirtci/bootstrap:v20210311-09ebaa2
+        command:
+        - "echo"
+        - "test-doc-change-ignore lane executed"
+        resources:
+          requests:
+            memory: "10Mi"


fgimenez:
/close 

won't work according to https://github.com/kubernetes/test-infra/blob/master/prow/jobs.md#protecting-status-contexts

dhiller:
> /close
> 
> won't work according to https://github.com/kubernetes/test-infra/blob/master/prow/jobs.md#protecting-status-contexts

So sad to see that `skip_if_only_changed` won't work for us :disappointed:
But that makes sense also. So IIUC if the `skip_if_only_changed` condition was not satisfied, the `required` flag didn't appear on the GitHub gui right?

fgimenez:
> So sad to see that skip_if_only_changed won't work for us disappointed
+1 /o\

> So IIUC if the `skip_if_only_changed` condition was not satisfied, the `required` flag didn't appear on the GitHub ui for that job right?

Yep, but the job would not even show up, If it was not satisfied for all the files changed in a PR the lane would not be executed. This doesn't play well with `optional: false` lanes bc branch-protector would configure the repo to require those. The PRs could not be merged bc the status of the required lane would never be "pass". Does that make sense?

dhiller:
> > So sad to see that skip_if_only_changed won't work for us disappointed
> 
> +1 /o\
> 
> > So IIUC if the `skip_if_only_changed` condition was not satisfied, the `required` flag didn't appear on the GitHub ui for that job right?
> 
> Yep, but the job would not even show up, If it was not satisfied for all the files changed in a PR the lane would not be executed. This doesn't play well with `optional: false` lanes bc branch-protector would configure the repo to require those. The PRs could not be merged bc the status of the required lane would never be "pass". Does that make sense?

This is very bad, IIUC the `optional: false` job still _is_ required and tide doesn't see a status on it, so it won't get merged, and users even don't understand why :disappointed: 

Maybe it showed up somewhere in the tide dashboard for the PR but not on the PR itself. Do you have a link BTW to the PR where you checked it?

fgimenez:
> Maybe it showed up somewhere in the tide dashboard for the PR but not on the PR itself. Do you have a link BTW to the PR where you checked it?

Sorry, I didn't even try it after reading https://github.com/kubernetes/test-infra/blob/master/prow/jobs.md#protecting-status-contexts, I thought it would block all PRs from merging. Do you think we should give it a try, maybe not on kubevirt/kubevirt?

dhiller:
> > Maybe it showed up somewhere in the tide dashboard for the PR but not on the PR itself. Do you have a link BTW to the PR where you checked it?
> 
> Sorry, I didn't even try it after reading https://github.com/kubernetes/test-infra/blob/master/prow/jobs.md#protecting-status-contexts, I thought it would block all PRs from merging. Do you think we should give it a try, maybe not on kubevirt/kubevirt?

Would really like to try that out, especially as we have the problem of doc changes in kubevirt/kubevirt triggering e2d builds. But don't really have an idea where else to test this.

PR #1426, opened 2021-07-07
Scrape directly Prow exporter metrics

fgimenez:
Prow exporter metrics are not available in pushgateway. These changes will also make it easier to debug.

/cc @rmohr @dhiller  


github/ci/services/prometheus-stack/manifests/service-monitors/production-control-plane/prow-exporter.yaml
```diff
@@ -0,0 +1,13 @@
...
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  name: prow-exporter
+  labels:
+    group: kubevirtci
+spec:
+  selector:
+    matchLabels:
+      app: prow-exporter
```
rmohr, line 11: Would it make sense to simply create on `ServiceMonitor` and just add labels to the appropriate services?

fgimenez: Sorry, I don't follow, we are creating here a `ServiceMonitor` and a service with the matching labels, is that what you mentioned?

rmohr: I just meant that this seems to apply the pattern of having one service monitor per service. I just meant that you may want to create a `prow` ServiceMonitor which matches a more generic label which is not on the app base.

fgimenez: Aha, sure makes sense to have a common ServiceMonitor for all of our services, I've pushed changes for this, the only requirements for the services to be handled by it are the `monitoring: "true"` label and having a port named `metrics` PTAL.

PR #1431, opened 2021-07-09
Add local PV for greenhouse in workloads cluster

fgimenez:
Creates a local PV and correspondent PVC for greenhouse on the workloads cluster, where it needs to run on the cache node. This way we can control the size of the volume used by greenhouse without affecting other apps running on the same host.

/cc @dhiller @rmohr


enp0s3 (approved):
/lgtm

dhiller (approved):
/approve

PR #1433, opened 2021-07-12
Do not run optional lanes on doc changes

fgimenez:
According to https://github.com/kubernetes/test-infra/blob/master/prow/jobs.md#protecting-status-contexts conditionally-run jobs may or may not post a status context to GitHub, and as such, they cannot be required through the branch protection mechanism. This prevents us from not executing required jobs depending on some condition, but we can still stop running the non-required ones. This PR changes the job definition of kubevirt/kubevirt presubmits to not trigger them on docs changes.

/cc @dhiller 


dhiller (commented):
@fgimenez sorry for holding up this one. I need to prepare my changes on the require-e2e automation so that we don't get into this exact situation.

/hold

dhiller (approved):
/approve

dhiller:
Please see https://github.com/kubevirt/project-infra/pull/1438

dhiller:
/unhold

As the job will only run at the start of next month there'll be enough time to merge it :)

PR #1437, opened 2021-07-13
Expose job's always_run and brancher settings

fgimenez:
This will allow us to estimate the resource utilization per full execution on each PR (taking into account only presubmit jobs that have `always_run: true` and discarding release presubmits through the brancher.branches and brancher.skip_branches settings).

/cc @dhiller 


dhiller (approved):
/approve

PR #1440, opened 2021-07-15
Revert "Do not run optional lanes on doc changes"

fgimenez:
Reverts kubevirt/project-infra#1433

Sinker and other pods are crashlooping with errors like `job pull-kubevirt-e2e-k8s-1.21-sig-network is set to always run but also declares skip_if_only_changed targets, which are mutually exclusive`

/cc @dhiller

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
@@ -17,10 +17,7 @@ presubmits:
       preset-shared-images: "true"
     max_concurrency: 11
     name: pull-kubevirt-e2e-k8s-1.21-sig-network
-    # WARNING: please remember to remove skip_if_only_changed if this job is
-    # set to required (optional: false)
     optional: true
-    skip_if_only_changed: "^docs/|.md$|^LICENSE$"
     skip_branches:
     - release-\d+\.\d+
     spec:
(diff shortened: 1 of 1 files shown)


dhiller:
I think that this is because the default for `always_run` is `true`, can that be? (Didn't find hard evidence for that though, although the docs implicitly suggest that)

dhiller (approved):
/approve

fgimenez:
> I think that this is because the default for `always_run` is `true`, can that be? (Didn't find hard evidence for that though, although the docs implicitly suggest that)

exactly, according to the error message `always_run: true` and `skip_if_only_changed` are mutually exclusive. Not only for required (`optional: false`) jobs. Reading the docs I mistakenly thought that when `optional: true` both settings could co-exist, but it is not the case.

fgimenez:
i wonder why `check-prow-config` succeeded, the pods affected reported a failure building the config agent.

dhiller:
> > I think that this is because the default for `always_run` is `true`, can that be? (Didn't find hard evidence for that though, although the docs implicitly suggest that)
> 
> exactly, according to the error message `always_run: true` and `skip_if_only_changed` are mutually exclusive. Not only for required (`optional: false`) jobs. Reading the docs I mistakenly thought that when `optional: true` both settings could co-exist, but it is not the case.

Yeah, I think this is a bug in configcheck somewhere.

dhiller:
> > > I think that this is because the default for `always_run` is `true`, can that be? (Didn't find hard evidence for that though, although the docs implicitly suggest that)
> > 
> > 
> > exactly, according to the error message `always_run: true` and `skip_if_only_changed` are mutually exclusive. Not only for required (`optional: false`) jobs. Reading the docs I mistakenly thought that when `optional: true` both settings could co-exist, but it is not the case.
> 
> Yeah, I think this is a bug in configcheck somewhere.

I looked at the code here, seems like this is not validated in any way: https://github.com/kubernetes/test-infra/blob/6bba2f0544a9ef2d58bc20ef7ea1e5ed1a93e626/prow/cmd/checkconfig/main.go#L519

PR #1445, opened 2021-07-15
Update Prometheus settings

fgimenez:
We still had issues with the Prometheus PV being filled, these changes make the retention even lower (no need to retain a lot of data given that now Thanos is in place) and increase a bit the PV size.

/cc @dhiller 


dhiller (approved):
/approve

PR #1457, opened 2021-07-19
Increase Prometheus CPU resources in control plane

fgimenez:
Prometheus pod is still being OOMKilled in ibm-prow-jobs cluster, this leads to these gaps in metrics:

Screenshot from 2021-07-19 15-26-27

These changes give it resources according to the recorded resource utilization:

Screenshot from 2021-07-19 15-35-17 

/cc @dhiller 


fgimenez:
After increasing the CPU resources for testing there are no more memory spikes:

Screenshot from 2021-07-20 09-25-12

Let's only increase CPU requests and limits for now, changes pushed.

dhiller (approved):
/approve

PR #1459, opened 2021-07-20
Define greenhouse CPU resources

fgimenez:
Sets the limits above the highest recent recorded utilization:

Screenshot from 2021-07-20 11-24-36
 
/cc @rmohr @dhiller 

### DIFF
--- a/github/ci/prow-deploy/kustom/components/greenhouse/base/resources/deployment.yaml
+++ b/github/ci/prow-deploy/kustom/components/greenhouse/base/resources/deployment.yaml
@@ -52,8 +52,10 @@ spec:
           mountPath: /data
         resources:
           requests:
+            cpu: 1.2
             memory: 3Gi
           limits:
+            cpu: 1.2
             memory: 3Gi
       volumes:
       - name: cache


github/ci/prow-deploy/kustom/components/greenhouse/base/resources/deployment.yaml
```diff
@@ -52,8 +52,10 @@ spec:
           mountPath: /data
         resources:
           requests:
+            cpu: 1.2
             memory: 3Gi
           limits:
+            cpu: 1.2
```
rmohr, line 58: If we limit it to 1.2 it can not use more. Do you just want to reserve enough resources, or do you also want to limit it to 1.2? It may be better to just put a request and let it use more CPU if it wants to.

fgimenez: Yes, the goal of setting the limit field is having a Guaranteed QoS. If we don't set limit and other non-constraint pods start eating CPU in the same host this one can get into trouble with a Burstable QoS.

1.2 gives some space (20%) above the maximum we have registered. Also, being CPU, if it tries to use more it won't be evicted and will be just throttled, in that case we can increase the resource again, wdyt?

rmohr: Hm, you can also consider assigning a priority class. In kubevirt/kubevirt we try to avoid giving components limits. Instead we use priority classes. But just optional thoughts.

rmohr (approved):
/lgtm
/approve

github/ci/prow-deploy/kustom/components/greenhouse/base/resources/deployment.yaml
fgimenez: Sure, will take a look thanks! I think in this case Guaranteed QoS can help more, it not only prevents preemption but makes sure that the pod is going to be scheduled on a node with enough resources according to its definition. AIUI a priority class won't help when the required resources are not specified and non-constrained pods start fighting for the available resources.

rmohr: The main issue with Guaranteed is that one either gets killed or throttled on the limits on spikes, even when the rest of the node is idle. You still have the amount of cpu-time and memory reserved for you, also without limits, but you can use more if available. Also the scheduler will still only schedule the workload to nodes where it fits due to the requests (not limits).

fgimenez: Yep, I understand, but in the end not specifying limits leads to the situation we had today (and the situation we had before with deck and hook), the pods without Guaranteed QoS or that request more than the CPU limits gets their CPU throttled and the performance decays. 

Setting limits to the same value as requests fixed deck and hook, the pods no longer need to compete for the resources defined. You just need to keep an eye on them to make sure that the resources are properly defined according to the usage, or leave a VPA do it for you :)

fgimenez: Probably a combination of only setting requests plus a priority class could have the same effect, plus the benefit of allowing to use more resources when they are available, of course.

PR #1462, opened 2021-07-20
test SRIOV lane with 'kindnet' CNI plugin instead of Calico

ormergi:
Test kubevirt SRIOV tests when deploying KinD
default CNI plugin, 'kindnet', instead of Calico.


maiqueb:
Would you elaborate the why ?

ormergi:
@maiqueb Basically it will reduce maintenance and simplify the provider
Check out https://github.com/kubevirt/kubevirtci/pull/644 description for more detailes

PR #1466, opened 2021-07-21
Move uploader to robots

fgimenez:
Also takes advantage of the single robots test presubmit and removes the plugins test jobs.

/cc @rmohr @dhiller  


dhiller (approved):
/approve

PR #1467, opened 2021-07-22
Add RedHatQE/openshift-python-wrapper to project

myakove:
Add RedHatQE/openshift-python-wrapper to project to use prow plugins.

### DIFF
--- a/github/ci/prow-deploy/kustom/base/configs/current/plugins/plugins.yaml
+++ b/github/ci/prow-deploy/kustom/base/configs/current/plugins/plugins.yaml
@@ -328,23 +328,23 @@ plugins:
   - override
 
   RedHatQE/openshift-python-wrapper:
-    - size
-    - label
-    - hold
-    - lgtm
-    - approve
-    - release-note
-    - owners-label
-    - assign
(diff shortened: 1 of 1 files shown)


github/ci/prow-deploy/kustom/base/configs/current/plugins/plugins.yaml
```diff
@@ -327,6 +327,25 @@ plugins:
   - trigger
   - override
 
+  RedHatQE/openshift-python-wrapper:
```
fgimenez, line 330: Could you please move the array elements two spaces to the left, so that they are aligned with the repo name? Just a matter of uniformity, same as we have with the rest repos.

fgimenez (commented):
Just a small nit, other than that looks good

github/ci/prow-deploy/kustom/base/configs/current/plugins/plugins.yaml
myakove: ACK

fgimenez (approved):
/approve

Thanks!

PR #1474, opened 2021-07-26
Bump cherrypicker image

fgimenez:
This will prevent issues with the new HMAC secret format.

/cc @enp0s3 


enp0s3 (approved):
/approve

PR #1476, opened 2021-07-28
Add daily job to run performance density test

marceloamaral:
Add a Prow job that runs 3 times a day a density test creating 100 VMs.

This prow job results will be used as the baseline performance of KubeVirt to evaluate possible performance regression.

Additionally, this job runs in the `prow-workloads` but will be modified in the future to run a the `prow-performance` cluster.

This PR Depends on kubevirt/kubevirt/pull/6139

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml
@@ -1293,3 +1293,64 @@ periodics:
       - name: pgp-bot-key
         secret:
           secretName: pgp-bot-key
+- name: periodic-kubevirt-e2e-k8s-1.21-sig-performance
+  annotations:
+    testgrid-dashboards: kubevirt-periodics
+    testgrid-days-of-results: "60"
+  cron: "40 7,15,23 * * *"
+  decorate: true
+  cluster: prow-workloads
+  decoration_config:
+    timeout: 4h
+    grace_period: 5m
+  labels:
+    preset-dind-enabled: "true"
+    preset-docker-mirror-proxy: "true"
+    preset-bazel-cache: "true"
+    preset-bazel-unnested: "true"
+    ci.kubevirt.io/prometheus: ""
+  extra_refs:
+  - org: kubevirt
+    repo: kubevirt
+    base_ref: main
+    work_dir: true
+  reporter_config:
+    slack:
+      job_states_to_report: []
+  spec:
+    nodeSelector:
+      type: bare-metal-external
+    containers:
+      - image: quay.io/kubevirtci/bootstrap:v20210311-09ebaa2
+        env:
+          - name: KUBEVIRT_PROVIDER
+            value: "k8s-1.21"
+          - name: KUBEVIRT_STORAGE
+            value: "hpp"
+          - name: KUBEVIRT_MEMORY_SIZE
+            value: "9G"
+          - name: KUBEVIRT_NUM_NODES
+            value: "4"
+          - name: KUBEVIRT_DEPLOY_PROMETHEUS
+            value: "true"
+          - name: KUBEVIRT_PROVIDER_EXTRA_ARGS
+            value: "--prometheus-port 30007"
+        command:
+          - "/usr/local/bin/runner.sh"
+          - "/bin/sh"
+          - "-c"
+          - |
+            # Create k8s cluster
+            make cluster-up
+
+            # Push and deploy kubevirt
+            make cluster-sync
+
+            FUNC_TEST_ARGS="--noColor --seed=42" make perftest            
+        # docker-in-docker needs privileged mode
+        securityContext:
+          privileged: true
+        resources:
+          requests:
+            memory: "44Gi"
+            cpu: "12000m"


marceloamaral:
cc @rmohr
cc @fgimenez

github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml
```diff
@@ -1293,3 +1293,65 @@ periodics:
       - name: pgp-bot-key
         secret:
           secretName: pgp-bot-key
+- name: periodic-kubevirt-e2e-k8s-1.21-sig-performance
```
rmohr, line 1296: If you want you could also add an optional presubmit PR, to e.g. test perf changes already now on PRs, without thinking about capacity.

github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml
```diff
@@ -1293,3 +1293,65 @@ periodics:
...
+          - name: KUBEVIRT_STORAGE
+            value: "hpp"
+          - name: KUBEVIRT_MEMORY_SIZE
+            value: "9G"
+          - name: KUBEVIRT_NUM_NODES
+            value: "4"
+          - name: KUBEVIRT_DEPLOY_PROMETHEUS
+            value: "true"
+          - name: KUBEVIRT_DEPLOY_GRAFANA
+            value: "true"
```
rmohr, line 1335: Would it make sense to not start grafana? Do you use it somewhere?

github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml
```diff
@@ -1293,3 +1293,65 @@ periodics:
...
+        command:
+          - "/usr/local/bin/runner.sh"
+          - "/bin/sh"
+          - "-c"
+          - |
+            # Make sure we cleanup first
+            make cluster-down
```
rmohr, line 1342: You probably took this from `automation/test.sh`. The initial cleanup is a very old artifact from our old CI infra which was still jenkins based and did not have container isolation. You can safely drop this. There will never be the need to clean anything up.

rmohr (changes requested):
Looks excellent. Only two small nits.

github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml
marceloamaral: Make sense, I will create it later!

github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml
marceloamaral: Well observed, we don't need Grafana.

github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml
```diff
@@ -1293,3 +1293,60 @@ periodics:
...
+            # Push and deploy kubevirt
+            make cluster-sync
+
+            FUNC_TEST_ARGS="--noColor --seed=42" make perftest            
+        # docker-in-docker needs privileged mode
+        securityContext:
+          privileged: true
+        resources:
+          requests:
+            memory: "36Gi"
```
rmohr, line 1351: Consider adding a  few more GB to the request to avoid OOM kills, since the VMs already use 4*9G. Maybe add 8 more?

github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml
```diff
@@ -1293,3 +1293,60 @@ periodics:
...
+    testgrid-dashboards: kubevirt-periodics
+    testgrid-days-of-results: "60"
+  cron: "40 7,15,23 * * *"
+  decorate: true
+  cluster: prow-workloads
+  decoration_config:
+    timeout: 4h
+    grace_period: 5m
+  labels:
+    preset-dind-enabled: "true"
```
rmohr, line 1307: I think here we would need the label which the service monitor uses to collect the metrics. The labels here get directly propagated to the pods.

github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml
```diff
@@ -1293,3 +1293,60 @@ periodics:
...
+    preset-docker-mirror-proxy: "true"
+    preset-shared-images: "true"
```
fgimenez, line 1309: This preset only makes sense for windows or rhel tests. It would be interesting to add the presets related to bazel cache instead to speed up the build:
```suggestion
    preset-bazel-cache: "true"
    preset-bazel-unnested: "true"
```

github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml
marceloamaral: Right, 8G more sounds reasonable!

fgimenez (commented):
Looks good! Small nit about presets

github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml
marceloamaral: I will add `group: kubevirtci`

github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml
marceloamaral: Makes sense, thanks!

github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml
marceloamaral: I will also include 1 more CPU. 4*6+1

github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml
marceloamaral: If we add this test for each PR, we probably do not need to run daily.

But, let's start with this test and update it later.
The goal is to run daily a medium size test, e.g., creating 500 to 1000VMs.

github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml
rmohr: According to https://github.com/kubevirt/project-infra/blob/15807c5f709d63fd002cb21e1ed31f0f4ae76452/github/ci/services/prometheus-stack/manifests/stack/production-e2e-workloads/prometheus-federation.yaml#L17 I think you have to add 

```
ci.kubevirt.io/prometheus: ""
```

github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml
```diff
@@ -1293,3 +1293,62 @@ periodics:
...
+    work_dir: true
+  reporter_config:
+    slack:
+      job_states_to_report: []
+  spec:
+    nodeSelector:
+      type: bare-metal-external
+    containers:
+      - image: quay.io/kubevirtci/bootstrap:v20210311-09ebaa2
+        env:
```
rmohr, line 1325: Can it be that we have to add

```yaml
- name: KUBEVIRT_PROVIDER_EXTRA_ARGS
  value: "--prometheus-port 30007"
```

to give prometheus the fixed reachable port which we expect at https://github.com/kubevirt/project-infra/blob/15807c5f709d63fd002cb21e1ed31f0f4ae76452/github/ci/services/prometheus-stack/manifests/stack/production-e2e-workloads/prometheus-federation.yaml#L12 ?
```

File: github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml (hunk 4 above)
Comment (rmohr) [reply github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml]: Hm, we normally don't add cpu requests at the moment. I would at least suggest to request only a part of the CPUs.

File: github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml (hunk 7 above)
Comment (marceloamaral) [reply github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml:1325]: Thanks, it should be `ci.kubevirt.io/prometheus: ""`
Right, we need to expose Prometheus in the "localhost" with the 30007 port

File: github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml (hunk 4 above)
Comment (marceloamaral) [reply github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml]: hum, can it be 4*6?
I am wondering it could become a flaky test when there are not enough CPU, so that VMIs will not be scheduled.

Comment (rmohr) [reply github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml]: All other jobs don't reqeust any CPU resources at all (using the default of 100m implicitly). We control the job density right now through the memory requests (since there are also other aspects like disk io which can't be managed on that level). This can probably still be improved, but I think it is best if you just leave that out right now. You can still request the normal amount of vcpus in for the qemu nodes (just leaving the default like you do right now).

File: github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml [hunk 8]
```diff
@@ -1293,3 +1293,64 @@ periodics:
...
+            make cluster-sync
+
+            FUNC_TEST_ARGS="--noColor --seed=42" make perftest            
+        # docker-in-docker needs privileged mode
+        securityContext:
+          privileged: true
+        resources:
+          requests:
+            memory: "44Gi"
+            cpu: "24000m"
```
rmohr, line 1356: @marceloamaral can you reduce this more? Like maybe to 12 cpus? Just to be on the safe side and avoid evictions of other e2e jobs.

github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml
```diff
@@ -1293,3 +1293,64 @@ periodics:
       - name: pgp-bot-key
         secret:
           secretName: pgp-bot-key
+- name: periodic-kubevirt-e2e-k8s-1.21-sig-performance
+  annotations:
+    testgrid-dashboards: kubevirt-periodics
+    testgrid-days-of-results: "60"
+  cron: "40 7,15,23 * * *"
+  decorate: true
+  cluster: prow-workloads
```
fgimenez, line 1302: What is the status of the perf cluster you have been working on? If it is ready we could include it in Prow and start running these jobs there, wdyt?

github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml
marceloamaral: Done!

github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml
marceloamaral: I am investigating an [issue](https://docs.google.com/document/d/17SHOUTWF3tpOA3FhSOpz3rXia-5JgYvsX9LLi562-_0/edit#heading=h.8zwzln5oc6n) about the max number of VMIs per node in the cluster. After that, we can fully dedicate the cluster to the CI.
Then, we can update this test with CPU requirements in the new cluster.
Does it sound resonable?

fgimenez (commented):
/lgtm

marceloamaral:
@rmohr if you agree that the test is ok now, could you please approve it?

fgimenez:
/approve

I think all the suggestions have been addressed, if not we can work on them on a follow-up, thank you @marceloamaral!

rmohr:
> @rmohr if you agree that the test is ok now, could you please approve it?

Was unexpectedly out for a few days. :+1:

PR #1477, opened 2021-07-28
Do not create HMAC secret

fgimenez:
Now that we are using the [hmac tool](https://github.com/kubernetes/test-infra/blob/master/prow/cmd/hmac/README.md) to handle the HMAC setup we no longer need to create the k8s secret ourselves with a plain random string, the secret is created by the tool and the contents are updated every time we reexecute it. 

The changes includes code to initialize the hmac secret on the integration testing env, where we don't call the hmac tool.

/cc @enp0s3 @dhiller 

### DIFF
--- a/github/ci/prow-deploy/kustom/overlays/ibmcloud-production/kustomization.yaml
+++ b/github/ci/prow-deploy/kustom/overlays/ibmcloud-production/kustomization.yaml
@@ -298,10 +298,6 @@ secretGenerator:
     files:
       - secret=secrets/cookie
     type: Opaque
-  - name: hmac-token
-    files:
-      - hmac=secrets/hmac-token
-    type: Opaque
   - name: github-oauth-config
     files:
       - secret=secrets/github-oauth-config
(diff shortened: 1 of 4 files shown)


rmohr:
/hold 

hm, something seems to be wrong:

```
fatal: [instance]: FAILED! => {"msg": "The field 'environment' has an invalid value, which includes an undefined variable. The error was: 'shell_environment' is undefined\n\nThe error appears to be in '/home/prow/go/src/github.com/kubevirt/project-infra/github/ci/prow-deploy/tasks/cleanup.yml': line 3, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n# So you run it twice and the second time it will fail, and there's no easy way to understand if it's legit\n- name: Delete all remote resources.\n  ^ here\n"}
```

fgimenez:
> hm, something seems to be wrong:

indeed, we need the hmac secret on the test env, where we don't run the hmac tool, on it

fgimenez:
> > hm, something seems to be wrong:
> 
> indeed, we need the hmac secret on the test env, where we don't run the hmac tool, on it

Done, now the secret is initialized on the testing env, PTAL.

PR #1489, opened 2021-08-02
openshift-python-wrapper remove release-note plugin

myakove:
openshift-python-wrapper remove release-note plugin


fgimenez (approved):
/approve

PR #1468, opened 2021-07-22
Add RedHatQE/openshift-python-wrapper to managed webhooks

fgimenez:
This will configure gh webhook on RedHatQE/openshift-python-wrapper repo, kubevirt-bot is already admin there.

/cc @rmohr 

Signed-off-by: Federico Gimenez <fgimenez@redhat.com>

PR #1490, opened 2021-08-03
Remove openshift-python-wrapper from prow

myakove:


fgimenez (commented):
Looks good, can you please also remove RedHatQE/openshift-python-wrapper from the managed webhooks? The change was added here https://github.com/kubevirt/project-infra/pull/1468/files

PR #1491, opened 2021-08-04
Only run check-prow-config when Prow config changes

fgimenez:
/cc @rmohr @dhiller 


fgimenez:
/close

It would make it non-required

PR #1500, opened 2021-08-09
Configuration fixes of KubeVirt Velero Plugin's CI.

tomob:
This PR fixes configuration of KubeVirt Velero Plugin's CI.

There are two fixes:
- pull-kvp-unit-test needs to be run in priviledged mode, because it
  runs the tests inside Docker container

- pull-kvp-functional-test has cluster setup pulled from the test target
  to the job configuration (i.e. test config instead of chainging
  dependencies inside Makefile target definition).

Signed-off-by: Tomasz Baranski <tbaransk@redhat.com>

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt-velero-plugin/kubevirt-velero-plugin-presubmits.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt-velero-plugin/kubevirt-velero-plugin-presubmits.yaml
@@ -26,7 +26,7 @@ presubmits:
             - "make test"
           # docker-in-docker needs privileged mode
           securityContext:
-            privileged: false
+            privileged: true
           resources:
             requests:
               memory: "4Gi"
@@ -53,7 +53,7 @@ presubmits:
             - "/usr/local/bin/runner.sh"
             - "/bin/sh"
             - "-c"
-            - "make test-functional"
+            - "make build-builder cluster-up cluster-push-image cluster-sync test-functional CLUSTER_PREFIX='-p pull-kvp-functional-test'"
           # docker-in-docker needs privileged mode
           securityContext:
             privileged: true


github/ci/prow-deploy/files/jobs/kubevirt/kubevirt-velero-plugin/kubevirt-velero-plugin-presubmits.yaml
```diff
@@ -53,7 +53,7 @@ presubmits:
             - "/usr/local/bin/runner.sh"
             - "/bin/sh"
             - "-c"
-            - "make test-functional"
+            - "make build-builder cluster-up cluster-push-image cluster-sync test-functional CLUSTER_PREFIX='-p pull-kvp-functional-test'"
```
awels, line 56: So the prevailing consensus is that it makes sense to create a script file, and call that script file from here. This way you can independently modify the script and retest changes without having to modify the project-infra code. So I suggest you make a script file that basically does what this line does and call that script from here.

awels:
@fgimenez Can you review as well?

github/ci/prow-deploy/files/jobs/kubevirt/kubevirt-velero-plugin/kubevirt-velero-plugin-presubmits.yaml
tomob: I can also have those as prerequisites for `test-functional` target but it makes local testing annoying. And because I do need the change to unit tests anyway (`priviledged: true`), I thought I change this as well and have better developer experience.

fgimenez: Yep, agree with @awels, you can just call the script from here and have whatever logic you want inside, this way the CI logic is kept in the same repo as the source code. In this case you can start with the whole `make` call, but later on you can update the CI logic from the target repo itself.

tomob:
I've broken my fork of the repo. I need to recreate the PR in a new fork. I'm closing this one and I'll open a new PR with the change you requested.

PR #1504, opened 2021-08-16
Configuration fixes of KubeVirt Velero Plugin's CI.

tomob:
This PR fixes configuration of KubeVirt Velero Plugin's CI.

There are two fixes:
 - pull-kvp-unit-test needs to be run in priviledged mode, because it
   runs the tests inside Docker container

 - pull-kvp-functional-test calls a script to setup a cluster and run
   the tests rather than call Make target directly.

Signed-off-by: Tomasz Baranski tbaransk@redhat.com


tomob:
@awels @fgimenez 
This is the PR which replaces #1500

tomob:
@fgimenez Can you take a look?

PR #1506, opened 2021-08-19
HCO: fix kubevirt providers

nunnatsa:
KUBEVIRT_PROVIDER env var was not set in HCO. Added


nunnatsa:
Not needed. closing

PR #1520, opened 2021-08-25
Make the kubevirt sig-storage 1.19 and 1.20 lanes optional

maya-r:
These have flakiness, and we have the more stable 1.21 lane
required already, which should avoid introducing flakes.


maya-r:
@dhiller - will this run into issues with kubevirt-presubmit-requirer, or does it only touch the latest version (1.21)?

dhiller:
> @dhiller - will this run into issues with kubevirt-presubmit-requirer, or does it only touch the latest version (1.21)?

this only checks the latest one

EdDev (approved):
Thanks. I guess this is temporary (at least for 1.20), so can we make sure we track this in an issue (so we will not leave it like this?).

fgimenez (approved):
/approve

Makes total sense, thanks! We need to keep an eye on the presubmit and periodic executions to check when they can be required again.

PR #1525, opened 2021-08-30
Add nonroot keyword to target

xpivarc:
This will allow us to run specific tests
to verify we are testing nonroot


dhiller (approved):
/approve

PR #1535, opened 2021-09-01
Set periodic-kubevirtci-bump-kubevirt to run every 12h

fgimenez:
Sets the job schedule with cron instead of interval, which depends on the time when the job was executed for the last time.

/cc @oshoval @dhiller 

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/kubevirtci/kubevirtci-periodics.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/kubevirtci/kubevirtci-periodics.yaml
@@ -70,7 +70,7 @@ periodics:
         secret:
           secretName: gcs
 - name: periodic-kubevirtci-bump-kubevirt
-  interval: 24h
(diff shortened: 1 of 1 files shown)


oshoval (approved):
Thanks, just note that it will rebase the PRs every 12h 
so sometimes it will stall them from getting merge

we should fix the job to not push changes that are pure rebase without actual bump
i might do it next week or so
meanwhile we can either keep it cron 24 or 12 and take the chance, up to you

dhiller:
> Thanks, just note that it will rebase the PRs every 12h
> so sometimes it will stall them from getting merge
> 
> we should fix the job to not push changes that are pure rebase without actual bump
> i might do it next week or so
> meanwhile we can either keep it cron 24 or 12 and take the chance, up to you

@oshoval I've created a PR to tackle this here: https://github.com/kubevirt/project-infra/pull/1528

dhiller (approved):

dhiller (approved):
/approve

PR #1538, opened 2021-09-01
git-pr, Fix case where the summary is empty but rebase exists

oshoval:
In case the summary command returns empty output,
for example in case there isn't bump detected,
but the rebase do exists, the pr-creator will try
to push a rebase, and fail because the PR title field
must not be empty.

Add a condition that if the commit summary is empty,
there will be no PR created, because a pure rebase is not needed.


oshoval:
I am not sure its possible
because it always start with the latest, so iif no files were added due to the command itself, it exits anyway

PR #1546, opened 2021-09-04
Prepull bazel 4.2.1 in bootstrap image

fgimenez:
/cc @dhiller @rmohr

Signed-off-by: Federico Gimenez <federico.gimenez@gmail.com>

### DIFF
--- a/images/bootstrap/Dockerfile
+++ b/images/bootstrap/Dockerfile
@@ -98,8 +98,7 @@ RUN  curl -Lo ./bazelisk https://github.com/bazelbuild/bazelisk/releases/downloa
 
 # Cache the most commonly used bazel versions inside the image
 # and remove resulting cache directories to save a few hundred MB
-RUN USE_BAZEL_VERSION=3.4.1 bazel version && \
-    USE_BAZEL_VERSION=4.0.0 bazel version && \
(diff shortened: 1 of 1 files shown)


images/bootstrap/Dockerfile
```diff
@@ -99,8 +99,9 @@ RUN  curl -Lo ./bazelisk https://github.com/bazelbuild/bazelisk/releases/downloa
 # Cache the most commonly used bazel versions inside the image
 # and remove resulting cache directories to save a few hundred MB
 RUN USE_BAZEL_VERSION=3.4.1 bazel version && \
-  USE_BAZEL_VERSION=4.0.0 bazel version && \
-  rm -rf /root/.cache/bazel
+    USE_BAZEL_VERSION=4.0.0 bazel version && \
```
dhiller, line 102: Should we maybe also prepull 4.1.0 which is used in project-infra IIRC?

dhiller:
Great, thanks for this!

images/bootstrap/Dockerfile
fgimenez: Good catch, in order to keep the image size low I've changed the PR to only prepull the versions used in kubevirt and project-infra, PTAL

dhiller (approved):
/approve

Thanks!

PR #1549, opened 2021-09-05
Don't always run sig-storage flaky lanes.

maya-r:
People are likely spending effort on understanding why they see
failures, and we know these lanes are currently flaky.
Running the 1.21 lane will be enough to prevent regressions.


fgimenez (approved):
/approve

Makes sense to me, thanks!

PR #1550, opened 2021-09-06
Re-add cgroupsv2 periodic

fgimenez:
Adds cgroupsv2 periodic now that the fix for the recurring error has been merged here https://github.com/kubevirt/kubevirt/pull/6335

/cc @vasiliy-ul @dhiller 

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml
@@ -1460,3 +1460,50 @@ periodics:
         privileged: true
     nodeSelector:
       type: bare-metal-external
+- name: periodic-kubevirt-e2e-k8s-1.20-cgroupsv2
+  annotations:
+    testgrid-dashboards: kubevirt-periodics
+    testgrid-days-of-results: "60"
+  cron: "0 9,17,0 * * *"
+  decorate: true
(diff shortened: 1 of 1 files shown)


github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml
```diff
@@ -1460,3 +1460,50 @@ periodics:
...
+      type: bare-metal-external
+    containers:
+      - image: quay.io/kubevirtci/bootstrap:v20210311-09ebaa2
+        env:
+          - name: TARGET
+            value: "k8s-1.20"
+          - name: KUBEVIRT_CGROUPV2
+            value: "true"
+          - name: KUBEVIRT_E2E_SKIP
+            value: "Multus|SRIOV|GPU|Macvtap|\\[sig-operator\\]"
```
vasiliy-ul, line 1498: I think makes sense to adjust the skip-list a bit:
```suggestion
            value: Multus|SRIOV|GPU|Macvtap|MediatedDevices|QUARANTINE
```

fgimenez: Good catch! We don't need to add `QUARANTINE` here, this is handled separately. I've removed `\\[sig-operator\\]` and added `\\[sig-operator\\]`, PTAL.

PR #1554, opened 2021-09-07
Revert "Add nonroot keyword to target (#1525)"

rmohr:
Non root lanes are failing with:

```
The cluster provider k8s-1.20-nonroot does not exist
```

See for example https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/6361/pull-kubevirt-e2e-k8s-1.20-sig-compute-nonroot/1435140916326174720


rmohr:
@xpivarc maybe you wanted to to a change in kubevirt/kubevirt so that this change does the right thing and just the orchestration went wrong in #1525

fgimenez (approved):
/approve

Thanks!

PR #1556, opened 2021-09-07
Add nonroot gpu lane

xpivarc:

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
@@ -970,6 +970,70 @@ presubmits:
           path: /dev/vfio/
           type: Directory
         name: vfio
+  - always_run: false
+    annotations:
+      fork-per-release: "true"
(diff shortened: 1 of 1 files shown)


github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
```diff
@@ -970,6 +970,70 @@ presubmits:
...
+      testgrid-dashboards: kubevirt-presubmits
+    cluster: prow-workloads
+    decorate: true
+    decoration_config:
+      grace_period: 30m0s
+      timeout: 4h0m0s
+    labels:
+      preset-dind-enabled: "true"
+      preset-docker-mirror-proxy: "true"
+    max_concurrency: 1
```
dhiller, line 985: I don't know much about the requirements of the kind lanes, so this might make sense. But in general this enforces that only one of these jobs will run on the cluster at the same time.

```
	// MaxConcurrency restricts the total number of instances
	// of this job that can run in parallel at once
```

dhiller (approved):
/approve
/hold

Let's see how the job result looks like.

PR #1568, opened 2021-09-10
Fix bazel cache on ibm-prow-jobs

fgimenez:
When alerts were enabled we started getting an alert about an estimation of bazel's disk being filled within 4 days. Turns out that because of how bazel handles the cache, writing and evicting entries, this is normal behavior and the alert would be flapping forever, triggering when new entries are written and resolving when old entries are evicted.

To prevent that behavior in this PR:
* the configuration of bazel in ibm-prow-jobs is aligned to what we have in prow-workloads, leaving 30% of the volume to have enough room to grow and sink the cache.
* the defined PV is increased to 700Gb.

/cc @dhiller   


dhiller (approved):
/approve

PR #1569, opened 2021-09-10
Fix kubemacpool flakefinder weekly report job

fgimenez:
We started getting alerts about one job in non-ready state, looking at the pod list it was in pending state, from the description:
```
$ kubectl describe pod -n kubevirt-prow-jobs b38e4b9b-124b-11ec-ba99-4ab30ff1269c 
......
Containers:
  test:
......
    Environment:
......
      JOB_NAME:                        periodic-publish-kubemacpool-flakefinder-four-weekly-report
......
Events:
  Type     Reason            Age                   From               Message
  ----     ------            ----                  ----               -------
  Warning  FailedScheduling  2m17s (x73 over 67m)  default-scheduler  0/6 nodes are available: 6 node(s) didn't match Pod's node affinity/selector.
```
Looking at the job definition it doesn't have a cluster defined and it has a node selector that doesn't match any of the nodes in the cluster. This has been probably happening for a while, after the 24h deadline period the job will be probably removed by plank. We might find other periodics with this same problem, alerts will bring these issues to the surface.

/cc @dhiller 

### DIFF
--- a/github/ci/prow-deploy/files/jobs/k8snetworkplumbingwg/kubemacpool/kubemacpool-periodics.yaml
+++ b/github/ci/prow-deploy/files/jobs/k8snetworkplumbingwg/kubemacpool/kubemacpool-periodics.yaml
@@ -80,10 +80,8 @@ periodics:
     annotations:
       testgrid-create-test-group: "false"
     decorate: true
+    cluster: ibm-prow-jobs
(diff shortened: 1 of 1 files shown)


rmohr:
Very nice.

/lgtm
/approve

PR #1570, opened 2021-09-13
Add area/monitoring label to kubevirt/kubevirt

rmohr:
We have some people on the project which are very intested in tracking
monitoring changes for e.g. performance and scale, dashboard creation or
alerting.

This change will allow people to write `/area monitoring` to label issues and PRs.


fgimenez (approved):
/approve

Thanks!

PR #1573, opened 2021-09-14
Add CI KPI alerts

fgimenez:
Basic alerts for ci-health metrics, warning for yellow badge levels and critical for red badge levels as defined in https://github.com/fgimenez/ci-health/blob/main/pkg/constants/main.go#L23-L30

Also renamed the manifests directory to `rules` instead of `alert-rules`, we have both alerts and record rules there.

/cc @dhiller 


dhiller (approved):
/approve

PR #1575, opened 2021-09-14
Do not update existing KCI provider bump PR when hold label exists

enp0s3:
There are some cases when the automatic bump of a new provider needs
some manual adjsutments. In that case we would like to freeze the
push to existing bump PR until we manage to fix the issues.


dhiller (approved):
/approve

Thanks!

PR #1590, opened 2021-09-17
Updated prow-deploy image

fgimenez:
The main goal is shipping a more recent kubectl that allows `apply --server-side` to prevent errors like https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_project-infra/1587/pull-project-infra-prow-deploy-test/1438747463094636544#1:build-log.txt%3A552 

I've also taken the opportunity to switch the base image to the latest fedora-based bootstrap, to bump the test-infra sources from which we build several tools and to add the which package required here https://github.com/kubevirt/project-infra/blob/main/hack/patch_node.sh#L15.

/cc @dhiller 


dhiller (approved):
/approve

PR #1592, opened 2021-09-17
Do not deploy default IngressClass in ibm-prow-jobs

fgimenez:
We no longer need to deploy the default IngressClass in ibm-prow-jobs cluster, it is deployed with different values by the managed services.

This change will fix errors like https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/logs/post-project-infra-prow-control-plane-deployment/1438837799989022720

/cc @dhiller 


dhiller (commented):
Can we then also remove the manifest or is it still used elsewhere? (I didn't find any references though)

EDIT: spelling

fgimenez:
sure the ingress class manifest is already removed in this PR, or do you mean something else?

dhiller (approved):
/approve

PR #1603, opened 2021-09-22
Use images from Quay

fgimenez:
I've mirrored the missing images to quay and made the new repositories public, all is accessible now.

/cc @dhiller 


dhiller (approved):
/approve

Thanks!

PR #1604, opened 2021-09-22
Add blackbox-exporter and probe + alert for prow.ci.kubevirt.io

fgimenez:
* Adds simple deployment + service + configmap for https://github.com/prometheus/blackbox_exporter
* Moved the additionalScrapeConfigs secret to each cluster files, and explicitly create it for each one (in fact no credentials on it) instead of the previous base64 value.
* In the prow-workloads additionalScrapeConfigs secret sets the config to scrape the blackbox-exporter setting the target to https://prow.ci.kubevirt.io. This way we check the status of the service from a cluster different of the one running it. 
* Adds a basic alert for the probe being unsuccessful

/cc @dhiller 


dhiller (approved):
/approve

dhiller:
Great to have that, so I don't get confused next time on a local DNS issue ;-)

PR #1611, opened 2021-09-23
Allow one unavailable pod in several deployments

fgimenez:
We are having a issues with the docker-mirror-proxy and ghproxy deployments, the new pods are blocked in `ContainerCreating` status bc the old pods have the associated volume attached and the new one can't attach it in exclusive mode.

These changes aim to improve automation (no need to manually scale down and up the deployment to unblock de rollout) at the cost of non zero-downtime deployments. In case it is required to replace the pod it might happen that after the old pod is down and before the new pod is up the service provided by it could be unavailable.

/cc @dhiller 


enp0s3 (approved):
Thanks!
/approve

PR #1620, opened 2021-09-24
Symlink prow-deploy requirements.txt to the file used on CI

fgimenez:
`github/ci/prow-deploy/requirements.txt` is only used for local testing during development, it should always have the same content as the one used on CI, `images/prow-deploy/requirements.txt`

/cc @dhiller @enp0s3 


dhiller (approved):
/approve

PR #1621, opened 2021-09-24
Switch hostpath provisioner unit test running to golang container

awels:

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/hostpath-provisioner/hostpath-provisioner-presubmits.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/hostpath-provisioner/hostpath-provisioner-presubmits.yaml
@@ -114,7 +114,7 @@ presubmits:
       preset-docker-mirror-proxy: "true"
     spec:
       containers:
-        - image: quay.io/kubevirtci/bootstrap:v20210311-09ebaa2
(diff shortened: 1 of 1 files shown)


github/ci/prow-deploy/files/jobs/kubevirt/hostpath-provisioner/hostpath-provisioner-presubmits.yaml
```diff
@@ -114,12 +114,15 @@ presubmits:
       preset-docker-mirror-proxy: "true"
     spec:
       containers:
-        - image: quay.io/kubevirtci/bootstrap:v20210311-09ebaa2
+        - image: quay.io/kubevirtci/golang:v20210924-7b670cb
           command:
             - "/usr/local/bin/runner.sh"
             - "/bin/sh"
             - "-c"
             - "make test"
```
fgimenez, line 122: You would need also to execute `eval $(gimme)` either here or as part of `make test`. Maybe better to have all (including the golang version) self-contained in the repo, wdyt?

awels: Okay so don't set the env variable then.

fgimenez (approved):
/approve

Thanks!

PR #1622, opened 2021-09-25
Adds new lane for running periodic realtime perf tests

jordigilh:
Adds a new lane for running the realtime performance tests. These tests capture the maximum CPU latency when running a VM in realtime and compare it against a given latency threshold. The test will pass if the maximum latency is below the given threshold.

@fgimenez @vladikr @acardace @pkliczewski @rmohr @dhiller

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml
@@ -1475,3 +1475,69 @@ periodics:
         resources:
           requests:
             memory: "29Gi"
+- annotations:
+    testgrid-dashboards: kubevirt-periodics
+    testgrid-days-of-results: "60"
+  cluster: prow-workloads
+  cron: 0 6 * * *
+  decorate: true
+  decoration_config:
+    grace_period: 5m0s
+    timeout: 1h0m0s
+  extra_refs:
+  - base_ref: main
+    org: kubevirt
+    repo: kubevirt
+  labels:
+    ci.kubevirt.io/prometheus: ""
+    preset-bazel-cache: "true"
+    preset-bazel-unnested: "true"
+    preset-dind-enabled: "true"
+    preset-docker-mirror-proxy: "true"
+  name: periodic-kubevirt-e2e-k8s-1.22-sig-performance-realtime
+  reporter_config:
+    slack:
+      job_states_to_report: []
+  spec:
+    containers:
+    - command:
+      - /usr/local/bin/runner.sh
+      - /bin/sh
+      - -c
+      - |
+        # Create k8s cluster
+        make cluster-up
+
+        # Push and deploy kubevirt
+        make cluster-sync
+
+        FUNC_TEST_ARGS="--noColor --seed=42" make realtime-perftest
+      env:
+      - name: KUBEVIRT_PROVIDER
+        value: k8s-1.22
+      - name: KUBEVIRT_STORAGE
+        value: rook-ceph-default
+      - name: KUBEVIRT_MEMORY_SIZE
+        value: 9G
+      - name: KUBEVIRT_NUM_NODES
+        value: "2"
+      - name: KUBEVIRT_E2E_REALTIME_PERF_TEST
+        value: "true"
+      - name: KUBEVIRT_E2E_CYCLIC_DURATION_IN_SECONDS
+        value: "60"
+      - name: KUBEVIRT_E2E_REALTIME_LATENCY_THRESHOLD_IN_MICROSECONDS
+        value: "40"
+      - name: KUBEVIRT_HUGEPAGES_2M
+        value: "512"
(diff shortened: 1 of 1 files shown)


github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml
```diff
@@ -1475,3 +1475,67 @@ periodics:
         resources:
           requests:
             memory: "29Gi"
+- annotations:
+    testgrid-dashboards: kubevirt-periodics
+    testgrid-days-of-results: "60"
+  cluster: prow-workloads
+  cron: 40 7,15,23 * * *
```
fgimenez, line 1482: I think we should shift the schedule a bit, there is another periodic lane that starts at the same times.

github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml
```diff
@@ -1475,3 +1475,67 @@ periodics:
...
+      - name: KUBEVIRT_E2E_REALTIME_LATENCY_THRESHOLD
+        value: "40"
+      - name: KUBEVIRT_HUGEPAGES2M_COUNT
+        value: "512"
+      image: quay.io/kubevirtci/bootstrap:v20210906-994b913
+      name: ""
+      resources:
+        requests:
+          cpu: "12"
+          memory: 44Gi
```
fgimenez, line 1537: I think we should adjust the requests, 26Gi of memory (2 nodes * 9 Gi per node + 8 Gi for cluster management). I think we don't need to request cpus, but if we want to match the other perf lane we should request less than 12 (I guess around 8 for 2 nodes + cluster management).

github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml
jordigilh: Changed it to run once a day at 6am.

github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml
jordigilh: Noted. I reduced the memory to 26Gi and cores to 10.

github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml
```diff
@@ -1475,3 +1475,69 @@ periodics:
...
+
+        # Push and deploy kubevirt
+        make cluster-sync
+
+        FUNC_TEST_ARGS="--noColor --seed=42" make realtime-perftest
+      env:
+      - name: KUBEVIRT_PROVIDER
+        value: k8s-1.22
+      - name: KUBEVIRT_STORAGE
+        value: hpp
```
dhiller, line 1519: I'm not clear about how this is going to work. IIRC we only have support for values rook-ceph and rook-ceph-default in kubevirt/kubevirt and I didn't see that you did add it in the PR https://github.com/kubevirt/kubevirt/pull/6182? Or did I overlook a PR in kubevirt/kubevirtci? Could you point me to the right direction here please?

jordigilh: Oh my bad then. I copied it from the [performance job](https://github.com/kubevirt/project-infra/blob/main/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml#L1225-L1226) thinking that it would be the closest example I could use. Should I change it then to `rook-ceph`?

dhiller: No, please use `rook-ceph-default` as that will use the embedded support from kubevirtci. `rook-ceph` would just work when using `automation/test.sh`.

dhiller: @awels or do we implicitly have support for `hpp` value for env `KUBEVIRT_STORAGE` that I am not aware of?

awels: No just rook-ceph-default.

jordigilh: Done :smile:

dhiller (approved):
/approve
/hold

Holding this to wait for the required PR to go in.

dhiller:
@jordigilh I've triggered the lane, results will come up here: https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/logs/periodic-kubevirt-e2e-k8s-1.22-sig-performance-realtime/1446074801864052736

Let's see how it goes.

PR #1629, opened 2021-09-29
[Sig-Scale] adding the config files to create the sig-scale cluster

marceloamaral:
Adding the configuration files and scripts to create the Sig-Scale cluster.
The files are used to configure kubespray, and the scripts are used to create the cluster.

The new Sig-Scale cluster will be used to run performance jobs for the Continuous Performance Evaluation of the KubeVirt control plane. As detailed here: https://github.com/kubevirt/kubevirt/issues/6309

There are still two missing pieces to automate the cluster creation:
- add the needed private key in `'prowPerformance.sshKey'`
- create a new configmap `'prowPerformance.hostsYml'` with the cluster configuration

### DIFF
--- a/github/ci/prow-performance-workloads/bootstrap.yml
+++ b/github/ci/prow-performance-workloads/bootstrap.yml
@@ -1,0 +1,6 @@
+---
+- hosts: kube-node
+  tasks:
+    - name: include bootstrap role
+      include_role:
+        name: bootstrap
--- a/github/ci/prow-performance-workloads/hack/bootstrap.sh
+++ b/github/ci/prow-performance-workloads/hack/bootstrap.sh
@@ -1,0 +1,13 @@
+#!/bin/bash
+
+source $(dirname "$0")/common.sh
+
+main(){
+    setup
+
+    cd ${BASE_DIR}
+
+    ansible-playbook -i inventory/prow-performance-workloads/hosts.yaml  --become --become-user=root --private-key ~/.ssh/id_rsa bootstrap.yml
+}
+
+main "$@"
--- a/github/ci/prow-performance-workloads/hack/common.sh
+++ b/github/ci/prow-performance-workloads/hack/common.sh
@@ -1,0 +1,18 @@
+#!/bin/bash
+
+PROJECT_PATH=$(realpath $(dirname "$0")/../../../..)
+BASE_DIR=${PROJECT_PATH}/github/ci/prow-performance-workloads
+
+setup(){
+    cd ${PROJECT_PATH}
+
+    source ./hack/manage-secrets.sh
+    decrypt_secrets
+    # ssh key
+    mkdir -p ~/.ssh && chmod 0700 ~/.ssh
+    extract_secret 'prowPerformance.sshKey' ~/.ssh/id_rsa
+    chmod 0600 ~/.ssh/id_rsa
+    # ansible config files
+    extract_secret 'prowPerformance.hostsYml' ${BASE_DIR}/inventory/prow-performance-workloads/hosts.yml
+    extract_secret 'prowPerformance.groupVarsAllYml' ${BASE_DIR}/inventory/prow-performance-workloads/group_vars/all/all.yml
+}
--- a/github/ci/prow-performance-workloads/hack/deploy.sh
+++ b/github/ci/prow-performance-workloads/hack/deploy.sh
@@ -1,0 +1,13 @@
+#!/bin/bash
+
+source $(dirname "$0")/common.sh
+
+main(){
+    setup
+
+    cd ${BASE_DIR}
+
+    ansible-playbook -i inventory/prow-performance-workloads/hosts.yaml  --become --become-user=root --private-key ~/.ssh/id_rsa /kubespray/cluster.yml
+}
+
+main "$@"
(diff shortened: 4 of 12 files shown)


marceloamaral:
cc @fgimenez

github/ci/prow-sig-scale-workloads/hack/bootstrap.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+source $(dirname "$0")/common.sh
+
+main(){
+    setup
+
+    cd ${BASE_DIR}
+
+    ansible-playbook -i inventory/prow-performance-workloads/hosts.yaml  --become --become-user=root --private-key ~/.ssh/id_ed25519 bootstrap.yml
```
fgimenez, line 10: Is the ssh an `ed25519` key? If it's an rsa (seems like it looking at the private part) we should name it accordingly.

github/ci/prow-sig-scale-workloads/hack/common.sh
```diff
@@ -0,0 +1,18 @@
...
+BASE_DIR=${PROJECT_PATH}/github/ci/prow-performance-workloads
+
+setup(){
+    cd ${PROJECT_PATH}
+
+    source ./hack/manage-secrets.sh
+    decrypt_secrets
+    # ssh key
+    mkdir -p ~/.ssh && chmod 0700 ~/.ssh
+    extract_secret 'prowPerformance.sshKey' ~/.ssh/id_ed25519
```
fgimenez, line 13: Same as above, we should name the ssh private key according to its type.

github/ci/prow-sig-scale-workloads/hack/bootstrap.sh
fgimenez, line 10: I think the name under the `inventory` dir should match the top-level dir, IMO we should either name both `prow-sig-scale-workloads` or `prow-performance-workloads`.

github/ci/prow-sig-scale-workloads/roles/bootstrap/tasks/main.yml
```diff
@@ -0,0 +1,133 @@
...
+      replace:
+        path: /etc/fstab
+        regexp: '(.*swap.*)'
+        replace: '#\1'
+    - name: mask systemctl swap unit
+      systemd:
+        name: dev-sda3.swap
+        masked: yes
+
+- name: enable mirrors for yum repos
```
fgimenez, line 17: This was needed in the `prow-workloads` bare metals, maybe the new nodes have a different setup, have you checked if these changes are required? You can verify by trying to install any package from the configured repos.

github/ci/prow-sig-scale-workloads/roles/bootstrap/tasks/main.yml
```diff
@@ -0,0 +1,133 @@
...
+      - docker-latest-logrotate
+      - docker-logrotate
+      - docker-engine
+      - docker-ce 
+      - docker-ce-cli 
+    state: absent
+
+- name: NetworkManager DNS config
+  lineinfile:
+    path: /etc/sysconfig/network-scripts/ifcfg-bond1
```
fgimenez, line 65: Same here, have you checked if the nodes have a `bond1` nic?

github/ci/prow-sig-scale-workloads/roles/bootstrap/tasks/main.yml
```diff
@@ -0,0 +1,133 @@
...
+- name: configure fs.inotify
+  copy:
+    dest: /etc/sysctl.d/fs.inotify.max_user_watches.conf
+    content: |
+      fs.inotify.max_user_watches=524288
+
+- name: additional kernel modules
+  include_tasks: additional_kernel_modules.yml
+  loop:
+    - openvswitch
```
fgimenez, line 108: This was required by very specific sig-network tests, probably not needed in the perf cluster.

github/ci/prow-sig-scale-workloads/hack/bootstrap.sh
marceloamaral: It is an `rsa`. I will update it

marceloamaral: I started with `sig-scale-workloads`, then changed to `prow-performance-workloads`
It is a left over, I will update it

github/ci/prow-sig-scale-workloads/roles/bootstrap/tasks/main.yml
marceloamaral: I needed to install `yum install epel-release`
So, that is why I kept this part, the BM also need some repo configuration

github/ci/prow-sig-scale-workloads/roles/bootstrap/tasks/main.yml
marceloamaral: Yes, `bond0` and `bond1`

github/ci/prow-sig-scale-workloads/roles/bootstrap/tasks/main.yml
marceloamaral: Humm...I was not sure about this.
Right, we do not need it

marceloamaral:
@fgimenez thanks for the comments
It is ready to review again!

fgimenez (approved):
/approve

Awesome, thanks!

PR #1631, opened 2021-09-29
switch remaining hpp tests to golang container

awels:
Switch the remaining tests to the golang container as they use it
also.


aglitke (approved):
/lgtm

PR #1636, opened 2021-09-30
Set KubePodNotReady alert trigger to 90m in prow-workloads

fgimenez:
* Creates different definitions of `KubePodNotReady` for control plane and workloads clusters.
* Sets the trigger to 90m for workloads cluster.

/cc @dhiller 


dhiller (approved):
/approve

PR #1637, opened 2021-09-30
Switch hpp operator to use golang container

awels:
The hpp operator scripts are very similar to hpp and thus
should be using the golang container as well for testing
and releases.


aglitke (approved):
/lgtm

PR #1645, opened 2021-10-04
Move hpp(-operator) unit tests to phx-prow cluster

awels:
They are currently marked on prow-workloads but don't require
bm at all to run. So phx-prow is better.


fgimenez (approved):
/approve

Thanks!

PR #1666, opened 2021-10-11
add test suits arm64 to periodic E2E tests for ARM

zhlhahaha:


github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml
```diff
@@ -1134,6 +1134,10 @@ periodics:
         curl -Lo ./yq https://github.com/mikefarah/yq/releases/download/3.4.1/yq_linux_amd64
         chmod +x ./yq && mv ./yq /usr/local/bin/yq
 
+        # get kubectl
+        curl -L "https://dl.k8s.io/release/v1.19.7/bin/linux/amd64/kubectl" -o /usr/local/bin/kubectl
```
rmohr, line 1138: Is 1.19.7 what we use on the arm servers?

Independent of that I don't think it matters much if we have newer servers. We do not really depende too much on kubectl specific features which change between versions.

zhlhahaha: Yes, I agree with you.
```
Client Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.7", GitCommit:"1dd5338295409edcfff11505e7bb246f0d325d15", GitTreeState:"clean", BuildDate:"2021-01-13T13:23:52Z", GoVersion:"go1.15.5", Compiler:"gc", Platform:"linux/arm64"}
```

Issue #1649, opened 2021-10-05
KubeVirt and KUBEVIRT_NUM_NODES=3


rmohr:
I have a case where I would need a three node cluster. We should probably consider if we can add such a cluster to for every PR right now. 
* We could either increase the number of nodes and maybe also increase the number of parallel test runs, 
* or we could just increase the number of nodes for the periodics
* or we can have our onw lane which just runs very short for a small number of tests

The PR where I have code which would need that is: https://github.com/kubevirt/kubevirt/pull/6522

rmohr:
@fgimenez FYI where do we stand on the resource perspective?

fgimenez:
We have been at capacity for some time, around 12 UTC every weekday jobs start queuing. For instance, at this moment there are e2e 19 jobs that have been waiting in the queue for more than 1h30m. I would suggest using the approach that requires fewer resources, maybe a periodic lane with the increased number of nodes.

PR #1670, opened 2021-10-12
Add a migration test lane with three lanes

rmohr:
This is neede for some e2e tests which need two nodes with CPU manager
and hugepages available. Starting with the job as optional.

The tests are introduced in https://github.com/kubevirt/kubevirt/pull/6522.

Fixes https://github.com/kubevirt/project-infra/issues/1649


rmohr:
> Looks good, but I think we need changes in kubevirt/kubevirt, we are hardcoding here https://github.com/kubevirt/kubevirt/blob/main/automation/test.sh#L69-L76 the number of nodes to use and we are not taking into account the env var.

Yes you are right, I added this to https://github.com/kubevirt/kubevirt/pull/6522/commits/21686e002f6ddab427c028a83902abcbea134c39. Forgot to mention it.

fgimenez:
/lgtm

cool thanks!

PR #1678, opened 2021-10-15
Add a presubmit job of hco in openshift prow into testgrid

erkanerol:
We want to track jobs of hyperconverged cluster operator in kubevirt's testgrid since openshift's testgrid doesn't belong upstream projects.

This PR tries to add "pull-ci-kubevirt-hyperconverged-cluster-operator-main-hco-e2e-upgrade-prev-index-azure" into testgrid as a first example. The definition of that job stands here: https://github.com/openshift/release/blob/6f928dd345bc49fc52561e124d169bf7873f4fa0/ci-operator/jobs/kubevirt/hyperconverged-cluster-operator/kubevirt-hyperconverged-cluster-operator-main-presubmits.yaml#L922

As far as I see from other examples*, we have to add the jobs here one by one and there is no easy way to add all jobs by using an annotation. See
- https://github.com/kubernetes/test-infra/blob/master/config/testgrids/openshift/openshift-virtualization.yaml
- https://github.com/kubernetes/test-infra/blob/master/config/testgrids/jetstack/gen-config.yaml

### DIFF
--- a/github/ci/testgrid/gen-config.yaml
+++ b/github/ci/testgrid/gen-config.yaml
@@ -5,6 +5,19 @@ dashboards:
 - name: kubevirt-libguestfs-appliance-periodics
 - name: kubevirt-hyperconverged-cluster-operator-presubmits
 - name: kubevirt-hyperconverged-cluster-operator-periodics
+- name: kubevirt-hyperconverged-cluster-operator-openshift-ci-presubmits
+  dashboard_tab:
+  - name: pull-ci-kubevirt-hyperconverged-cluster-operator-main-hco-e2e-upgrade-prev-index-azure
+    test_group_name: pull-ci-kubevirt-hyperconverged-cluster-operator-main-hco-e2e-upgrade-prev-index-azure
+    open_test_template:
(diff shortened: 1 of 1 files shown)


github/ci/testgrid/gen-config.yaml
```diff
@@ -5,7 +5,19 @@ dashboards:
 - name: kubevirt-libguestfs-appliance-periodics
 - name: kubevirt-hyperconverged-cluster-operator-presubmits
 - name: kubevirt-hyperconverged-cluster-operator-periodics
-
+- name: kubevirt-hyperconverged-cluster-operator-openshift-ci-presubmits
+  dashboard_tab:
+    - name: pull-ci-kubevirt-hyperconverged-cluster-operator-main-ci-index
```
fgimenez, line 10: Could you please decrease the indentation of the array element 2 spaces? see how other array elements are indented in this file

fgimenez (changes requested):
Awesome! Only a small nit, could you pls add blank lines between the `dashboards`, `dashboard_groups` and `test_groups` blocks?

fgimenez (approved):
/approve

Thanks!

PR #1680, opened 2021-10-15
Request an API repository for KubeVirt

rmohr:
This repository will eventually repolace client-go by only exporting the
API of KubeVirt which can then be usede by client-gen.


fgimenez (approved):
/approve

PR #1682, opened 2021-10-17
Add all jobs of hco in openshift ci

erkanerol:
As HCO team, we want to add our jobs in OpenShift ci into this testgrid so that we can track our tests statuses and flakiness. 

With https://github.com/kubevirt/project-infra/pull/1678, I added a job as an example and we observed that it works. See https://testgrid.k8s.io/kubevirt-hyperconverged-cluster-operator-openshift-ci-presubmits

With this PR, I am adding all jobs  here https://github.com/openshift/release/tree/master/ci-operator/jobs/kubevirt/hyperconverged-cluster-operator

Note that some tests have not junit.xml output yet and the results don't seem tidy in testgrid now. We will fix them after adding them into testgrid. 


fgimenez (approved):
/approve

Thanks!

PR #1701, opened 2021-10-29
Fix workloads cluster deployment

fgimenez:
Currently we are not continuously deploying to the workloads cluster, these changes set the job to run on relevant changes and update some of the resources to be able to be deployed there. 

They also add logic in the deployment role for not using server-side apply on the workloads cluster, which runs k8s 1.19 and doesn't work well with it. In any case, server-side apply is required only for Prow control plane because of the length of the CRDs. In any case we need to update the workloads cluster k8s version, but for now these changes will allow us to keep deploying changes there.

Tested successfully here https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/logs/post-project-infra-prow-workloads-cluster-deployment/1454002949243015168

/cc @dhiller @rmohr 


rmohr (approved):
/lgtm
/approve

PR #1702, opened 2021-10-29
Perfscale-audit binary is missing from the Perf periodic

rthallisey:
Run `make go-build` to build the binary locally in order for the test to run

Signed-off-by: Ryan Hallisey <rhallisey@nvidia.com>

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml
@@ -1211,6 +1211,9 @@ periodics:
       - /bin/sh
       - -c
       - |
+        # Build the perfscale-audit binary
+        make bazel-build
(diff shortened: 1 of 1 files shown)


github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml
```diff
@@ -1211,6 +1211,9 @@ periodics:
       - /bin/sh
       - -c
       - |
+        # Build the perfscale-audit binary
+        make go-build
+
```
davidvossel, line 1216: why go-build instead of make bazel-build?

rthallisey: Thought I changed this... it should be bazel-build

fgimenez (approved):
/approve

PR #1709, opened 2021-11-03
Do not use GIT_ASKPASS for prow-deploy-test

fgimenez:
Fixes errors like https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_project-infra/1701/pull-project-infra-prow-deploy-test/1455871425150717952 The `GIT_ASKPASS` definition is a leftover, now that we don't mount the github token it has started causing the job to fail.

Also added code to deploy the unprivileged github token used during tests, and changed the `pull-project-infra-prow-deploy-test` to mount and use it.

/cc @rmohr @dhiller  

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/project-infra/project-infra-presubmits.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/project-infra/project-infra-presubmits.yaml
@@ -359,8 +359,8 @@ presubmits:
           env:
           - name: GOOGLE_APPLICATION_CREDENTIALS
             value: /etc/gcs/service-account.json
-          - name: GIT_ASKPASS
-            value: "/home/prow/go/src/github.com/kubevirt/project-infra/hack/git-askpass.sh"
+          - name: GITHUB_TOKEN
+            value: "/etc/github/oauth"
           command:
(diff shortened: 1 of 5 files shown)


github/ci/prow-deploy/kustom/overlays/ibmcloud-production/kustomization.yaml
```diff
@@ -281,11 +281,6 @@ secretGenerator:
     files:
       - oauth=secrets/oauth-token
     type: Opaque
-  - name: oauth-token
```
fgimenez, line 284: Cleanup, this entry was repeated.

fgimenez:
This is ready to go, I've taken the opportunity to deploy the unprivileged token as a secret and remove the need to decrypt the secrets. Successful execution with all the changes in place here https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_project-infra/1709/pull-project-infra-prow-deploy-test/1456172592237907968

@rmohr I needed to push more changes and the lgtm was removed, can you please take another look?

fgimenez:
/unhold

thanks!

PR #1713, opened 2021-11-08
Make the required switch from 1.19, 1.20 to 1.21, 1.22 in CDI

akalenyu:
Other than that sticking to the same subset as we have now & deploy prometheus on 2 lanes & cleanup of entries.

Signed-off-by: Alex Kalenyuk <akalenyu@redhat.com>

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/containerized-data-importer/containerized-data-importer-presubmits-1.34.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/containerized-data-importer/containerized-data-importer-presubmits-1.34.yaml
@@ -1,0 +1,68 @@
+presubmits:
+  kubevirt/containerized-data-importer:
+  - name: pull-containerized-data-importer-e2e-k8s-1.19-hpp
+    branches:
+      - release-v1.34
+    annotations:
+      fork-per-release: "true"
+    always_run: true
+    optional: false
+    decorate: true
+    decoration_config:
+      timeout: 5h
+      grace_period: 5m
+    max_concurrency: 6
+    cluster: prow-workloads
+    labels:
+      preset-dind-enabled: "true"
+      preset-docker-mirror-proxy: "true"
+      preset-shared-images: "true"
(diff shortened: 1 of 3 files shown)


awels:
I would just delete the old lanes completely and make the new ones mandatory and always run. Instead of making the old lanes not required and not always run.

akalenyu:
> I would just delete the old lanes completely and make the new ones mandatory and always run. Instead of making the old lanes not required and not always run.

what about the release branches that still need those? 1.34/1.38 etc

awels:
You can make presubmits for those specific branches. There are some examples for previous branches already

akalenyu:
> See for instance https://github.com/kubevirt/project-infra/blob/main/github/ci/prow-deploy/files/jobs/kubevirt/containerized-data-importer/containerized-data-importer-presubmits-1.28.yaml#L4-L5 which tells it to run for the 1.28 branch. You can do something similar for the older branches that you are concerned about. Basically copy the job and setting the branch it reacts to by listing the branches.

Yep, that was the plan, but I don't see a 1.38 file anywhere so it made me wonder how we even have 1.20 lanes running on that

awels:
If you check the different lanes on the main one, some of them are configured to run on basically every lane. And others are configured to run on only main. That is why the number of tests run in branches is much smaller.

akalenyu:
> If you check the different lanes on the main one, some of them are configured to run on basically every lane. And others are configured to run on only main. That is why the number of tests run in branches is much smaller.

oh ok I misinterpreted the `always_run` flag. Yeah will need to make some changes then

awels:
The always_run flag means always run this lane on a PR update for branches that match the filter (or don't match the skip filter)

akalenyu:
> The always_run flag means always run this lane on a PR update for branches that match the filter (or don't match the skip filter)



> The always_run flag means always run this lane on a PR update for branches that match the filter (or don't match the skip filter)

Updated

awels (approved):
/lgtm
/approve

PR #1719, opened 2021-11-10
add new group to allow rerun jobs in HCO

nunnatsa:
Also, update the members for must-gather team.

### DIFF
--- a/github/ci/prow-deploy/files/orgs.yaml
+++ b/github/ci/prow-deploy/files/orgs.yaml
@@ -651,7 +651,8 @@ orgs:
         members:
           - pkliczewski
           - mmirecki
-          - yuvalturg
+          - nunnatsa
+          - tiraboschi
         privacy: closed
         repos:
           must-gather: admin
@@ -714,6 +715,11 @@ orgs:
           - isaacdorfman
         repos:
           terraform-provider-kubevirt: admin
+      rerun-jobs-hco:
+        description: can rerun failed job in the hyperconverged-cluster-operator repository
+        maintainers:
+        - nunnatsa
+        - tiraboschi
       sig-security:
         description: ""
         maintainers:


github/ci/prow-deploy/files/orgs.yaml
```diff
@@ -714,6 +715,13 @@ orgs:
           - isaacdorfman
         repos:
           terraform-provider-kubevirt: admin
+      rerun-jobs-hco:
+        description: can rerun failed job in the hyperconverged-cluster-operator repository
+        maintainers:
+        - nunnatsa
+        - tiraboschi
+        repos:
+          hyperconverged-cluster-operator: admin
```
dhiller, line 724: This line would make you repo admin. In general we rather strive to remove those privileges from users and rely on peribolos for org and repo management instead. Peribolos will perform the changes from this PR once it gets merged. And it will revert any changes you do to the repo manually.

If you don't need this right, please remove the full repos: section here.

dhiller (changes requested):

github/ci/prow-deploy/files/orgs.yaml
```diff
@@ -714,6 +715,11 @@ orgs:
           - isaacdorfman
         repos:
           terraform-provider-kubevirt: admin
+      rerun-jobs-hco:
+        description: can rerun failed job in the hyperconverged-cluster-operator repository
+        maintainers:
```
dhiller, line 720: nit: I even think that we could also just use members: here instead, but it should be fine. Just remember that while you can change the members of the team then again peribolos would overwrite your changes.

dhiller (approved):
/approve

dhiller:
@nunnatsa the id of the created team is 5359252, do you want to create the pr for adding to rerun_auth_config as bespoke?

PR #1734, opened 2021-11-17
Include skopeo in the bootstrap image

rmohr:
The tool is useful to query remote images. Needed to determine in https://github.com/kubevirt/containerdisks to determine if images need to be updated.


fgimenez (approved):
/approve

PR #1739, opened 2021-11-19
Initial set of jobs for kubevirt/containerdisks

rmohr:
* An optional presubmit job which solely builds without pushing images
* A daily periodic job which determines if new images need to be built and pushes them on demand.


rmohr:
/approve

Thanks @ksimon1 will iterate on them if they don't work properly.

PR #1740, opened 2021-11-19
Move jobs scheduled on phx-prow to ibm-prow-jobs

fgimenez:
phx-prow cluster is being decommissioned by EOY and the jobs running there are giving a lot of connectivity issues, we need to migrate all the jobs currently running there. The end goal is having a new separate workloads cluster, it is not yet in place but for now we have already scaled up ibm-prow-jobs cluster to handle the load. 

This PR takes care of configuring all the jobs currently running on phx-prow to run on ibm-prow-jobs, it also removes the node selector settings that are only meaningful in phx-prow.

/cc @rmohr @dhiller 


rmohr:
/lgtm
/approve

looks good.

PR #1742, opened 2021-11-19
Ensure that go is available for the developer build pusher

rmohr:
The build now executes `hack/publish-staging.sh` which needs golang.


dhiller (approved):
/approve

PR #1743, opened 2021-11-19
Bump bootstrap image to fc34

rmohr:
fc32 is EOL for quite some time and newer versions of some tools are
needed.


rmohr:
/close

@fgimenez has something prepared already.

fgimenez:
> @fgimenez has something prepared already.

indeed, draft PR ready here #1746, will ping you once it is tested

PR #1749, opened 2021-11-22
Make go available for nightly builds

rmohr:
Go is needed to build tooling for syncing staging to their target repos.


dhiller (approved):
/approve

PR #1752, opened 2021-11-22
Ignore unknown cotexts for nmstate/kubernetes-nmstate

fgimenez:
Will prevent tide being blocked because of gh actions errors

/cc @qinqon @dhiller 

### DIFF
--- a/github/ci/prow-deploy/kustom/base/configs/current/config/config.yaml
+++ b/github/ci/prow-deploy/kustom/base/configs/current/config/config.yaml
@@ -186,6 +186,10 @@ tide:
         dashboard:
           skip-unknown-contexts: true
           from-branch-protection: true
+     nmstate:
+      repos:
+        kubevirt-nmstate:
+          skip-unknown-contexts: true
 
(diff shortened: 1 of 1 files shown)


github/ci/prow-deploy/kustom/base/configs/current/config/config.yaml
```diff
@@ -186,6 +186,10 @@ tide:
         dashboard:
           skip-unknown-contexts: true
           from-branch-protection: true
+     nmstate:
+      repos:
```
qinqon, line 190: It has bad identation, it's missing an space.

qinqon (changes requested):

github/ci/prow-deploy/kustom/base/configs/current/config/config.yaml
fgimenez: thanks for the heads up! it is not required for the context_options to be applied but good to have it fixed, already pushed the changes

dhiller (approved):
/approve

PR #1758, opened 2021-11-23
We need an absolute askpass path due to directory switching

rmohr:
`hack/publish-staging.sh` is switching its path. If the git askpass path is not absolute, it will point to a non-existing file.


fgimenez (approved):
/approve

PR #1762, opened 2021-11-24
Mirror images from docker.io to quay.io

erkanerol:
docker.io has a pull limit and we want to consume some images from quay.io. 

This PR add a postsubmit job which mirror some specific images from docker.io to quay.io. It aims to add an automated script which will work whenever you add a new image into the list. 

**Open Points**
The list of images is not clear yet.  Please add the images as comments. 

**Notes:**
As far as I see, calico is fetched from quay.io now.  https://github.com/kubevirt/kubevirtci/blob/main/cluster-provision/k8s/1.22/extra-pre-pull-images

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/project-infra/project-infra-postsubmits.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/project-infra/project-infra-postsubmits.yaml
@@ -1019,3 +1019,29 @@ postsubmits:
                 memory: "8Gi"
               limits:
                 memory: "8Gi"
+    - name: post-project-infra-mirror-images-from-docker-hub
+      always_run: false
+      run_if_changed: "hack/mirror-images.sh|hack/images_to_mirror.csv"
+      annotations:
+        testgrid-create-test-group: "false"
+      decorate: true
+      cluster: ibm-prow-jobs
+      labels:
+        preset-kubevirtci-quay-credential: "true"
+        preset-docker-mirror-proxy: "true"
+      spec:
+        containers:
+        - image: quay.io/kubevirtci/bootstrap:v20211122-ca88fb4
+          command: ["/bin/sh"]
+          args:
+          - -c
+          - |
+            set -e
+
+            cat "$QUAY_PASSWORD" | docker login --username $(cat "$QUAY_USER") --password-stdin=true quay.io
+            hack/mirror-images.sh
+          resources:
+            requests:
+              memory: "200Mi"
+            limits:
+              memory: "200Mi"
--- a/hack/mirror-images.sh
+++ b/hack/mirror-images.sh
@@ -1,0 +1,28 @@
+#!/bin/bash
+
+set -euo pipefail
+
+if ! command -V skopeo; then
+    echo "skopeo required!"
+    exit 1
+fi
+
+BASEDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
+IMAGE_LIST_FILE="${BASEDIR}/images_to_mirror.csv"
+IMAGE_LIST=($(cat "${IMAGE_LIST_FILE}"))
+
+for image in "${IMAGE_LIST[@]}";
+do
+  source_registry="$(echo "$image" |cut -d',' -f1)"
+  image_in_source="$(echo "$image" |cut -d',' -f2)"
+  target_registry="$(echo "$image" |cut -d',' -f3)"
+  # since we are mirroring all images into a specific organization
+  # we cannot use full path of original image which contain user+image name.
+  # example: docker.io/user/image:v1 -> quay.io/kubevirtci/user-image:v1
+  image_in_target="${image_in_source//\//-}" # replace / with -
+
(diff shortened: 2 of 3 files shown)


github/ci/prow-deploy/files/jobs/kubevirt/project-infra/project-infra-postsubmits.yaml
```diff
@@ -1048,3 +1048,28 @@ postsubmits:
...
                 memory: "8Gi"
+    - name: mirror-images-from-docker-hub
+      always_run: false
+      run_if_changed: "hack/mirror-images.sh"
+      annotations:
+        testgrid-create-test-group: "false"
+      decorate: true
+      cluster: ibm-prow-jobs
+      labels:
+        preset-kubevirtci-quay-credential: "true"
```
fgimenez, line 1059: Given that we are going to deal with docker hub it would be good to include docker proxy settings `preset-docker-mirror-proxy: "true"`

github/ci/prow-deploy/files/jobs/kubevirt/project-infra/project-infra-postsubmits.yaml
```diff
@@ -1048,3 +1048,28 @@ postsubmits:
...
+      spec:
+        containers:
+        - image: quay.io/kubevirtci/bootstrap:v20210906-994b913
```
fgimenez, line 1062: We should use the latest bootstrap image, it is based in f34 and ships a recent skopeo.

hack/mirror-images.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
+
+set -euo pipefail
+
+if command -V skopeo; then
+    echo "skopeo is available"
+else
+    echo "skopeo is not available. installing..."
+    sudo dnf -y install skopeo
+fi
```
fgimenez, line 10: No need for this if we use a recent bootstrap image

github/ci/prow-deploy/files/jobs/kubevirt/project-infra/project-infra-postsubmits.yaml
erkanerol: Added.

github/ci/prow-deploy/files/jobs/kubevirt/project-infra/project-infra-postsubmits.yaml
erkanerol: Updated

hack/mirror-images.sh
erkanerol: Removed

erkanerol:
@fgimenez  I updated my PR according to your comments. Now, we need to clarify the image list.

fgimenez:
> @fgimenez I updated my PR according to your comments. Now, we need to clarify the image list.

@erkanerol I think the logic is good to go. The image list is ok to start too IMO. But I think there is no proper separation of concerns, it would be great to have the image list (the data of the application) on a separate file, this way people interested in maintaining the mirrored images can propose changes to that file without worrying about the code, wdyt?

erkanerol:
> > @fgimenez I updated my PR according to your comments. Now, we need to clarify the image list.
> 
> @erkanerol I think the logic is good to go. The image list is ok to start too IMO. But I think there is no proper separation of concerns, it would be great to have the image list (the data of the application) on a separate file, this way people interested in maintaining the mirrored images can propose changes to that file without worrying about the code, wdyt?

Done :)

github/ci/prow-deploy/files/jobs/kubevirt/project-infra/project-infra-postsubmits.yaml
```diff
@@ -1019,3 +1020,28 @@ postsubmits:
                 memory: "8Gi"
               limits:
                 memory: "8Gi"
+    - name: mirror-images-from-docker-hub
```
fgimenez, line 1023: Can you please rename the job to add the prefix `post-project-infra-`?

fgimenez (changes requested):
Tried to run it here https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/logs/mirror-images-from-docker-hub/1466051144693321728 but looks like `hack/mirror-images.sh` is not executable.

github/ci/prow-deploy/files/jobs/kubevirt/project-infra/project-infra-postsubmits.yaml
erkanerol: Done

erkanerol:
> Tried to run it here https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/logs/mirror-images-from-docker-hub/1466051144693321728 but looks like `hack/mirror-images.sh` is not executable.

I fixed that issue :+1:

fgimenez (approved):
/approve

Awesome, successful execution here https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/logs/post-project-infra-mirror-images-from-docker-hub/1466355516941799424 thanks!

PR #1764, opened 2021-11-25
Update kubespray context to enable ansible-playbook find the right inventory configuration

marceloamaral:
After deploying the performance cluster, I realized that the version of Kubernetes was wrong.

We configured the cluster to deploy Kubernetes 1.21 https://github.com/kubevirt/project-infra/blob/05413874f9fe7bfe148eb503e7489225219[…]ow-performance-workloads/group_vars/k8s_cluster/k8s-cluster.yml
But, version 1.19 has been installed....

Kubespray uses ansible-playbook to deploy the cluster, where the cluster settings are in `inventory / prow-performance-workloads` in our case.

Previously, we were running the following command:
`ansible-playbook -i inventory/prow-performance-workloads /hosts.yml --become --become-user = root --private-key ~ /.ssh/id_rsa /kubespray/cluster.yml`

In the previous command, the `inventory/prow-performance-workloads` folder is not inside the `/kubespray/inventory` folder. So I believe this is the reason why the ansible-playbook does not identify the correct inventory context.

Therefore, with PR, I'm moving `inventory/prow-performance-workloads` to the kubespray folder like the kubespray documentation recommends to do.

### DIFF
--- a/github/ci/prow-performance-workloads/hack/deploy.sh
+++ b/github/ci/prow-performance-workloads/hack/deploy.sh
@@ -5,9 +5,11 @@ source $(dirname "$0")/common.sh
 main(){
     setup
 
-    cd ${BASE_DIR}
+    cd /kubespray
+    
+    cp -r ${BASE_DIR}/inventory/prow-performance-workloads inventory/
 
-    ansible-playbook -i inventory/prow-performance-workloads/hosts.yml  --become --become-user=root --private-key ~/.ssh/id_rsa /kubespray/cluster.yml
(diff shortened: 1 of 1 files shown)


fgimenez (commented):
Hmm not sure how this can help, if we change directory to /kubespray then the relative inventory path passed with `-i` to `ansible-playbook` will be wrong. Also, `group_vars` and `host_vars` directories won't be found, ansible looks for them in the inventory's directory, see https://docs.ansible.com/ansible/latest/user_guide/intro_inventory.html#organizing-host-and-group-variables

PR #1766, opened 2021-11-25
sriov-1.19, Readd check-up-kind-1.19-sriov lane

oshoval:
We still use kind-1.19-sriov on kubevirt,
let's return the lane until we fully switch to sriov-1.22,
so it will be part of the gating.


fgimenez (approved):
/approve

thanks!

oshoval:
/hold
thanks
lets see it passes

ormergi (approved):
Thanks

oshoval:
Thanks

2 tests are flaky, but since kubevirt also uses 1.19 its not due to this PR
and at least one of them i saw already anyhow flakes in the past

removing the hold, unless you think something else

/hold cancel

oshoval:
nothing changed, i just think that maybe because there was a rehearse which didn't pass it might hold the PR
so pushed an empty amend

fgimenez (approved):
/lgtm

PR #1771, opened 2021-11-29
Rebuild golang image if bootstrap changes

rmohr:

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/project-infra/project-infra-postsubmits.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/project-infra/project-infra-postsubmits.yaml
@@ -85,7 +85,7 @@ postsubmits:
                 memory: "1Gi"
     - name: publish-bootstrap-image
       always_run: false
-      run_if_changed: "images/bootstrap/.*"
+      run_if_changed: "images/golang/.*|images/bootstrap/.*"
       annotations:
         testgrid-create-test-group: "false"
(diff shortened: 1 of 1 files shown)


fgimenez:
Good catch! But are we tagging bootstrap with `latest`? Not sure if the golang image is going to pick the right one as base https://github.com/kubevirt/project-infra/blob/main/images/golang/Dockerfile#L1

rmohr:
> Good catch! But are we tagging bootstrap with `latest`? Not sure if the golang image is going to pick the right one as base https://github.com/kubevirt/project-infra/blob/main/images/golang/Dockerfile#L1

Hm, that does obviously not work. Right. :/

I  think I never intended treating them in separate jobs. I will just combine them. It will rebuild bootstrap then also when only golang is changed, but I think that is still good enough.

rmohr:
@fgimenez PTAL.

fgimenez:
I think we still need to tag bootstrap with `latest` at least locally (no need to push it) before building the golang image, if not it is not going to be available, wdyt?

rmohr:
> I think we still need to tag bootstrap with `latest` at least locally (no need to push it) before building the golang image, if not it is not going to be available, wdyt?

It is tagged with `bootstrap` already in addition with its `quay.io` name: https://github.com/kubevirt/project-infra/blob/main/images/publish_image.sh#L68. I think that should be sufficient.

fgimenez (approved):
/approve

PR #1778, opened 2021-11-30
add prow job to deploy prometheus and grafana in the performance cluster

marceloamaral:
Scripts to deploy prometheus operator with grafana in the performance cluster

In order to use the updated metrics from k8s and the grafana dashboard in the performance cluster, I needed to implement a new set of rules from prometheus.

I also needed to move service-monitors to the service-monitor folders to create it in the right namespace. Before, everything was deployed in the monitoring namespace and it couldn't find the endpoints.

I also needed to fix some matchLabels and jobLabel rules in some services and service monitor definitions to find the endpoints in the performance cluster and generate the correct job name in the metrics to be compatible with the grafana dashboards.

/cc @fgimenez @dhiller 


github/ci/prow-deploy/files/jobs/kubevirt/project-infra/project-infra-postsubmits.yaml
```diff
@@ -637,6 +637,106 @@ postsubmits:
...
+              # deploy a storage class to use with prometheus, so the pod restart doesn't lose everything
+              kubectl apply -f https://github.com/rancher/local-path-provisioner/raw/master/deploy/local-path-storage.yaml
+
+              # install helm to deploy prometheus operator
+              curl -fsSL -o get_helm.sh https://raw.githubusercontent.com/helm/helm/main/scripts/get-helm-3
+              chmod 700 get_helm.sh
+              ./get_helm.sh
+
+              helm repo add prometheus-community https://prometheus-community.github.io/helm-charts
+              helm install \
```
fgimenez, line 694: Why introduce a new tool? We don't use helm in any of our other deployments. 

As I see it it would be a good idea to align with what we already have using bazel's rules_gitops, including tests, making it super easy to integrate with thanos and our alerting pipeline.

For instance this is the deployment job of the monitoring stack https://github.com/kubevirt/project-infra/blob/main/github/ci/prow-deploy/files/jobs/kubevirt/project-infra/project-infra-postsubmits.yaml#L795 we would only need to add a new line for deploying to the new environment and entries for the new cluster here https://github.com/kubevirt/project-infra/tree/main/github/ci/services/prometheus-stack let me know wdyt

marceloamaral: Right, it might be better to make the prometheus deployment consistent with the other clusters, especially since you have some testing...

The idea of using helm was because it is very easy to update the prometheus-stack versions....

My only concern is with respect to grafana dashboards, I was deploying all the default ones plus the kubevirt one control plane dashboard. So, we might need some extra configuration besides adding a new line in `post-project-infra-prometheus-stack-deployment`

fgimenez: Sure, you should add entries for the new cluster here https://github.com/kubevirt/project-infra/tree/main/github/ci/services/prometheus-stack for instance, we have different manifests for each of the environments and a common set of them here https://github.com/kubevirt/project-infra/tree/main/github/ci/services/prometheus-stack/manifests/stack 

You can extract the `grafana-dashboards-default` ConfigMap from https://github.com/kubevirt/project-infra/blob/main/github/ci/services/prometheus-stack/manifests/stack/production-control-plane/grafana.yaml#L180, leave the rest common to `production-control-plane` and `production-e2e-performance-workloads` and have different config maps with the actual dashboard contents for each of them

marceloamaral: Unfortunately we cannot use the https://github.com/kubevirt/project-infra/blob/main/github/ci/services/prometheus-stack/manifests/stack/common/kube-prometheus-stack.yaml

KubeVirt only exposes the metrics if prometheus operator is in the `monitoring` namespace, and the ServiceAccount is `prometheus-k8s`.....

So, we can change the current `kube-prometheus-stack.yaml` or create a new one for the performance cluster...

fgimenez: No worries, rules_gitops takes care of moving all the resources to the namespace defined in the build file https://github.com/kubevirt/project-infra/blob/main/github/ci/services/prometheus-stack/BUILD.bazel#L40

In any case of any conflict we can always refactor and leave only the manifests required for each environment, but the defined namespace should not be a problem.

marceloamaral:
@dhiller and @fgimenez

I updated the PR to use the rules_gitops bazel to deploy the prometheus operator instead of the rudder.

To make it easier to review, I break the changes down into multiple commits and update the description of this PR with the main changes needed.

It is ready to be revised again.

github/ci/services/prometheus-stack/manifests/stack/common/kube-prometheus-stack.yaml
```diff
@@ -795,130 +795,130 @@ spec:
     app: alertmanager
     alertmanager: prometheus-stack-kube-prom-alertmanager
   type: "ClusterIP"
+# ---
```
fgimenez, line 798: I think it would be better to remove this dead code

github/ci/services/prometheus-stack/manifests/stack/common/kube-prometheus-stack.yaml
```diff
@@ -795,130 +795,130 @@ spec:
...
-  clusterIP: None
-  ports:
-    - name: http-metrics
-      port: 10251
-      protocol: TCP
-      targetPort: 10251
-  selector:
-    component: kube-scheduler
-  type: ClusterIP
+# # Source: kube-prometheus-stack/templates/exporters/kube-scheduler/service.yaml
```
fgimenez, line 898: Same as above, better not keep this code.

github/ci/services/prometheus-stack/manifests/stack/common/kube-prometheus-stack.yaml
```diff
@@ -237,7 +237,7 @@ imagePullSecrets:
 apiVersion: v1
 kind: ServiceAccount
 metadata:
-  name: prometheus-stack-kube-prom-prometheus
+  name: prometheus-k8s
```
fgimenez, line 240: This can be disruptive for the existing deployed stacks, can't we keep it as is for them and only apply the new name in the performance cluster?

github/ci/services/prometheus-stack/hack/deploy.sh
```diff
@@ -11,19 +11,25 @@ main(){
...
+    if [ "${environment}" = "production-control-plane" ] || [ "${environment}" = "testing" ] || [ "${environment}" = "production-performance-workloads" ]; then
         bazelisk run //github/ci/services/common/k8s/cmd/wait -- -namespace monitoring -selector grafana -kind deployment
     fi
 
     bazelisk run //github/ci/services/common/k8s/cmd/wait -- -namespace monitoring -selector node-exporter-prometheus-node-exporter -kind daemonset
 
     bazelisk run //github/ci/services/prometheus-stack:${environment}-service-monitors.apply
 
-    bazelisk run //github/ci/services/prometheus-stack:${environment}-rules.apply
+    if [ "${environment}" = "production-control-plane" ] || [ "${environment}" = "testing" ] || [ "${environment}" = "production-e2e-workloads" ]; then
```
fgimenez, line 28: Maybe better check if `${environment}` is not `production-performance-workloads`?

github/ci/services/prometheus-stack/hack/deploy.sh
```diff
@@ -11,19 +11,25 @@ main(){
...
+        bazelisk run //github/ci/services/prometheus-stack:${environment}-common-rules.apply
+    fi
+
+    bazelisk run //github/ci/services/prometheus-stack:${environment}-custom-rules.apply
```
fgimenez, line 32: Maybe the naming is a bit misleading, common rules are not always applied and custom rules are always applied, shouldn't it be the other way around?

fgimenez (commented):
Nice find about the kube-system services! A few minor comments, main concern is the potential changes to the deployed stacks because of the service account rename.

github/ci/services/prometheus-stack/manifests/stack/common/kube-prometheus-stack.yaml
marceloamaral: Agree!

github/ci/services/prometheus-stack/manifests/stack/common/kube-prometheus-stack.yaml
marceloamaral: I will remove them!

github/ci/services/prometheus-stack/manifests/stack/common/kube-prometheus-stack.yaml
marceloamaral: I am using patches to replace the service account name only in the performance cluster

github/ci/services/prometheus-stack/hack/deploy.sh
marceloamaral: done!

github/ci/services/prometheus-stack/hack/deploy.sh
marceloamaral: I changed the logic, now I am using patches to change the prometheus rules in the performance cluster
So, there is no need for applying different rules

marceloamaral:
@fgimenez I updated accordingly to all your comments, so it is ready to review again!

github/ci/services/prometheus-stack/patches/stack/production-performance-workloads/kube-prometheus-stack.yaml
```diff
@@ -0,0 +1,114 @@
+---
```
fgimenez, line 1: I think that for the patch not all the fields are required, but should be ok too if they are present

github/ci/services/prometheus-stack/patches/stack/production-control-plane/prometheus.yaml
```diff
@@ -14,3 +14,12 @@ spec:
     limits:
       cpu: 1
       memory: 12Gi
+  additionalScrapeConfigs:
```
fgimenez, line 17: Where does this change come from, aren't we just moving the file?

github/ci/services/prometheus-stack/patches/stack/production-e2e-workloads/prometheus.yaml
```diff
@@ -26,6 +26,15 @@ spec:
           matchLabels:
             app: prometheus
             type: local
+  additionalScrapeConfigs:
```
fgimenez, line 29: Same as above, where does this `additionalScrapeConfigs` come from?

fgimenez (commented):
Looks very good, only concern is the change in thanos config, other than that all good

github/ci/services/prometheus-stack/patches/stack/production-control-plane/prometheus.yaml
marceloamaral: This additional scrap config was in the original kube-prometheus-stack.yaml files
https://github.com/kubevirt/project-infra/blob/main/github/ci/services/prometheus-stack/manifests/stack/common/kube-prometheus-stack.yaml#L1239

Since we do not need it in the performance cluster, I moved this to patch in the other clusters

github/ci/services/prometheus-stack/patches/stack/production-e2e-workloads/prometheus.yaml
marceloamaral: As I commented before, it was in the original kube-prometheus-stack.yaml files

github/ci/services/prometheus-stack/patches/stack/production-performance-workloads/kube-prometheus-stack.yaml
marceloamaral: Humm, that is right, we don't need the fields that are not changing, but it might not be a problem to keep them

fgimenez (approved):
/approve

Thanks!

PR #1787, opened 2021-12-03
knmstate: Use golang image at postsubmit

qinqon:
The post submit job publish and release need to be fixed too. This
change use the golang image and the GIMMI env var and also align release
with publish job.


fgimenez (approved):
/approve

PR #1794, opened 2021-12-08
updated the kube proxy metrics bind address to use 0.0.0.0

marceloamaral:
The default configuration in kubespray when creating a k8s cluster does not expose kube proxy metrics.
But to scrape these metrics using prometheus, we need to change the `kube_proxy_metrics_bind_address` from using localhost address to 0.0.0.0 to bind all IPv4 interfaces.

/cc @fgimenez @dhiller 


fgimenez (approved):
/approve
/hold

Great catch, thanks! 

Maybe we could change the deployment job to be triggered on changes of files under `github/ci/prow-performance-workloads` instead of a tag push/changes in the deployment branch, wdyt? Can be done in a followup PR of course, feel free to unhold this one.

marceloamaral:
@fgimenez, I agree, we can change the trigger to deploy on changes of files under `github/ci/prow-performance-workloads`
And, yes, it might be better to be done in another PR
/unhold

PR #1802, opened 2021-12-12
k8s-1.22-ipv6, network: Add kubevirt presubmit job

oshoval:
Depends on https://github.com/kubevirt/kubevirtci/pull/721
Added with `always_run: false`, `optional: true` until we make kubevirt support single stack ipv6.

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
@@ -1603,6 +1603,45 @@ presubmits:
           privileged: true
       nodeSelector:
         type: bare-metal-external
+  - always_run: false
+    optional: true
+    annotations:
+      fork-per-release: "true"
+      testgrid-dashboards: kubevirt-presubmits
+    cluster: prow-workloads
+    decorate: true
+    decoration_config:
+      grace_period: 5m0s
+      timeout: 4h0m0s
+    labels:
+      preset-bazel-cache: "true"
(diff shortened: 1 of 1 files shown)


github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
```diff
@@ -1606,6 +1606,44 @@ presubmits:
           privileged: true
       nodeSelector:
         type: bare-metal-external
+  - always_run: false
```
qinqon, line 1609: Can you add `optional: true` ?

qinqon (changes requested):

github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
oshoval: sure, but aint it the default ?
i saw that we removed this all over and just using `always_run: false`

oshoval:
added `optional: true`

phoracek (commented):
/approve

qinqon:
Ahh right kubevirt haven't being bumped
/lgtm

oshoval:
Even when it is bumped, it was not tested with the new provider, there will be an async effort to adapt kubevirt to single ipv6.
The lane is not voting, developers will be able to use it manually, until it becomes official.

I will amend the PR, because once there is a failed rehearse it won't get merged.

oshoval:
rebase

PR #1812, opened 2021-12-16
Reduce the scope of cgroups v2 periodic job

vasiliy-ul:
The job is very unstable when all possible tests are executed. Focus only on sig-compute to make it less flaky.

https://prow.ci.kubevirt.io/?job=periodic-kubevirt-e2e-k8s-1.20-cgroupsv2

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml
@@ -1373,11 +1373,9 @@ periodics:
       - automation/test.sh
       env:
       - name: TARGET
-        value: k8s-1.20
+        value: k8s-1.20-sig-compute
       - name: KUBEVIRT_CGROUPV2
         value: "true"
-      - name: KUBEVIRT_E2E_SKIP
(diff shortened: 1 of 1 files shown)


vasiliy-ul:
@rmohr , @fgimenez what do you think about the PR? Currently the job fails randomly and unfortunately that makes it pretty useless.

fgimenez:
> @rmohr , @fgimenez what do you think about the PR? Currently the job fails randomly and unfortunately that makes it pretty useless.

makes sense to me, we have had 100% failures for a very long time https://testgrid.k8s.io/kubevirt-periodics#periodic-kubevirt-e2e-k8s-1.20-cgroupsv2&width=5

rmohr:
/lgtm
/approve

Definitely, makes a lot of sense!

/hold

Feel free to unhold when happy with it. If this succeeds, should we make it a presubmit to ensure that we can build on the current support?

vasiliy-ul:
> If this succeeds, should we make it a presubmit to ensure that we can build on the current support?

I think yes, makes sense. Maybe lets see how it goes with this periodic and if the results turn out to be more stable then I can raise a PR to setup the presubmit.

> Feel free to unhold when happy with it.

Should be fine to give it a try.

/unhold

PR #1814, opened 2021-12-17
Always run cgroups v2 presubmit job

vasiliy-ul:
The PR makes cgroups v2 presubmit job mandatory instead of runnig the periodic job.

With the test scope reduced to sig-compute only the results look more stable now: https://prow.ci.kubevirt.io/?job=periodic-kubevirt-e2e-k8s-1.20-cgroupsv2

@rmohr , @fgimenez FYI. We can monitor the periodic a bit more if there are doubts about it. But I think it should be fine to have  a mandatory presubmit.


rmohr:
/lgtm
/approve

great to see the progress secured.

PR #1817, opened 2021-12-21
k8s-1.22-ipv6: Make the lane mandatory

oshoval:
Since the lane is already passing with it's current configuration,
and it is part of the kubevirtci package,
it should be marked as mandatory.


oshoval:
https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_project-infra/1817/rehearsal-check-provision-k8s-1.22-ipv6/1473205672781811712

passed

PR #1822, opened 2022-01-04
Enable reporting for sig-monitoring

erkanerol:
After observing the lane for a while and ensuring it works, we can enable reporting now.
It will be still optional but developers will be warned about the issues. After a while, it can
be mandatory as well.

https://prow.ci.kubevirt.io/?job=pull-kubevirt-e2e-k8s-1.22-sig-monitoring


dhiller (approved):
/approve

Thanks!

PR #1824, opened 2022-01-04
Whitespace change to trigger rebuild of bootstrap image

maya-r:
This updates moby-engine and picks up moby/moby#42681 which fixes
pthread_create for glibc 2.34 or newer.

### DIFF
--- a/images/bootstrap/Dockerfile
+++ b/images/bootstrap/Dockerfile
@@ -70,7 +70,6 @@ RUN wget -q https://dl.google.com/dl/cloudsdk/channels/rapid/google-cloud-sdk.ta
     gcloud components install alpha beta kubectl && \
     gcloud info | tee /workspace/gcloud-info.txt
 
-
 #
 # BEGIN: DOCKER IN DOCKER SETUP
 #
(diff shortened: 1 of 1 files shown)


maya-r:
cc @fgimenez @dhiller

fgimenez (commented):
The change is ok but I think it would be better if we could have reproducible builds, doesn't sound good that two consecutive builds, just changing whitespace, can lead to completely different images. Do you think we could use a more specific tag or even a digest instead of the generic `34` tag in the `FROM` entry? and maybe a comment about the underlying fedora version being used. Would be good to hear what others think about this.

It could be also nice to use an image from quay instead of docker hub, but not sure if fedora is available there. In any case, not related to this PR.

maya-r:
if our method of fetching packages is `RUN dnf...` it is dependent on what's in the dnf repository at a given time, so not reproducible. Would have to use a different method of creating a container, like bazel.

fgimenez (commented):
/lgtm

> if our method of fetching packages is RUN dnf... it is dependent on what's in the dnf repository at a given time, so not reproducible. Would have to use a different method of creating a container, like bazel.

makes sense, in any case maybe worth discussing this on a separate PR, not blocking this one to fix the underlying issue asap

dhiller (commented):
/approve

PR #1829, opened 2022-01-05
Update bootstrap image used for CDI repo

maya-r:
We seem to use multiple versions and I don't keep a close enough eye on all the different ways it could break, so limiting the change to CDI only.

This is needed to pick up updated docker, which allows us to run centos stream9.

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/containerized-data-importer/containerized-data-importer-periodics.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/containerized-data-importer/containerized-data-importer-periodics.yaml
@@ -127,7 +127,7 @@ periodics:
     nodeSelector:
       type: bare-metal-external
     containers:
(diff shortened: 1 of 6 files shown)


maya-r:
cc @awels @mhenriks

awels:
/lgtm
@mhenriks would this affect your PR in CDI?

maya-r:
it should be a no-op for CDI :)

awels:
Well we are releasing fedora based images with CDI, and if the builde makes centos binaries, we might have some issues.

maya-r:
This is the image we run `make test-functional` on, but it is a no-op because we run most of our things in the builder image (and also, because the bootstrap image is unchanged except for a newer docker and maybe a few other packages)

awels:
Oh this is the image used in prow to start our builder. Okay.

maya-r:
Not needed after #1616

PR #1832, opened 2022-01-08
Kubevirt prow namespace to monitoring

marceloamaral:
After updating the prometheus stack to be deployed to the performance cluster, the prow-workloads cluster started getting alerts about some service monitors.

Although the [original code](https://github.com/kubevirt/project-infra/pull/1778/files#diff-8ac0dadb1c90800200076b5a36d0b123da3490b0ec96c52574119230d91e0f40L63) has been configured to deploy the service monitors in the `kubevirt-prow` namespace. `kubevirt-prow` prow cluster actually has the service monitors deployed in the `monitoring` namespace.
image
So this PR is changing all service monitor deployments to be in the `monitoring` namespace.

Also, the service monitors to extract metrics from `controller-manager`, kube-proxy` and `kube-scheduler` are not healthy. Even the old ones in the `monitoring` namespace and the new ones deployed in the `kubevirt-prow` namespace.
image
As far as I know, no one is looking at these metrics, so I've commented out these service monitors.
In fact, the kube proxy is not being deployed to expose metrics on the clusters, except on the performance cluster.

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml
@@ -1012,6 +1012,120 @@ periodics:
 - annotations:
     testgrid-dashboards: kubevirt-periodics
     testgrid-days-of-results: "60"
+  cluster: ibm-prow-jobs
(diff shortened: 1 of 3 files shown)


marceloamaral:
I am closing this PR because for some reason it has an old commit of another PR that has already been merged and I could not fix it

PR #1833, opened 2022-01-08
Move prometheus stack service monitors from kubevirt-prow to monitoring namespace

marceloamaral:
After updating the prometheus stack to be deployed to the performance cluster, the prow-workloads cluster started getting alerts about some service monitors.

Although the [original code](https://github.com/kubevirt/project-infra/pull/1778/files#diff-8ac0dadb1c90800200076b5a36d0b123da3490b0ec96c52574119230d91e0f40L63) has been configured to deploy the service monitors in the `kubevirt-prow` namespace. `kubevirt-prow` prow cluster actually has the service monitors deployed in the `monitoring` namespace.
image
So this PR is changing all service monitor deployments to be in the `monitoring` namespace.

Also, the service monitors to extract metrics from `controller-manager`, kube-proxy` and `kube-scheduler` are not healthy. Even the old ones in the `monitoring` namespace and the new ones deployed in the `kubevirt-prow` namespace.
image
As far as I know, no one is looking at these metrics, so I've commented out these service monitors.
In fact, the kube proxy is not being deployed to expose metrics on the clusters, except on the performance cluster.

/cc @fgimenez @dhiller 

### DIFF
--- a/github/ci/services/prometheus-stack/manifests/service-monitors/common/servicemonitors.yaml
+++ b/github/ci/services/prometheus-stack/manifests/service-monitors/common/servicemonitors.yaml
@@ -49,31 +49,6 @@ spec:
   - port: http-metrics
     bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
 ---
-# Source: kube-prometheus-stack/templates/exporters/kube-controller-manager/servicemonitor.yaml
-apiVersion: monitoring.coreos.com/v1
-kind: ServiceMonitor
-metadata:
-  name: prometheus-stack-kube-prom-kube-controller-manager
-  namespace: default
-  labels:
-    app: kube-prometheus-stack-kube-controller-manager
-    chart: kube-prometheus-stack-13.6.0
-    release: "prometheus-stack"
-    heritage: "Helm"
-    group: kubevirtci
-spec:
-  jobLabel: jobLabel
-  selector:
-    matchLabels:
-      app: kube-prometheus-stack-kube-controller-manager
-      release: "prometheus-stack"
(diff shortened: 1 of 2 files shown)


dhiller (approved):
Thanks @marceloamaral, please ping me if you need anything else!

github/ci/services/prometheus-stack/manifests/service-monitors/common/servicemonitors.yaml
```diff
@@ -48,31 +48,31 @@ spec:
...
-    matchLabels:
-      app: kube-prometheus-stack-kube-controller-manager
-      release: "prometheus-stack"
-  namespaceSelector:
-    matchNames:
-      - kube-system
-  endpoints:
-  - port: http-metrics
-    bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
+# ---
```
fgimenez, line 51: I think it would be better to remove the service monitors not being used instead of keeping dead code, wdyt?

fgimenez (commented):
Looks very good thanks! Only a small suggestion about removing commented code, other than that looks good

github/ci/services/prometheus-stack/manifests/service-monitors/common/servicemonitors.yaml
marceloamaral: Removed! Can you approve and lgtm?

dhiller (approved):
/approve
/hold

Waiting for @fgimenez to verify also

fgimenez (approved):
/approve
/unhold

awesome thanks!

PR #1834, opened 2022-01-10
kind-1.19-sriov: Disable job autorun

oshoval:
We now have kind-1.22-sriov which is the operational one.


oshoval:
@ormergi @qinqon please take a look

PR #1846, opened 2022-01-13
Remove CDI devmapper link in jobs

awels:
We no longer use the devmapper ln and it is causing issues with
the new builder image. Removing them from the pre-submits.

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/containerized-data-importer/containerized-data-importer-presubmits.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/containerized-data-importer/containerized-data-importer-presubmits.yaml
@@ -1,38 +1,5 @@
 presubmits:
   kubevirt/containerized-data-importer:
-  - name: pull-containerized-data-importer-e2e-k8s-1.17-ember-lvm
-    skip_branches:
-      - release-v1.13
-    annotations:
-      fork-per-release: "true"
-    always_run: false
-    optional: true
-    decorate: true
-    decoration_config:
-      timeout: 5h
-      grace_period: 5m
-    max_concurrency: 6
-    cluster: prow-workloads
-    labels:
-      preset-dind-enabled: "true"
-      preset-docker-mirror-proxy: "true"
-      preset-shared-images: "true"
-    spec:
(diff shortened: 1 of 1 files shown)


dhiller (commented):
/approve

akalenyu:
so https://github.com/kubevirt/project-infra/pull/573 was the reason we introduced this, but that is no longer needed?

dhiller:
/hold

Just in case so that it doesn't get merged before we are sure everything works

awels:
So the devmapper thing was needed for improved registry import, but the devmapper was for a format we don't support, and it was included due to the way we imported it. That got fixed and we don't need it anymore AFAIK

akalenyu:
> So the devmapper thing was needed for improved registry import, but the devmapper was for a format we don't support, and it was included due to the way we imported it. That got fixed and we don't need it anymore AFAIK

gotcha, you're right:
https://github.com/kubevirt/containerized-data-importer/pull/1375

maya-r:
Sounds like we can cancel the hold, what do you think?

dhiller (approved):

dhiller:
> Sounds like we can cancel the hold, what do you think?

Remove hold whenever you see fit, to me it looks like the tests are running fine. Regarding the test failures in the lane I can't tell.

awels:
/hold cancel
I am not worried about the failure. I wasn't expecting it to run the entire lane but I guess it is a good test. The failure is a known flake.

dhiller:
/override error
rehearsal-pull-containerized-data-importer-e2e-k8s-1.22-hpp-destructive

dhiller:
/override rehearsal-pull-containerized-data-importer-e2e-k8s-1.22-hpp-destructive

dhiller:
@awels if you need that update quickly, I can stop the rehearsals and retire the contexts.

awels:
Yeah go ahead and do that, I see several of them actually passed without issues. So I am confident this will work.

dhiller:
/override rehearsal-pull-containerized-data-importer-e2e-k8s-1.21-ceph
/override rehearsal-pull-containerized-data-importer-e2e-k8s-1.21-hpp

PR #1852, opened 2022-01-14
Increase alert time for high context switching to reduce noise from short spikes

brianmcarey:
Short spikes are causing lots of alerts to fire for high context switching. 

These alerts are not causing any issues for the e2e jobs so increasing the time to reduce noise. 

/cc @dhiller @fgimenez 


dhiller (approved):
/approve

PR #1872, opened 2022-01-21
Add presets for github credentials to project infra 

brianmcarey:
Add presets for github credentials to project infra presubmit

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/project-infra/project-infra-periodics.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/project-infra/project-infra-periodics.yaml
@@ -4,8 +4,6 @@ periodics:
   decorate: true
   annotations:
     testgrid-create-test-group: "false"
-  labels:
-    preset-github-credentials: "true"
   cluster: ibm-prow-jobs
   spec:
     containers:
@@ -37,6 +35,13 @@ periodics:
       - --template
       - --ceiling=1
       - --confirm
+      volumeMounts:
+      - name: token
+        mountPath: /etc/github
+    volumes:
+    - name: token
+      secret:
+        secretName:  commenter-oauth-token
 - name: periodic-project-infra-close
   interval: 1h
   annotations:
--- a/github/ci/prow-deploy/files/jobs/kubevirt/project-infra/project-infra-periodics.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/project-infra/project-infra-periodics.yaml
@@ -4,6 +4,8 @@ periodics:
   decorate: true
   annotations:
     testgrid-create-test-group: "false"
+  labels:
+    preset-github-credentials: "true"
   cluster: ibm-prow-jobs
   spec:
     containers:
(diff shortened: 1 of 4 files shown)


github/ci/prow-deploy/files/jobs/kubevirt/project-infra/project-infra-periodics.yaml
```diff
@@ -35,17 +37,12 @@ periodics:
       - --template
       - --ceiling=1
       - --confirm
-      volumeMounts:
-      - name: token
-        mountPath: /etc/github
-    volumes:
-    - name: token
-      secret:
-        secretName:  commenter-oauth-token
```
dhiller, line 44: I think that this is a different token than what we are using everywhere else. Can you check please? If that is the case, you might create a separate preset (i.e. `commenter-github-token`)

github/ci/prow-deploy/files/jobs/kubevirt/project-infra/project-infra-presubmits.yaml
```diff
@@ -375,9 +372,6 @@ presubmits:
       - name: pgp-bot-key
         secret:
           secretName: pgp-bot-key
-      - name: unprivileged-token
-        secret:
-          secretName: unprivileged-oauth-token
```
dhiller, line 380: I think we should not touch the job config for this job, it actually uses the unprivileged token to not cause any interference when testing the prow instance within the kind cluster IIRC.
Especially not fetching GitHub events for the real prow instance (which is what happened when we first tried this accidentally with the same token as the production prow uses) and also not messing with PR status then.

github/ci/prow-deploy/kustom/base/configs/current/config/config.yaml
```diff
@@ -538,6 +538,20 @@ presets:
...
+    preset-github-credentials: "true"
+  env:
+  - name: GIT_ASKPASS
+    value: ../project-infra/hack/git-askpass.sh
+  - name: GITHUB_TOKEN
+    value: "/etc/github/oauth"
+  volumes:
+  - name: github-token
+    secret:
+      secretName: oauth-token
```
dhiller, line 551: Maybe we can set user readonly, but not sure whether that would work for every job?

```suggestion
      defaultMode: 0400
      secretName: oauth-token
```

dhiller (commented):
Hey, looks very good already! Some comments inline.

github/ci/prow-deploy/files/jobs/kubevirt/project-infra/project-infra-periodics.yaml
brianmcarey: yeah I will double check this.

brianmcarey: The tokens were different so I added commenter-oauth-token and unprivileged token back in latest update.

github/ci/prow-deploy/kustom/base/configs/current/config/config.yaml
brianmcarey: Tested a couple of jobs and they succeeded with the above suggested permissions.

dhiller (approved):
/approve

PR #1875, opened 2022-01-25
Add HPP ceph lane

akalenyu:
Add a lane that bases the HPP storage pool on ceph PVC
Optional at first so we can make necessary adjustments in the hostpath-provisioner repo.

Signed-off-by: Alex Kalenyuk <akalenyu@redhat.com>


dhiller (approved):
/approve

PR #1885, opened 2022-02-01
Add a GOPROXY fallback if the main proxy rejects us

rmohr:
Add a second proxy and ensure that go will download from the next
specified source in case of any failure.

The old value was `https://proxy.golang.org,direct`, the new value is `https://proxy.golang.org|https://goproxy.io|direct`.

This fixes an issue where the jobs sometimes gets a GOAWAY which leads
to an immediate error, since direct download is only considered if a 404
or a 410 error code is returned.

Note that there is a difference between `,` and `|` separators. If `,` is used, only in the mentioned cases above retries are happening. Details are here: https://go.dev/doc/go1.15#go-command


dhiller (approved):
/approve

PR #1898, opened 2022-02-04
Create an optional Performance presubmit

rthallisey:
Signed-off-by: Ryan Hallisey <rhallisey@nvidia.com>

The [performance periodic job](https://testgrid.k8s.io/kubevirt-periodics#periodic-kubevirt-e2e-k8s-1.21-sig-performance&width=20) has been running for a few months now.  The job is stable and it can provide PR owners with valuable performance data about their change.

Soon sig-scale will be adding performance and scale thresholds to the job https://github.com/kubevirt/kubevirt/pull/7180 which will cause it to fail when a PR doesn't meet perf-scale requirements.


rthallisey:
@dhiller can you review?

marceloamaral:
This is interesting. Is it possible to also include a label that will trigger the job?
For example adding the label `needs-performance-test` in a PR.

rthallisey:
Not sure how to do that with a label.  You can run `/test  pull-kubevirt-e2e-k8s-1.21-sig-performance` to trigger the test.

github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
```diff
@@ -388,6 +388,66 @@ presubmits:
...
         type: bare-metal-external
+  - always_run: false
+    annotations:
+      fork-per-release: "true"
+      testgrid-dashboards: kubevirt-presubmits
+    cluster: prow-workloads
+    decorate: true
+    decoration_config:
+      grace_period: 5m0s
+      timeout: 4h0m0s
```
davidvossel, line 399: minor, but we can reduce this to probably 1 hour. we don't have to change this right now though. it's not really important.

github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
```diff
@@ -388,6 +388,66 @@ presubmits:
...
+    labels:
+      preset-bazel-cache: "true"
+      preset-bazel-unnested: "true"
+      preset-dind-enabled: "true"
+      preset-docker-mirror-proxy: "true"
+      preset-shared-images: "true"
+    max_concurrency: 11
```
davidvossel, line 406: since this doesn't run by default, this likely won't cause an issue. but given how expensive this test is, we likely want to limit the concurrency to something like 2

davidvossel (commented):
/approve

@rthallisey can you reduce the concurrency to 2 and then i'll LGTM this. That should give us some guarantees that this test won't have the opportunity to overwhelm our infra if lots of people request it.

dhiller (approved):

PR #1906, opened 2022-02-09
Add nonroot equivalent to migration lane

xpivarc:


xpivarc:
I will switch this to required once I can see at least one successful run on main.

dhiller:
See https://github.com/kubernetes/test-infra/pull/25175

/override pull-project-infra-check-testgrid-config

dhiller (approved):
/approve

PR #1919, opened 2022-02-16
Add manifests to deploy loki and promtail to kubevirt-ci clusters

brianmcarey:
Loki and Promtail are required to enable monitoring of pod logs in kubevirt-ci clusters.

These manifests are based on the loki and promtail helm charts.

Loki will provide better monitoring of pod logs from the control plane cluster and allow us to see when errors occur in control services such as the status reconciler.

/cc @dhiller 


github/ci/services/loki/manifests/common/loki.yaml
```diff
@@ -0,0 +1,230 @@
...
+        {}
+      affinity:
+        {}
+      tolerations:
+        []
+      terminationGracePeriodSeconds: 4800
+      volumes:
+        - name: config
+          secret:
+            secretName: loki-config
```
dhiller, line 228: So IIUC the target configuration is to be added to our secrets?

dhiller (commented):
Looks great already! One question inline.

github/ci/services/loki/manifests/common/loki.yaml
brianmcarey: That is correct - the helm chart passes the loki configuration as a secret. I am looking at the configuration reuired for the production-control-plane and if there is no sensitive information it can be added here - otherwise it would have to be added to our secrets.

github/ci/services/loki/secrets/testing/loki-config/loki.yaml
```diff
@@ -0,0 +1,44 @@
...
+  chunk_retain_period: 1m
+  lifecycler:
+    ring:
+      kvstore:
+        store: inmemory
+      replication_factor: 1
+  max_transfer_retries: 0
+  wal:
+    dir: /data/loki/wal
+limits_config:
```
dhiller, line 19: We haven't yet configured retention here, right? Are there defaults or would we be storing everything forever with this configuration? IIUC the [docs](https://grafana.com/docs/loki/latest/operations/storage/retention/#retention-configuration) correctly you explicitly need to activate this?

dhiller (commented):
Looks very good already! One question regarding retention inline.

github/ci/services/loki/secrets/testing/loki-config/loki.yaml
brianmcarey: Good point Daniel, I have made an update to the configuration to enable retention and to retain the logs for 2 weeks.

dhiller: Great, thanks!

dhiller (approved):
/approve

Great work, thanks! Holding so you can decide when you see fit it going in.

/hold

PR #1921, opened 2022-02-17
Sync version Istioctl version 1.13

rhrazdil:
Substitute 1.10.1 with 1.13, as we're not using 1.10.1 anywhere anyway.


dhiller (approved):
/approve

PR #1927, opened 2022-02-23
knmstate: Call publish only on main

qinqon:
The publish job push to quay.io latest but latest has to point only to code at
main branch. This change disable publish at release branches.


rhrazdil (commented):
/lgtm
/approve

PR #1936, opened 2022-03-01
Increase resource requests and limits for the Loki service

brianmcarey:
The default resource limits were too small for a large cluster which
causes the loki pod to restart

Increasing the resource limits to match what is used for prometheus

/cc @dhiller 


dhiller (approved):
/approve

PR #1942, opened 2022-03-03
[kubevirt-presubmits]: update the sig-monitoring job configuration

enp0s3:
remove the optional flag so the lane will always run upon
monitoring code change. Also make the report visible since
not all the contributors are familiar with the Prow dashboard.

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
@@ -2042,11 +2042,9 @@ presubmits:
       preset-dind-enabled: "true"
       preset-docker-mirror-proxy: "true"
     name: pull-kubevirt-e2e-k8s-1.22-sig-monitoring
-    optional: true
     run_if_changed: ^pkg/monitoring/.*|tests/monitoring/.*$
     skip_branches:
     - release-\d+\.\d+
-    skip_report: true
     spec:
       containers:
       - command:


go.mod
```diff
@@ -17,9 +17,7 @@ require (
 	github.com/onsi/ginkgo v1.16.4
 	github.com/onsi/gomega v1.13.0
 	github.com/prometheus/client_golang v1.11.0
-	github.com/shurcooL/githubv4 v0.0.0-20210725200734-83ba7b4c9228 // indirect
```
dhiller, line 20: This looks like an unrelated change?

github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits-0.49.yaml
```diff
@@ -1867,7 +1867,6 @@ presubmits:
       preset-dind-enabled: "true"
       preset-docker-mirror-proxy: "true"
     name: pull-kubevirt-e2e-k8s-1.22-sig-monitoring-0.49
-    optional: true
```
dhiller, line 1870: Do we really want to do this for the 0.49 lane, not sure we have enough data for this? Or is this required as we are going to backport some stuff here?

go.mod
enp0s3: @dhiller Correct. Fixed

github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits-0.49.yaml
enp0s3: @dhiller Currently we don't have any PR related to monitoring that has to be backported. I was only thinking about a possibility for that to happen. I think that in general if that will happen and we will keep it optional in 0.49, it will be tested on main but not tested on the 0.49 which is inconsistent, WDYT?

dhiller: If you remove the `optional: true` it will still be tested but the merge will not be gated by tide should the lane fail.

enp0s3: @dhiller I see, and since the report will appear, the reviewer will notice the failure of the lane. So lets keep optional then. Thank you for the brainstorm!

dhiller (approved):
/approve

PR #1944, opened 2022-03-04
Add new secret to allow prow pods to start

brianmcarey:
This secret has been added to the deployment specs for prow services
such as sinker as part of the prow bump. The pods remain in pending state when this secret is not
in place.

See #1768

/cc @dhiller


brianmcarey:
@dhiller 
pull-project-infra-prow-deploy-test is passing now with the bump-prow changes and the added secret - this should unblock the bump-prow job now. Can you please take a look?

dhiller (approved):
/approve

Great work, thank you!

PR #1563, opened 2021-09-09
Release configs for kubevirt/kubevirt release branch release-0.52

davidvossel:
adds new release configs

PR #1954, opened 2022-03-10
Update kubevirt-presubmits-0.49 config to 1.23

fossedihelm:
Align the release-0.49 config with the main one.
- Update 1.20 lanes with 1.23.
- Update kind-1.19-vgpu lane with kind-1.23-vgpu
- Adding *-sig-compute-migrations lanes

These are necessary to backport the update of poddisruption budget api from v1beta1 to v1

Signed-off-by: fossedihelm <ffossemo@redhat.com>

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits-0.49.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits-0.49.yaml
@@ -29,7 +29,7 @@ presubmits:
         env:
         - name: TARGET
           value: k8s-1.21-sig-network
-        image: quay.io/kubevirtci/bootstrap:v20210906-994b913
+        image: quay.io/kubevirtci/bootstrap:v20220211-d7d6c59
         name: ""
         resources:
           requests:
@@ -65,7 +65,7 @@ presubmits:
         env:
         - name: TARGET
           value: k8s-1.21-sig-storage
-        image: quay.io/kubevirtci/bootstrap:v20210906-994b913
+        image: quay.io/kubevirtci/bootstrap:v20220211-d7d6c59
         name: ""
         resources:
           requests:
@@ -101,7 +101,7 @@ presubmits:
         env:
         - name: TARGET
           value: k8s-1.21-sig-compute
-        image: quay.io/kubevirtci/bootstrap:v20210906-994b913
+        image: quay.io/kubevirtci/bootstrap:v20220211-d7d6c59
         name: ""
         resources:
           requests:
@@ -137,7 +137,7 @@ presubmits:
         env:
         - name: TARGET
           value: k8s-1.21-sig-operator
-        image: quay.io/kubevirtci/bootstrap:v20210906-994b913
+        image: quay.io/kubevirtci/bootstrap:v20220211-d7d6c59
         name: ""
         resources:
           requests:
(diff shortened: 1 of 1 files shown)


fossedihelm:
@dhiller When you have some free time, can you have a look on this? Thanks

dhiller:
I didn't yet take a deeper look, just a quick question upfront. So we do want to guarantee compatibility with 1.23 for 0.49, as is generally stated here, right?
https://github.com/kubevirt/kubevirt/blob/main/docs/kubernetes-compatibility.md#kubernetes-version-compatibility

I was just wondering, because I don't remember us doing this before.

dhiller (approved):
/approve
/hold 

One other question: shouldn't we do this for the releases > 0.49 also?

fossedihelm:
> I didn't yet take a deeper look, just a quick question upfront. So we do want to guarantee compatibility with 1.23 for 0.49, as is generally stated here, right? https://github.com/kubevirt/kubevirt/blob/main/docs/kubernetes-compatibility.md#kubernetes-version-compatibility
> 
> I was just wondering, because I don't remember us doing this before.

The main reason of the upgrade is basically that OCP 4.10 is based on k8s 1.23. OCP 4.12 release could be based on k8s 1.25 in which poddisruptionbudget api version v1beta1 is gonna be deleted. So we should backport [https://github.com/kubevirt/kubevirt/pull/6955](https://github.com/kubevirt/kubevirt/pull/6955) otherwise the upgrade from 4.10 to 4.12 could be difficult. To backport the latter PR we need to  update the k8s dependencies and the lanes. Hope it is clear, otherwise don't hesitate to ask. Thanks

fossedihelm:
> One other question: shouldn't we do this for the releases > 0.49 also?

Generally yes, but it depends because there is the possibility that they are still present. This backport happens because the release0.49 job has been created before the update of k8s dependencies occurs. Also, I cannot see any presubmit job for release >0.49: only an open PR for 0.51

dhiller:
> > I didn't yet take a deeper look, just a quick question upfront. So we do want to guarantee compatibility with 1.23 for 0.49, as is generally stated here, right? https://github.com/kubevirt/kubevirt/blob/main/docs/kubernetes-compatibility.md#kubernetes-version-compatibility
> > 
> > I was just wondering, because I don't remember us doing this before.

> The main reason of the upgrade is basically that OCP 4.10 is based on k8s 1.23. OCP 4.12 release could be based on k8s 1.25 in which poddisruptionbudget api version v1beta1 is gonna be deleted. So we should backport [https://github.com/kubevirt/kubevirt/pull/6955](https://github.com/kubevirt/kubevirt/pull/6955) otherwise the upgrade from 4.10 to 4.12 could be difficult. To backport the latter PR we need to  update the k8s dependencies and the lanes. Hope it is clear, otherwise don't hesitate to ask. Thanks

Thanks for the explanation. Makes absolute sense then.

> > One other question: shouldn't we do this for the releases > 0.49 also?
> 
> Generally yes, but it depends because there is the possibility that they are still present. This backport happens because the release0.49 job has been created before the update of k8s dependencies occurs. Also, I cannot see any presubmit job for release >0.49: only an open PR for 0.51

You are right. I think we need to fork the configs manually as that seemingly didn't happen since 0.50: https://github.com/kubevirt/project-infra/pull/1563#issuecomment-1066535898

/unhold

I don't see any other reason to block on this. Thanks again!

dhiller:
Created kubevirt/project-infra#1964 to create the missing configs for 0.50 and 0.51

PR #1955, opened 2022-03-14
Ensure that vgpu jobs are not scheduled to the same nodes

rmohr:
We reconfigure the whole GPU when executing vgpu jobs. During that time
no other vgpu job should get scheduled to the node, so that they are not
in the way of each other.

Add a pod anti affinity which ensures that pods labeled with `vgpu:
true` are not scheduled on the same nodes at the same time.


rmohr:
@vladikr herer finally the PR which we talked about some time ago. I hope it will further reduce the vgpu lane failure rate.

brybacki:
I have no experience with configuring anti-affinity, but it looks good to me.

brianmcarey (commented):
/lgtm

rmohr:
thanks!

PR #1976, opened 2022-03-23
Switch retest-bot to use /retest-required

maya-r:
Use less CI resources at no impact.


brianmcarey (commented):
This makes sense to me - @dhiller what do you think?

/lgtm

dhiller (approved):
Makes sense, thanks!

/approve

PR #1978, opened 2022-03-24
Upgrade Grafana to version 8.4

brianmcarey:
Grafana version 8.4 provides an improved view of alerts from both
prometheus and loki

/cc @dhiller 


dhiller (approved):
/approve

PR #1980, opened 2022-03-24
Use recreate strategy for Grafana deployment update

brianmcarey:
Updating the grafana deployment using the RollingUpdate strategy fails due to being unable to claim a mount from existing pod

The [Recreate strategy](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#recreate-deployment) will remove the old pod before creating the updated pod.

/cc @dhiller 


dhiller:
The only thing we need to keep in mind that if we are updating grafana we need to announce a service disruption then.

dhiller (approved):
/approve

brianmcarey:
Updating PVC access modes is not allowed - the PVC would have to be recreated. The recreate update strategy will have less impact as the grafana deployment does not get updated that often. 

/remove-hold

PR #1983, opened 2022-03-24
Make nonroot sriov lane required

xpivarc:


github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
```diff
@@ -559,7 +559,7 @@ presubmits:
       sriov-pod: "true"
     max_concurrency: 10
     name: pull-kubevirt-e2e-kind-1.22-sriov-nonroot
-    optional: true
+    optional: false
```
dhiller, line 562: nit: you can just remove the line, default is `false` for `optional`

dhiller (approved):
/approve

/override pull-project-infra-check-testgrid-config

Unrelated test failure.

github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
xpivarc: I wanted to be explicit because I always forget what the default is.

xpivarc:
@dhiller Can you re-apply approve, please? I think there was downtime :)

EdDev:
@dhiller , we have been in a CI crisis mode for the last 1.5 days due to issues with the SR-IOV workload on prow.
How is it possible that now we increased the jobs count and added this one as required?
And at the same time #1994 was approved, which just collides with this one if I understand it correctly.

If I am missing something, please help me out here to understand.

xpivarc:
> @dhiller , we have been in a CI crisis mode for the last 1.5 days due to issues with the SR-IOV workload on prow. How is it possible that now we increased the jobs count and added this one as required? And at the same time #1994 was approved, which just collides with this one if I understand it correctly.
> 
> If I am missing something, please help me out here to understand.

I am not aware of increased load to CI other than spikes caused by Github issues. The reason for this job is simple we want to test the non-root variant. 

Thanks for pointing out #1994 I have commented to include this lane.

EdDev:
> I am not aware of increased load to CI other than spikes caused by Github issues.

There is an email thread about it: `[kubevirt-dev] Impact to SR-IOV test lanes`

> The reason for this job is simple we want to test the non-root variant.

I am actually not in the loop about the status of this non-root feature.
But I do know we do not have SR-IOV HW to accommodate for more workloads, which should be also true for the GPU ones.

xpivarc:
> > I am not aware of increased load to CI other than spikes caused by Github issues.
> 
> There is an email thread about it: `[kubevirt-dev] Impact to SR-IOV test lanes`

I was thinking in general. It is true that SR-IOV has some problems now.

> 
> > The reason for this job is simple we want to test the non-root variant.
> 
> I am actually not in the loop about the status of this non-root feature. But I do know we do not have SR-IOV HW to accommodate for more workloads, which should be also true for the GPU ones.

Let's discuss this and see what we can do.

PR #1984, opened 2022-03-24
Update nonroot vgpu lane to 1.23

xpivarc:
We don't support 1.19 anymore with the main branch.

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
@@ -861,7 +861,7 @@ presubmits:
             path: /dev/vfio/
             type: Directory
           name: vfio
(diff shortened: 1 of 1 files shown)


xpivarc:
@rmohr Can you have a look, please?

jean-edouard (commented):
This PR both upgrades the lane and makes it gating.
How do we know the upgraded lane will be reliable? If it's not, it will block all PRs...
Would it make sense to upgrade + run always, but wait a few days before switching to required?

xpivarc:
> This PR both upgrades the lane and makes it gating. How do we know the upgraded lane will be reliable? If it's not, it will block all PRs... Would it make sense to upgrade + run always, but wait a few days before switching to required?

Makes sense. I pushed the changes. I think it should be enough to have a few runs as this is almost identical to the existing lane. The difference should cause to fail it almost in all cases if any.

jean-edouard (approved):
/approve

PR #1992, opened 2022-03-28
Add single-node single-replica lane

jean-edouard:
This new lane doesn't automatically run and is optional, for now.
It doesn't specify a sig to just run everything.
It uses a single node with a single replica of virt-api and virt-controller, which is a particular setup not currently tested by any lane.
I requested "only" 15GiB of memory since there's only one node, but maybe that's wrong?


dhiller (approved):
/approve

PR #2000, opened 2022-03-30
Add packages to bootstrap (to build multi-arch containers)

maya-r:
CDI uses `docker build` to build a builder, and it uses this builder
to do other builds.

Having an aarch64 builder will allow us to do native aarch64 builds,
and for that we need a fancier `docker build` like buildah (or buildx).

qemu-user-static allows us to execute non-native (e.g. aarch64)
binaries.

Part of tackling https://github.com/kubevirt/containerized-data-importer/issues/2211

### DIFF
--- a/images/bootstrap/Dockerfile
+++ b/images/bootstrap/Dockerfile
@@ -52,6 +52,8 @@ RUN dnf install -y \
     rsync-daemon \
     skopeo \
     sudo \
+    buildah \
+    qemu-user-static \
     wget &&\
   dnf -y clean all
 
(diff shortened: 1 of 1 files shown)


maya-r:
cc @brianmcarey @dhiller

maya-r:
cc @awels @akalenyu - I'd like to switch CDI to build using buildah so our `make builder-push` produce both arm64 and amd64 images.

brianmcarey (commented):
/lgtm

akalenyu:
So kubevirt is producing arm builder using `docker`:
https://github.com/kubevirt/kubevirt/commit/d2037f9c395c12ccb39013774385135a0a318b23

Is that something we can do as well?

maya-r:
[This is what I plan to do](https://github.com/maya-r/containerized-data-importer/commit/8f07ed125a47baaf3b87ef1ca56843a03f4bf7d2) - it's very similar.

I don't think most people would notice the extra dependency requirement since rebuilding the builder is uncommon. Maybe I should check for qemu-user-static's existence, though, since the failure mode is surprising if it doesn't exist.

PR #1900, opened 2022-02-07
Make NonRoot network lane required

xpivarc:

PR #2011, opened 2022-04-03
Revert "Make nonroot network lane required (#1900)"

EdDev:
There are only 2 nodes that can run SR-IOV jobs, each at the moment
can only run a single job at a time.

Running the nonroot jobs for pre-submit is overloading what seems to be
a not yet stable CI workload.

This reverts commit eb4115da872dcdff53ad5d87bbb322990164e84f.


brianmcarey (commented):
/lgtm

PR #2014, opened 2022-04-04
periodics: Move all non sriov jobs to non sriov nodes

EdDev:
Although the presubmit jobs have been updated to not run on SR-IOV
nodes, the periodic ones have not.
Include them as well to reduce load on the SR-IOV nodes to attempt and
recover them from the latest CI outage.


brianmcarey (commented):
/lgtm

EdDev:
Controlling on which node to run should be done by adding or removing labels from the nodes.
The jobs themselves should only require the HW needed (through specific labels).

Closing this PR and looking forward to prepare a procedure on how to control this in the future.

PR #2028, opened 2022-04-11
Add ceph to single replica lane

jean-edouard:
The single-replica lane is not bound to a sig, so it needs a storage backend to run sig-storage tests.


dhiller (approved):
/approve

PR #2030, opened 2022-04-12
images mirror: Add library/registry:2

oshoval:
library/registry:2.7.1 doesn't support arm64
where library/registry:2 does.


oshoval:
We might need to change to docker pull and push instead of skopeo copy
because the source image lregistry:2.7.1 does support arm64 (assuming it is indentical to `library/registry:2.7.1`)

PR #2033, opened 2022-04-12
kubevirtci, publish: Mount modules and devs

qinqon:
To be able to build the alpine image /lib/modules and /dev need to be
mounted.

Signed-off-by: Enrique Llorente <ellorent@redhat.com>


dhiller (approved):
/approve

PR #2034, opened 2022-04-12
single-node lane: add a "focus all" option to avoid storage-req auto-focus

jean-edouard:
Adding `KUBEVIRT_STORAGE="rook-ceph-default"` actually focuses the whole lane on `storage-req` tests...
Specifying a focus of "all tests" works around that.


dhiller (approved):
/approve

PR #2038, opened 2022-04-13
Add namespace label to loki alerts

brianmcarey:
Alertmanager is matching on the namespace label for the slack receiver

/cc @dhiller 


dhiller (approved):
/approve

PR #2047, opened 2022-04-25
[containerdisks] Add a pipeline test presubmit job

rmohr:
Run a minimal containerdisk pipeline check based on https://github.com/kubevirt/containerdisks/blob/main/pipeline.sh which will verify cirros as example workload.


rmohr:
/approve

Rehearsal passes.

PR #2051, opened 2022-04-26
Add quay oauth token to the quay credentials secret

rmohr:
To directly talk to the quay api an oauth token needs to be used. This token can be used for interacting with the containerdisks repo to push docs there.


dhiller (approved):
/approve

rmohr:
seems to pass.

/unhold

PR #2052, opened 2022-04-26
kubevirtci: Install libvirt + genisoimage

qinqon:
To build alpine now we are using virt-install for that the job needs
libvirt and mkisofs.

Signed-off-by: Enrique Llorente <ellorent@redhat.com>


dhiller (approved):
/approve

PR #1632, opened 2021-09-30
Replace docker with podman in bootstrap image

fgimenez:
Changes to start using podman in our CI instead of docker:

* Modification of the bootstrap image, replacing docker with podman. The image symlinks `/usr/bin/docker` to `/usr/bin/podman` to make the transition easier. Also changed the runner scripts to start podman as a system service and make the required setup and teardown.
* Added a new `preset-pinc-enabled` preset that sets up the env var, volume and volume mounts that will be used in podman-based jobs
* Modified docker-mirror-proxy preset to include HTTP proxy env vars that will be used by the podman process.

This PR represents the first step in the migration from docker to podman for the CI jobs, once it is merged and the bootstrap image built we need to gradually upgrade the jobs to use this new image and replace the `dind` preset with the new `pinc` preset, starting with periodics, then optional and always_run false presubmits, then required presubmits.

Signed-off-by: Federico Gimenez <fgimenez@redhat.com>

PR #2055, opened 2022-04-27
Replace docker with podman in bootstrap image

brianmcarey:
This PR aims to replace docker with podman in the bootstrap image. The bootstrap image is used by a large number of prow jobs in KubeVirt CI including the large e2e jobs.

The changes in this PR are based on #1632 with some minor updates to match changes in podman handling in kubevirtci. 

/cc @dhiller


images/bootstrap/Dockerfile
```diff
@@ -73,23 +105,28 @@ RUN wget -q https://dl.google.com/dl/cloudsdk/channels/rapid/google-cloud-sdk.ta
...
-# Install packages
 
-RUN dnf install -y \
-        kmod \
-        procps-ng \
-        moby-engine && \
-    dnf -y clean all
+RUN dnf -y update; dnf -y reinstall shadow-utils && \
+    dnf -y install crun podman --exclude container-selinux && \
+    rm -rf /var/cache /var/log/dnf* /var/log/yum.* && \
```
dankenigsberg, line 113: why is this used instead of the higher-level `dnf -y clean all` ?

images/bootstrap/Dockerfile
```diff
@@ -73,23 +105,28 @@ RUN wget -q https://dl.google.com/dl/cloudsdk/channels/rapid/google-cloud-sdk.ta
...
+    ln -s /usr/bin/podman /usr/bin/docker && \
+    dnf update -y --refresh --enablerepo=updates-testing podman
```
dankenigsberg, line 115: do we really want to live on the edge ans use `podman` from Fedora testing? I would not add this without a clear explanation why.

images/bootstrap/Dockerfile
```diff
@@ -14,6 +14,38 @@
 # limitations under the License.
 
 # Includes basic workspace setup, with gcloud and a bootstrap runner
+FROM fedora:34 AS builder
```
dankenigsberg, line 17: time has passed, 35 may be better?

brianmcarey: yes fedora 35 has been available for a while now - updated.

images/bootstrap/Dockerfile
brianmcarey: I have tested this without enabling Fedora testing and it works so I removed this.

images/bootstrap/Dockerfile
brianmcarey: yes makes more sense to use `dnf -y clean all` - updated

images/bootstrap/Dockerfile
brianmcarey: updated to fedora 36

images/bootstrap/Dockerfile
```diff
@@ -73,23 +104,27 @@ RUN wget -q https://dl.google.com/dl/cloudsdk/channels/rapid/google-cloud-sdk.ta
...
 
-RUN dnf install -y \
-        kmod \
-        procps-ng \
-        moby-engine && \
-    dnf -y clean all
+RUN dnf -y update; dnf -y reinstall shadow-utils && \
+    dnf -y install crun podman --exclude container-selinux && \
+    dnf -y clean all && \
+    ln -s /usr/bin/podman /usr/bin/docker
```
oshoval, line 113: I think that we better auto detect the CRI instead of having the soft link
(kubevirt basically has already auto detect)

images/bootstrap/Dockerfile
```diff
@@ -14,6 +14,37 @@
...
+
+# Cache latest stable golang version
+RUN curl -sL -o /bin/gimme https://raw.githubusercontent.com/travis-ci/gimme/master/gimme && \
+    chmod +x /bin/gimme && \
+    mkdir -p /etc/setup.mixin.d/ && \
+    echo 'eval $(gimme ${GIMME_GO_VERSION})' > /etc/setup.mixin.d/golang.sh
+RUN cat /etc/setup.mixin.d/golang.sh
+RUN . /etc/setup.mixin.d/golang.sh
+
+RUN dnf install -y \
```
oshoval, line 34: maybe worth to install `buildah` and `qemu-static-user` 
`buildah` allows multi arch build with podman, and `qemu-static-user` allows the emulation
of cross compiling

images/bootstrap/runner.sh
```diff
@@ -41,94 +41,64 @@ setup_ca(){
...
         if [[ ${WAIT_N} -lt ${MAX_WAIT} ]]; then
             WAIT_N=$((WAIT_N+1))
-            echo "Waiting for docker to be ready, sleeping for ${WAIT_N} seconds."
+            echo "Waiting for podman socket to be ready, sleeping for ${WAIT_N} seconds."
             sleep ${WAIT_N}
         else
             echo "Reached maximum attempts, not waiting any longer..."
             break
         fi
     done
```
oshoval, line 99: 1. timeout until can be used instead please (nice to have)

2. In case of an error and also on success, it just break, shouldn't it exit 1 on error ?
(unless it catch it nicely somewhere else)

oshoval (commented):
Very nice thanks

github/ci/prow-deploy/kustom/base/configs/current/config/config.yaml
```diff
@@ -435,6 +450,10 @@ presets:
   env:
   - name: CA_CERT_FILE
     value: /etc/docker-mirror-proxy/ca.crt
+  - name: CONTAINER_HTTP_PROXY
+    value: http://docker-mirror-proxy.kubevirt-prow.svc:3128
+  - name: CONTAINER_HTTPS_PROXY
+    value: http://docker-mirror-proxy.kubevirt-prow.svc:3128
```
oshoval, line 456: https right?

github/ci/prow-deploy/kustom/base/configs/current/config/config.yaml
```diff
@@ -420,6 +420,21 @@ presets:
...
     mountPath: /docker-graph
+- labels:
+    preset-pinc-enabled: "true"
+  env:
+  - name: PODMAN_IN_CONTAINER_ENABLED
+    value: "true"
+  - name: KIND_EXPERIMENTAL_PROVIDER
+    value: "podman"
+  - name: PODMAN_IN_CONTAINER_IPV6_ENABLED
+    value: "true"
```
oshoval, line 431: question, do we really need all the parameters ?
what if we can have fewer global ones that if declared the others will be internally declared ?

images/bootstrap/Dockerfile
```diff
@@ -14,6 +14,37 @@
 # limitations under the License.
 
 # Includes basic workspace setup, with gcloud and a bootstrap runner
+FROM fedora:36 AS builder
+
+ENV PKG_CONFIG_PATH="/usr/lib/pkgconfig"
+ENV PODMAN_VERSION="v4.0.3"
```
oshoval, line 20: can we use please older version ? for example 3.4 ?
I have this error https://github.com/containers/podman/issues/13327 on CI [1]
and one mentioned WA is to downgrade

[1] https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_project-infra/2125/rehearsal-pull-cluster-network-addons-operator-unit-test/1529420827697614848

Locally i am using podman 3.4.4 and all works good
Thanks

oshoval:
podman is tls secured by default

We have here some TLS failures
https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_project-infra/2125/rehearsal-pull-e2e-cluster-network-addons-operator-workflow-k8s/1529420828771356672
"Error: initializing source docker://quay.io/kubevirtci/gocli:2205030954-99bd4d1: pinging container registry quay.io: Get "https://quay.io/v2/": x509: certificate signed by unknown authority"

Maybe we need to apply some certs ? or fix the certs on the image ?
(alternatively less secure solution is to disable podman tls checking, but not sure we want it)

images/bootstrap/Dockerfile
```diff
@@ -14,6 +14,37 @@
...
+
+ENV GIMME_GO_VERSION=1.16.8
+
+# Cache latest stable golang version
+RUN curl -sL -o /bin/gimme https://raw.githubusercontent.com/travis-ci/gimme/master/gimme && \
+    chmod +x /bin/gimme && \
+    mkdir -p /etc/setup.mixin.d/ && \
+    echo 'eval $(gimme ${GIMME_GO_VERSION})' > /etc/setup.mixin.d/golang.sh
+RUN cat /etc/setup.mixin.d/golang.sh
+RUN . /etc/setup.mixin.d/golang.sh
```
oshoval, line 32: It seems we don't have go in the PATH ?

https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_project-infra/2130/rehearsal-pull-kubernetes-nmstate-unit-test/1529702244696461312

images/bootstrap/Dockerfile
oshoval: Ok np

images/bootstrap/runner.sh
brianmcarey: It should exit 1 instead of break - I will make that update.

images/bootstrap/Dockerfile
brianmcarey: This part of the Dockerfile is used to build podman v4.0.3 

This may no longer needed as there is newer versions of podman available in the fedora repo that wouldn't have been available when the original PR was made. 

I think I will try to run a couple of tests with the version in the fedora 36 repos and we may be able to remove this builder section altogether.

Its podman v4.1.0 in Fedora 36 at the moment  - if that doesn't work we can look at building an older release

github/ci/prow-deploy/kustom/base/configs/current/config/config.yaml
brianmcarey: yes it might be better to set some of these in the runner.sh script instead.

images/bootstrap/runner.sh
brianmcarey: I made that change here https://github.com/kubevirt/project-infra/pull/2140 - a rebase will probably be required to take it in here.

images/bootstrap/runner.sh
```diff
@@ -41,94 +41,64 @@ setup_ca(){
...
     )
-    # the service can be started but the docker socket not ready, wait for ready
+    # the service can be started but the socket not ready, wait for ready
     WAIT_N=0
     MAX_WAIT=5
     while true; do
-        # docker ps -q should only work if the daemon is ready
-        docker ps -q > /dev/null 2>&1 && break
+        # podman ps -q should only work if the daemon is ready
+        podman ps -q > /dev/null 2>&1 && break
```
0xFelix, line 90: Maybe it would be better to use the socket for testing if podman works? Like done here: https://github.com/kubevirt/kubevirtci/blob/23ddfc29c03bb1ccaeb79cd204163f8108db1c36/cluster-up/cluster/ephemeral-provider-common.sh#L24

0xFelix: Or here https://github.com/kubevirt/kubevirtci/blob/23ddfc29c03bb1ccaeb79cd204163f8108db1c36/cluster-up/cluster/ephemeral-provider-common.sh#L18

brianmcarey: yes good points - I will look into improving this check - thanks @0xFelix !

brianmcarey: yes good points - I will look into improving this check - thanks @0xFelix !

images/bootstrap/Dockerfile
brianmcarey: I was looking today - the docker command is used quite a bit in the kubevirtci kind providers such as sriov

oshoval:
With the latest image, I see those errors here
https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_project-infra/2125/rehearsal-pull-e2e-cluster-network-addons-operator-workflow-k8s/1534095563631366144

```
Enabling IPV6.
/usr/local/bin/runner.sh: line 66: sysctl: command not found
/usr/local/bin/runner.sh: line 67: sysctl: command not found
modprobe: FATAL: Module ip6table_nat not found in directory /lib/modules/4.18.0-365.el8.x86_64
Podman in Container enabled, initializing in podman compatible mode...
```

images/bootstrap/runner.sh
```diff
@@ -41,94 +41,64 @@ setup_ca(){
...
-            ${DOCKER_OPTS} \
-                >/var/log/dockerd.log 2>&1 &
+        export HTTP_PROXY=${CONTAINER_HTTP_PROXY}
+        export HTTPS_PROXY=${CONTAINER_HTTPS_PROXY}
+	mkdir -p ${XDG_RUNTIME_DIR}/podman
+        podman system service \
+               -t 0 \
+               unix:///${XDG_RUNTIME_DIR}/podman/podman.sock \
+               >/var/log/podman.log 2>&1 &
+        echo "${!}" > /var/run/podman.pid
```
oshoval, line 83: I think that instead this we would able to use `systemctl enable --now podman.socket` once https://github.com/kubevirt/kubevirtci/pull/824 is merged (because it knows to detect the default root socket path `/run/podman/socket`)

images/bootstrap/Dockerfile
oshoval: About the last comment please see https://github.com/kubevirt/kubevirtci/pull/826

images/bootstrap/runner.sh
oshoval: I see thanks

Would it be good idea to include systemd?

What about using the default podman.socket path ? `run/podman/podman.socket`

images/bootstrap/runner.sh
```diff
@@ -82,6 +82,7 @@ if [[ "${PODMAN_IN_CONTAINER_ENABLED}" == "true" ]]; then
                unix:///${XDG_RUNTIME_DIR}/podman/podman.sock \
                >/var/log/podman.log 2>&1 &
         echo "${!}" > /var/run/podman.pid
+	ln -s ${XDG_RUNTIME_DIR}/podman/podman.sock /var/run/docker.sock
```
oshoval, line 85: Btw I tried that with vm based providers, and it didn't work
nor will be needed soon when 
https://github.com/kubevirt/kubevirt/pull/8010
https://github.com/kubevirt/kubevirt/pull/8048
are merged

I think we better fix whatever needed and remove all softlink to docker

brianmcarey: I am not sure if you saw the prowjob linked in the commit - https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_project-infra/2191/rehearsal-check-provision-k8s-1.23/1544729115590922240#1:build-log.txt%3A2152 - its a panic from gocli and I don't see updates in this area in the changes you linked? 

Here is the PR for making gocli work with podman - https://github.com/kubevirt/kubevirtci/pull/554

brianmcarey: It looks to be using a few docker specific modules so it may not be as straight forward as the other scripts.
https://github.com/kubevirt/kubevirtci/blob/111fd507537476c5e5ff9d7a914c6ece8862aaaa/cluster-provision/gocli/docker/docker.go#L15-L17

oshoval:
Wondering, can we have an image that has both docker and podman ?
at least at the beginning
This way it depends on the repo / job itself which to use, according the detection order / env vars

and will be easier to switch between them in case needed, until we stabilize it to use podman

oshoval:
https://github.com/kubevirt/project-infra/pull/2218 was merged

/lgtm

Thanks
(even if there more steps needed we can done later imho)

dhiller (approved):
Two nits inline, other than that looks great! Thanks!

/approve

/hold in case you want to handle the first one, feel free to remove the hold otherwise

EDIT: i mean the preset name

github/ci/prow-deploy/kustom/base/configs/current/config/config.yaml
```diff
@@ -420,6 +420,17 @@ presets:
   volumeMounts:
   - name: docker-graph
     mountPath: /docker-graph
+- labels:
+    preset-pinc-enabled: "true"
```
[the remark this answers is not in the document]
brianmcarey: I think you are right here - it might be better to use the full name so it easier to understand. I will make that update now.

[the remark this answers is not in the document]
oshoval: I think pinc is known a lil
https://github.com/kukaryambik/pinc
imho short name is better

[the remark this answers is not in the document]
dhiller: @oshoval I disagree. I think we rather should be clear here and not force the reader to have to remember what this preset is about.

And since Brian has made the change already, let's go with it.

dhiller (approved):
/unhold

images/bootstrap/runner.sh
```diff
@@ -41,94 +41,68 @@ setup_ca(){
...
-        if [[ "${DOCKER_DEBUG:-false}" == "true" ]]; then
-            echo "Copying docker log to ARTIFACTS"
-            cp /var/log/dockerd.log $ARTIFACTS
-        fi
-        echo "Cleaning up after docker"
-        docker ps -aq | xargs -r docker rm -f || true
-        kill "$(</var/run/docker.pid)" || true
-        wait "$(</var/run/docker.pid)" || true
+# runs custom podman data root cleanup binary and debugs remaining resources
+cleanup_pinc() {
```
oshoval, line 45: here you dont want to change the pinc ?

brianmcarey:
The `build-multiarch-bootstrap-image` prow job failed due to reaching the timeout. I have opened an issue (https://github.com/kubevirt/project-infra/issues/2252) for it and a PR (https://github.com/kubevirt/project-infra/pull/2253) to increase the timeout to 5 hours to unblock these changes.

github/ci/prow-deploy/kustom/base/configs/current/config/config.yaml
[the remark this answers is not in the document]
oshoval: Bttw I think that at the end pinc will come a term like dind, but whatever you prefer

images/bootstrap/runner.sh
```diff
@@ -41,94 +41,68 @@ setup_ca(){
...
-            --data-root=/docker-graph \
-            --init-path /usr/libexec/docker/docker-init \
-            --userland-proxy-path /usr/libexec/docker/docker-proxy \
-            ${DOCKER_OPTS} \
-                >/var/log/dockerd.log 2>&1 &
+        export HTTP_PROXY=${CONTAINER_HTTP_PROXY}
+        export HTTPS_PROXY=${CONTAINER_HTTPS_PROXY}
+	export KIND_EXPERIMENTAL_PROVIDER="podman"
+	mkdir -p ${XDG_RUNTIME_DIR}/podman
+        podman system service \
```
oshoval, line 80: If i run the new image
`docker run -it --privileged  -e PODMAN_IN_CONTAINER_ENABLED=true quay.io/kubevirtci/bootstrap:v20220816-91d8af2 bash`

I see `/podman/podman.sock` is created
because `XDG_RUNTIME_DIR` is empty here

btw it affects kubevirtci / kubevirt, because we look for `/run/podman/podman.sock`
wdyt about using `/run/podman/podman.sock` please ?
`XDG_RUNTIME_DIR` is empty anyhow
this way we won't need to export `KUBEVIRTCI_PODMAN_SOCKET`

Currently because the socket isn't found, it select docker / docker socket which are an alias to podman and its socket

brianmcarey: Yes I think we should update to the path that kubevirtci expects.

oshoval: Not tested yet https://github.com/kubevirt/project-infra/pull/2258
(lets see the presubmit job)

PR #2064, opened 2022-04-29
sig-performance: Increase KUBEVIRT_MEMORY_SIZE to 10G

lyarwood:
https://github.com/kubevirt/kubevirt/issues/7658 identified that the sig-performance job was timing out due to ~12 VMIs becoming unschedulable due to memory pressure in the test environment. 

This PR simply increases the available k8s/KubeVirt memory in the env by ~4G to allow these final 12 VMIs to launch. A previous increase to the overall job timeout is also removed as this shouldn't be required anymore with the job typically completing in around ~30 mins prior to this issue.

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
@@ -445,7 +445,7 @@ presubmits:
         - name: KUBEVIRT_STORAGE
           value: hpp
         - name: KUBEVIRT_MEMORY_SIZE
-          value: 9G
+          value: 10G
         - name: KUBEVIRT_NUM_NODES
           value: "4"
         - name: KUBEVIRT_DEPLOY_PROMETHEUS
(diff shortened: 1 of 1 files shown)


github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
```diff
@@ -445,7 +445,7 @@ presubmits:
         - name: KUBEVIRT_STORAGE
           value: hpp
         - name: KUBEVIRT_MEMORY_SIZE
-          value: 9G
+          value: 10G
```
xpivarc, line 448: I wonder if we should also increase the memory request. What do you think?

lyarwood: Yeah happy to do that as well I just wasn't sure what the infra we are deploying to actually looks like, IOW if we had >44G to play with. I'll bump it now assuming we can.

xpivarc: @dhiller keep me honest

PR #2066, opened 2022-04-29
knmstate: add e2e upgrade lane

rhrazdil:

Add job for kubernetes-nmstate upgrade lane.
It's not required at the beginning, we can make it required later when we see it's stable.


qinqon:
Let's unlock the chicken egg
/lgtm
/approve

PR #2068, opened 2022-05-02
sriov: Introduce place holder DaemonSet

oshoval:
CI SR-IOV nodes are running both SR-IOV job and regular jobs.
SR-IOV jobs have a higher priority class, which isn't preemptive.
Therefore a new SR-IOV might need to wait until a regular job finishes.

This PR introduces a DaemonSet that will spin a pod, 
that reserves the resources needed for SR-IOV job.

Once a SR-IOV job is scheduled it will preempt the place holder, and run until completed.
Once it is done, the place holder will be rescheduled, saving again the resources.
This way SR-IOV jobs won't need to wait and a dedicate place holder pod will wait for them.

Since the place holder requests the same resources that the real SR-IOV job needs and has
a higher priority than regular jobs, it will not let other pods replace it, nor take those resources.

The place holder pods have anti affinity of `sriov-pod` label, so it won't run when a real SR-IOV jobs runs.
On the other hand, it has the `sriov-pod` label as well, so it will be the one that is preempted by
the SR-IOV pod, because otherwise the SR-IOV pod can't run.
The scheduler preempts only pods that if preempted the designated pod can run instead.

### DIFF
--- a/github/ci/prow-deploy/kustom/overlays/workloads-production/resources/place-holder.yaml
+++ b/github/ci/prow-deploy/kustom/overlays/workloads-production/resources/place-holder.yaml
@@ -1,0 +1,42 @@
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: sriov-place-holder
+  labels:
+    name: sriov-place-holder
+spec:
+  selector:
+    matchLabels:
+      name: sriov-place-holder
+  updateStrategy:
+    type: RollingUpdate
+  template:
+    metadata:
+      labels:
+        name: sriov-place-holder
+        sriov-pod: "true"
+    spec:
+      affinity:
+        podAntiAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+          - labelSelector:
+              matchExpressions:
+              - key: sriov-pod
+                operator: In
+                values:
+                - "true"
+            topologyKey: kubernetes.io/hostname
+      nodeSelector:
+        hardwareSupport: sriov-nic
+      terminationGracePeriodSeconds: 1
+      priorityClassName: sriov-place-holder
+      tolerations:
+      - operator: Exists
+        effect: NoSchedule
+      containers:
+      - name: sriov-place-holder
+        image: k8s.gcr.io/busybox
+        command: [ "/bin/sh", "-c", "while true; do sleep 86400; done" ]
+        resources:
+          requests:
+            memory: 29Gi
--- a/github/ci/prow-deploy/kustom/overlays/workloads-production/resources/priority-classes.yaml
+++ b/github/ci/prow-deploy/kustom/overlays/workloads-production/resources/priority-classes.yaml
@@ -3,8 +3,8 @@ apiVersion: scheduling.k8s.io/v1
 kind: PriorityClass
 metadata:
   name: sriov
-value: 1000000
-preemptionPolicy: Never
+value: 1000001
+preemptionPolicy: PreemptLowerPriority
 globalDefault: false
 description: "Allows sriov jobs to be scheduled with higher priority."
 ---
@@ -16,3 +16,12 @@ value: 1000000
 preemptionPolicy: Never
 globalDefault: false
 description: "Allows gpu jobs to be scheduled with higher priority."
+---
+apiVersion: scheduling.k8s.io/v1
+kind: PriorityClass
+metadata:
+  name: sriov-place-holder
+value: 1000000
+preemptionPolicy: Never
+globalDefault: false
+description: "Allows to have sriov place holder jobs"


enp0s3:
@brianmcarey @dhiller I think that we need @oshoval to be available to support us if needed, so weekend will be problematic. Can we do it until Thursday?

github/ci/prow-deploy/kustom/overlays/workloads-production/resources/priority-classes.yaml
```diff
@@ -16,3 +16,12 @@ value: 1000000
 preemptionPolicy: Never
 globalDefault: false
 description: "Allows gpu jobs to be scheduled with higher priority."
+---
+apiVersion: scheduling.k8s.io/v1
+kind: PriorityClass
+metadata:
+  name: sriov-place-holder
+value: 1000000
```
dhiller, line 24: So IIUC the sriov place holder pods will have higher priority than the normal pods running with default priority.

K8s docs say:
> Preemption considers Pod priority and attempts to choose a set of targets with the lowest priority. Higher-priority Pods are considered for preemption only if the removal of the lowest priority Pods is not sufficient to allow the scheduler to schedule the preemptor Pod, or if the lowest priority Pods are protected by PodDisruptionBudget
[source](https://kubernetes.io/docs/concepts/scheduling-eviction/pod-priority-preemption/#interactions-of-pod-priority-and-qos)

This makes me think that the normal job pods are evicted first, as they have the lowest priority. Why don't we just change the Preemption like [here](https://github.com/kubevirt/project-infra/pull/2068/files#diff-56380dd553d457f6b73e214c4f728aa374364c300c133bbceddac81d8aa3f0f4R7) and preempt the job pods (which will be restarted by prow-controller-manager anyway)?

enp0s3: I was thinking that the regular pods with the lower priority belong to some default priority class of k8s which is not preemptive, @oshoval please keep me honest here. I will double check it

oshoval: @dhiller 
Since the DS has the label that the sriov has anti-affinity to
https://github.com/kubevirt/project-infra/pull/2068/files#diff-894e55e0973533b54e923ff56852bf9a6ec68892d7c9a97cc1a251a370eddb80R17
the place-holder pod will be evicted first, even that the default pods have lower priority
We don't want the sriov to stop random running jobs, but specifically the place-holder, there fore the following design.

@enp0s3 the default pods don't have pc, because there isn't a global default pc, so they have priority zero,
it is not preemptive, which means they can't preempt. but they still can be preempted,
but because the above reason (label and anti-affinity), the place-holder will be the one that is preempted

github/ci/prow-deploy/kustom/overlays/workloads-production/resources/place-holder.yaml
```diff
@@ -0,0 +1,42 @@
...
+      tolerations:
+      - operator: Exists
+        effect: NoSchedule
+      containers:
+      - name: sriov-place-holder
+        image: k8s.gcr.io/busybox
+        command: [ "/bin/sh", "-c", "while true; do sleep 86400; done" ]
+        resources:
+          requests:
+            memory: 29Gi
```
xpivarc, line 42: This will consume memory for jobs that could be ready to run when no SRIOV jobs are ready to run. Let's have a quick look at how many SRIOV jobs we see in our CI and verify this compromise is relevant.

oshoval: Just spoke with @brianmcarey and @EdDev
we have a buffer so we are safe to reserve this chunk.

Moreover we can't reduce the memory to 18Gi because those numbers are mainly in order to control how many run on the same time, not for how many memory it actually need (it needs 15.6 according the recent check with Brian).

Personally I find killing a random pod problematic, we will need to debug again and again why our jobs failed, even if they will be rescheduled.

@enp0s3 @dhiller FYI

EdDev: > This will consume memory for jobs that could be ready to run when no SRIOV jobs are ready to run. Let's have a quick look at how many SRIOV jobs we see in our CI and verify this compromise is relevant.

What exactly is consumed? The requested values are just plain resources, it does not allocate any memory.

The SR-IOV machines have been introduced to serve SR-IOV workloads, the fact that other workloads run on them is nice, but it should not come on the cost of the SR-IOV ones.

At the moment, SR-IOV jobs are delayed due to non SR-IOV jobs running on these machine. I would like this to be optimized and avoided.

If the CI maintainers prefer to preempt running jobs instead of preserving a slot for the SR-IOV jobs, I do not have a strong objection. I think it is less desired, but I prefer prioritizing SR-IOV as the first priority.

xpivarc: >Moreover we can't reduce the memory to 18Gi because those numbers are mainly in order to control how many run on the same time, not for how many memory it actually need (it needs 15.6Gi according the recent check with Brian).

Seems like a hack that should be replaced by a proper mechanism.

>What exactly is consumed? The requested values are just plain resources, it does not allocate any memory.

Kubernetes scheduler is working based on a request, not actual consumption. Therefore you don't get to run different jobs there.

>The SR-IOV machines have been introduced to serve SR-IOV workloads, the fact that other workloads run on them is nice, but it should not come on the cost of the SR-IOV ones.

Agree but let's explore an optimal solution that also provides the best utilization of the machine as we are seeing the requests to maximize our CI capacity.

>At the moment, SR-IOV jobs are delayed due to non SR-IOV jobs running on these machine. I would like this to be optimized and avoided.

Do you have any examples? It would greatly help us understand why this happen and if a simpler solution couldn't be applied.

>If the CI maintainers prefer to preempt running jobs instead of preserving a slot for the SR-IOV jobs, I do not have a strong objection. I think it is less desired, but I prefer prioritizing SR-IOV as the first priority.

The preemption was just an idea brought up during a discussion of this PR. Further exploring of consequences is needed.
Example: It doesn't make sense to use it if we would preempt every 15 minutes but it does make sense if it would be once per day.

github/ci/prow-deploy/kustom/overlays/workloads-production/resources/priority-classes.yaml
```diff
@@ -3,8 +3,8 @@ apiVersion: scheduling.k8s.io/v1
 kind: PriorityClass
 metadata:
   name: sriov
-value: 1000000
-preemptionPolicy: Never
+value: 1000001
```
xpivarc, line 6: Will this be properly updated on the cluster?

github/ci/prow-deploy/kustom/overlays/workloads-production/resources/place-holder.yaml
enp0s3: /hold
@xpivarc Didn't see your comments.

xpivarc (commented):
/hold 
Putting it on hold as I am not sure this change will be correctly applied to the cluster.
@dhiller @oshoval

github/ci/prow-deploy/kustom/overlays/workloads-production/resources/place-holder.yaml
xpivarc: No need to hold for this if others feel it's good for now. I just opened discussion about further steps.

github/ci/prow-deploy/kustom/overlays/workloads-production/resources/priority-classes.yaml
enp0s3: Hpefully #2087 should handle this situation

enp0s3:
/unhold
#2087 was merged

oshoval:
Thanks,
it isn't deployed, maybe we need to add it to the kustomize list ?

github/ci/prow-deploy/kustom/overlays/workloads-production/resources/place-holder.yaml
oshoval: > Seems like a hack that should be replaced by a proper mechanism.

FYI
There is this extended resource method if we find it helpful
(to create other one to control how many run on each machine)
https://github.com/kubevirt/project-infra/blob/main/hack/patch_node.sh

xpivarc: Yes, this would be the right path(IMHO).

PR #2074, opened 2022-05-04
sriov: Run cluster-down only once in trap

oshoval:
Since the trap had ERR as well as EXIT,
it would run the trap twice in case an error happened,
one for the ERR and one for the EXIT.
Having it on EXIT is enough because it happens
either on good flow and error flow.

SIGTERM is not needed, because EXIT will happen
anyway as well.

Remove deprecated kind-1.19-sriov job.


oshoval:
The failures are expected on the non 1.22 lane, as those lanes doesn't exists anymore on main branch
I will limit the PR to only change the main branch yamls and remove the kind-1.19 lane 

once the 1.22-sriov job finishes

PR #2083, opened 2022-05-06
Use podman for mkpj.sh script if installed

brianmcarey:
When the mkpj.sh script is run and the docker daemon is not running it fails. If podman is installed, I would prefer it to use podman as it doesn't require a running daemon.

```
./hack/mkpj.sh
Using default tag: latest 
Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running? 
```

/cc @dhiller 


dhiller (commented):
/approve

dhiller (approved):

PR #2087, opened 2022-05-11
prow-deploy: delete priority classes before re-deployment

enp0s3:
Ansible role was updated to delete the prirority classes prior
to the launch of the deployment itself. This is because updating
the priority class that is in use will fail, thus its requried
to re-create it.

**Discalimer**
* Trivial approach (no usage of `yq`)
* Currently w/o testing via Molecule. 

### DIFF
--- a/github/ci/prow-deploy/tasks/deploy.yml
+++ b/github/ci/prow-deploy/tasks/deploy.yml
@@ -28,6 +28,13 @@
     KUBECONFIG: "{{ kubeconfig_path }}"
   when: bootstrap_full_config
 
+- name: Delete priority classes on prow-workloads BM cluster
+  when: delete_priority_classes
+  shell: |
+    kubectl delete pc sriov vgpu
+  environment:
+    KUBECONFIG: "{{ kubeconfig_path }}"
+
 - name: Launch deployment
   shell: |
     # for now cant execute "kubectl apply --server-side" reliably on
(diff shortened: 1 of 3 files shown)


github/ci/prow-deploy/tasks/deploy.yml
```diff
@@ -28,6 +28,13 @@
     KUBECONFIG: "{{ kubeconfig_path }}"
   when: bootstrap_full_config
 
+- name: Delete priority classes on prow-workloads BM cluster
+  when: delete_priority_classes
+  shell: |
+    kubectl delete pc sriov vgpu
```
oshoval, line 34: once we have the new sriov-place-holder pc we would need to add it here as well

oshoval (approved):
Thanks

github/ci/prow-deploy/tasks/deploy.yml
oshoval: btw you might want to delete -f the pc file
it will auto delete all and just what is there

dhiller: > once we have the new sriov-place-holder pc we would need to add it here as well

Or add `--ignore-not-found=true`

dhiller: > btw you might want to delete -f the pc file it will auto delete all and just what is there (auto updating, but not backward compatible, in case we change the file by removing a pc)

@enp0s3 keep me honest, but we discussed that using the file is not possible, as it's already the new configuration, thus it will fail to remove older stuff (i.e. if something is removed)

dhiller:
/approve

Thanks for this!

oshoval (approved):
Thanks

oshoval:
this is the postsubmit ?
if so it failed

https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/logs/post-project-infra-prow-control-plane-deployment/1524310295969796096

PR #2092, opened 2022-05-11
prow-deploy: disable usage of remote caching for release configs

enp0s3:
Remote cache eviction happens too often, and since the garbage collector
is unaware of the Bazel build graph, it will delete the stale items in
a LRU manner w/o taking care about them being intermediate artifacts.

Therefore the lanes tend to fail and until the cache wipe won't
happen the job won't work.

As a short-term solution we would like to reduce the load on the
Bazel greenhouse cache deployment, thus disable its usage for the
release branches CI lanes.


dhiller (approved):
/approve

xpivarc:
@enp0s3 @dhiller  ^^^

PR #2093, opened 2022-05-11
sriov: Add namespace to the place holder ds

oshoval:
Without the namespace, the anti affinity
doesn't make sure that only either
sriov or place-holder are running.

Add sriov-place-holder PC to the delete command
so when we are redeploying it will clean the previous PCs as needed.


enp0s3 (approved):
/approve

@oshoval Thank you!

PR #2096, opened 2022-05-12
periodic, sriov: Add missing pc

oshoval:
Since we have now sriov place holder,
the jobs of sriov should have
pc of sriov, so they can preempt the place holder,
and run instead.

Remove trailing spaces.


dhiller (approved):
/approve

PR #2103, opened 2022-05-16
release-0.52, sriov: Add missing pc (#2096)

oshoval:
Since we have now sriov place holder,
the jobs of sriov should have
pc of sriov, so they can preempt the place holder,
and run instead.

Out of the releases, only 0.52 missed it,
so this PR add it manually.


xpivarc:
@oshoval This will be needed for every release, right?

dhiller (approved):
/approve

PR #2086, opened 2022-05-10
Add mirror for previous minor release of cri-o

brianmcarey:
An issue with the latest release of cri-o caused failures in kubevirtci
presubmits. Using the previous minor releases avoided these failures.
This change adds the previous cri-o minor releases to the kubevirtci
mirror

See https://github.com/kubevirt/kubevirtci/pull/796 for more details

/cc @xpivarc @dhiller 

Signed-off-by: Brian Carey <bcarey@redhat.com>

PR #2104, opened 2022-05-16
Remove previous minor release of cri-o from mirror job

brianmcarey:
The previous minor releases were added to the mirror due to an issue
with the latest cri-o release. The previous minor releases have been rebuilt
and now has the same issue.

The previous minor releases were introduced here https://github.com/kubevirt/project-infra/pull/2086 but are no longer needed.

/cc @dhiller 


dhiller (approved):
/approve

PR #2124, opened 2022-05-25
kubevirt: add lane used for running clean code checks

rhrazdil:
This PR adds lane to kubevirt project with a purpose to run checks
that ensures the introduced code change is aligned with kubevirt clean
code standards.

The lane executes `make lint`.

NOTE: I'm not sure we need all the labels, I have "forked" the job properties
from existing netowork-sig job.

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
@@ -1884,6 +1884,33 @@ presubmits:
   - always_run: true
     annotations:
       fork-per-release: "true"
+    cluster: ibm-prow-jobs
+    decorate: true
+    decoration_config:
+      grace_period: 5m0s
+      timeout: 1h0m0s
+    max_concurrency: 11
+    name: pull-kubevirt-code-lint
+    optional: true
+    skip_branches:
+    - release-\d+\.\d+
+    skip_report: false
+    spec:
+      containers:
+      - command:
+        - /usr/local/bin/runner.sh
+        - /bin/sh
+        - -c
+        - |
+          make lint
+        image: quay.io/kubevirtci/bootstrap:v20220512-78048f1
+        name: ""
+        resources:
+          requests:
+            memory: 4Gi
+  - always_run: true
+    annotations:
+      fork-per-release: "true"
       testgrid-dashboards: kubevirt-presubmits
     cluster: prow-workloads
     decorate: true


github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
```diff
@@ -1881,6 +1881,42 @@ presubmits:
...
+      - command:
+        - /usr/local/bin/runner.sh
+        - /bin/sh
+        - -c
+        - make lint
+        image: quay.io/kubevirtci/bootstrap:v20220512-78048f1
+        name: ""
+        resources:
+          requests:
+            memory: 6Gi
```
brianmcarey, line 1915: Does this quick check require 6Gi of memory? Would it run ok with 4Gi? 

Looks good apart from that.

github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
```diff
@@ -1881,6 +1881,42 @@ presubmits:
...
+      testgrid-dashboards: kubevirt-presubmits
+    cluster: prow-workloads
+    decorate: true
+    decoration_config:
+      grace_period: 5m0s
+      timeout: 1h0m0s
+    labels:
+      preset-bazel-cache: "true"
+      preset-bazel-unnested: "true"
+      preset-dind-enabled: "true"
```
brianmcarey, line 1897: I think this label is not needed for this check - it doesn't require docker in docker

github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
```diff
@@ -1881,6 +1881,42 @@ presubmits:
...
+        - make lint
+        image: quay.io/kubevirtci/bootstrap:v20220512-78048f1
+        name: ""
+        resources:
+          requests:
+            memory: 6Gi
+        securityContext:
+          privileged: true
+      nodeSelector:
+        type: bare-metal-external
```
xpivarc, line 1919: We should use `cluster: ibm-prow-jobs`

github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
```diff
@@ -1881,6 +1881,42 @@ presubmits:
...
+        - /bin/sh
+        - -c
+        - make lint
+        image: quay.io/kubevirtci/bootstrap:v20220512-78048f1
+        name: ""
+        resources:
+          requests:
+            memory: 6Gi
+        securityContext:
+          privileged: true
```
xpivarc, line 1917: Do we need this?

github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
```diff
@@ -1881,6 +1881,42 @@ presubmits:
           privileged: true
       nodeSelector:
         type: bare-metal-external
+  - always_run: true
+    optional: true
+    annotations:
+      fork-per-release: "true"
+      testgrid-dashboards: kubevirt-presubmits
```
xpivarc, line 1888: Do we need this?

xpivarc (changes requested):

github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
brianmcarey: Only really need this for docker in docker setups - i think this can be removed.

github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
rhrazdil: Since the lane doesn't atm run Go code either, I have removed the othe labels as well for now.
[Done](https://github.com/kubevirt/project-infra/compare/50af0d89e409b40261dddee10c90afe9299efe13..498d2712f9cba2c5ea9647c5cd468cd14f3b0ded)

github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
rhrazdil: [Removed.](https://github.com/kubevirt/project-infra/compare/50af0d89e409b40261dddee10c90afe9299efe13..498d2712f9cba2c5ea9647c5cd468cd14f3b0ded)

github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
rhrazdil: Probably not. The idea is to provide early feedback to contributors and to sort of foramalize
clean code standards in kubevirt. We don't need to report to the dashboard. [Removed](https://github.com/kubevirt/project-infra/compare/50af0d89e409b40261dddee10c90afe9299efe13..498d2712f9cba2c5ea9647c5cd468cd14f3b0ded) the annotation.

github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
rhrazdil: I think the answer is yes, [Done](https://github.com/kubevirt/project-infra/compare/50af0d89e409b40261dddee10c90afe9299efe13..498d2712f9cba2c5ea9647c5cd468cd14f3b0ded)

github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
```diff
@@ -1881,6 +1881,32 @@ presubmits:
           privileged: true
       nodeSelector:
         type: bare-metal-external
+  - always_run: true
```
EdDev, line 1884: It should always run, but I think we should start as optional first.
I would also prefer to see all the options explicitly set, at least those that have direct implications on the visibility and gateway (e.g. `optional`, `skip_report`).

github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
```diff
@@ -1881,6 +1881,32 @@ presubmits:
...
+    name: pull-kubevirt-code-lint
+    skip_branches:
+    - release-\d+\.\d+
+    spec:
+      containers:
+      - command:
+        - /usr/local/bin/runner.sh
+        - /bin/sh
+        - -c
+        - make lint
```
enp0s3, line 1902: @rhrazdil I think this line won't work because the `lint` word would be cut. Please use the syntax like in [this](https://github.com/kubevirt/project-infra/blob/d7cf80b4872a241a7bffca05e1f018249c08ad6d/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml#L1838) line

enp0s3 (changes requested):
@rhrazdil Hi. Thank you for the PR, overall looks good. Can you please address @EdDev comments and mine below?

rhrazdil:
/hold
To give more time to provide feedback

github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
rhrazdil: [Done](https://github.com/kubevirt/project-infra/compare/498d2712f9cba2c5ea9647c5cd468cd14f3b0ded..83308dc3a07b358862c995faa9e6ec6201879276)

Added `optional: true` and `skip_report:true` explicitly

github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
rhrazdil: You're right, thanks! [Fixed](https://github.com/kubevirt/project-infra/compare/498d2712f9cba2c5ea9647c5cd468cd14f3b0ded..83308dc3a07b358862c995faa9e6ec6201879276)

github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
```diff
@@ -1881,6 +1881,35 @@ presubmits:
...
+    decorate: true
+    decoration_config:
+      grace_period: 5m0s
+      timeout: 1h0m0s
+    max_concurrency: 11
+    name: pull-kubevirt-code-lint
+    optional: true
+    skip_branches:
+    - release-\d+\.\d+
+    skip_report: true
```
oshoval, line 1897: why `skip_report` true ? 
we do want to see the result on the PR, `optional` true is enough

btw imho we should not add fields that have the default value, this is the common practice in project-infra
but this is other story

EdDev: > why `skip_report` true ? we do want to see the result on the PR, `optional` true is enough

I agree.

> btw imho we should not add fields that have the default value, this is the common practice in project-infra but this is other story, it is confusing and potentially leading to mistakes (i.e copy paste)

I disagree
You are requiring me to remember the defaults instead of seeing the settings ecplicitly.

oshoval: > > btw imho we should not add fields that have the default value, this is the common practice in project-infra but this is other story, it is confusing and potentially leading to mistakes (i.e copy paste)
> 
> I disagree You are requiring me to remember the defaults instead of seeing the settings ecplicitly.

Imo, it isn't a sufficient reason for dragging those fields with us again and again.
We removed default fields at the past from jobs.
It worth that we will understand what are the defaults instead.
One can go to the main presubmits of kubevirt and see their config in case he need to remember what
they are for.
For example both `skip_report` and `optional` default is false, we can learn it from `pull-kubevirt-e2e-k8s-1.23-sig-network`, otherwise it won't consider it gating or won't show it on the PR.

enp0s3: I think this topic doesn't have absolute truth. Its rather opinionated one. For me, in some cases defaults may be an overhead, in other cases I would like to see explicit definitions although they are trivial. If we want we can define a policy on how to write a proper Prow config and add it to the project doc, and add some sanity check on the syntax. But I don't see this discussion about the defaults as a blocker.

github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
```diff
@@ -1881,6 +1881,35 @@ presubmits:
           privileged: true
       nodeSelector:
         type: bare-metal-external
+  - always_run: true
+    annotations:
+      fork-per-release: "true"
+    cluster: prow-workloads
```
enp0s3, line 1887: @rhrazdil here it should be `cluster: ibm-prow-jobs`

github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
```diff
@@ -1881,6 +1881,35 @@ presubmits:
...
+        - -c
+        - |
+          make lint
+        image: quay.io/kubevirtci/bootstrap:v20220512-78048f1
+        name: ""
+        resources:
+          requests:
+            memory: 4Gi
+      nodeSelector:
+        type: ibm-prow-jobs
```
enp0s3, line 1912: @rhrazdil This one is redundant because you will need to define `cluster: ibm-prow-jobs`

github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
rhrazdil: > why skip_report true ?
> we do want to see the result on the PR, optional true is enough

[Fixed](https://github.com/kubevirt/project-infra/compare/83308dc3a07b358862c995faa9e6ec6201879276..5dae944164a57c5c72d9a42c9769cb7fb1b2cb25) thanks, I thought for some reason that this knob was related to the testgrid-dashboard.

enp0s3 (changes requested):
@rhrazdil Sorry for bothering you, perhaps we have confused you a little bit. The jobs that we want to run on the `ibm-prow-jobs` cluster, we don't need to define `nodeSelector` but just use the `cluster: ibm-prow-jobs`, that's what I understood from going over the presubmit jobs configuration

github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
rhrazdil: Right, got it, thanks.
[Removed.](https://github.com/kubevirt/project-infra/compare/5dae944164a57c5c72d9a42c9769cb7fb1b2cb25..fbabfb67e773c6b1d7750ebaec719c181c69afbd)

github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
rhrazdil: [Updated](https://github.com/kubevirt/project-infra/compare/fbabfb67e773c6b1d7750ebaec719c181c69afbd..bcbd7937925cb334db92c2e5607a960ba17d1049)

brianmcarey (approved):
/approve

PR #2135, opened 2022-05-30
Update ci search alert description to better reflect error

brianmcarey:
The alert triggers when there is a high rate of failed writes occurring
in the logs of the ci-search pods. This can occur when the persistent
volume is full but it can also occur in other scenarios.

This updates the description to better reflect the error that is occurring.

/cc @dhiller @enp0s3 


dhiller (approved):
/approve

PR #2139, opened 2022-06-01
Register bridge-marker 0.9.1 to CI

phoracek:
New stable branch was created. This time it is awkwardly on a patch release due to breakage in Go version dependency introduced in 0.9.2.

Signed-off-by: Petr Horáček <phoracek@redhat.com>

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/bridge-marker/bridge-marker-presubmits-0.9.1.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/bridge-marker/bridge-marker-presubmits-0.9.1.yaml
@@ -25,7 +25,7 @@ presubmits:
               privileged: true
             resources:
               requests:
-                memory: "29Gi"
+                memory: "14Gi"
             command:
               - "/usr/local/bin/runner.sh"
               - "/bin/sh"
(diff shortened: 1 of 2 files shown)


github/ci/prow-deploy/files/jobs/kubevirt/bridge-marker/bridge-marker-presubmits-0.9.1.yaml
```diff
@@ -0,0 +1,33 @@
...
+      spec:
+        nodeSelector:
+          type: bare-metal-external
+        containers:
+          - image: quay.io/kubevirtci/bootstrap:v20220512-78048f1
+            securityContext:
+              privileged: true
+            resources:
+              requests:
+                memory: "29Gi"
```
brianmcarey, line 28: From what I can see - this job uses the default values for bringing up the 1.23 k8s provider from kubevirtci. I think this job could use a memory request amount similar to the check-provision jobs (14Gi) - this could help with getting the prow job scheduled quicker?

brianmcarey (commented):
Looks good - just one question on the memory request inline.

github/ci/prow-deploy/files/jobs/kubevirt/bridge-marker/bridge-marker-presubmits-0.9.1.yaml
phoracek: I see no reason not to, let me update the PR

brianmcarey (approved):
Thanks @phoracek 

/lgtm

PR #2161, opened 2022-06-16
kubevirt/presubmits/lint: Add privileged security context

EdDev:
The lint job runs kubevirt `lint` target (`make lint`) which uses the
dockerized wrapper.

In order to run the docker service, there is a need for privileged
access.


brianmcarey (commented):
/lgtm

EdDev:
@brianmcarey , @enp0s3 , can we override this unrelated `pull-project-infra-prow-deploy-test` job?
It does not run for 6h, which is very odd IMO. But it is also not relevant to the change here.

brianmcarey:
Yes this job shouldn't be impacted by this change.

PR #2163, opened 2022-06-17
Fix windows priority class by adding to prowjob spec

brianmcarey:
priorityClassName has to be added to the spec section in the prowjob
definition or it does not get applied correctly.

/cc @dhiller


dhiller (approved):
/approve

PR #2166, opened 2022-06-17
Increase node memory for perf jobs

rthallisey:
Increase node memory by 1Gb per node.


xpivarc (approved):

maiqueb:
Would you explain why the nodes now require more memory ? I'm curious.

rthallisey:
@maiqueb we're still seeing https://github.com/kubevirt/kubevirt/issues/7658 so kubevirt node memory needs to be increased.

PR #2168, opened 2022-06-21
cgroups: Make lanes non optional

oshoval:
Since cgroups jobs are conditional,
they have implicit `optional: true` added.
Once they run they run as non required.
Make them required by setting `optional: false`.

One of the jobs had even an explicit optional field that was updated.

Add also skip_report: false, to be explicit.

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
@@ -193,6 +193,9 @@ presubmits:
       nodeSelector:
         type: bare-metal-external
   - always_run: false
+    run_if_changed: pkg/virt-handler/cgroup/.*
+    optional: false
+    skip_report: false
     annotations:
       fork-per-release: "true"
       testgrid-dashboards: kubevirt-presubmits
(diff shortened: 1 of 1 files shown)


oshoval:
Note: Once it is merged I will check if it works as expected, since I can't rehearse this change.

iholder-redhat:
/lgtm
Thanks @oshoval!

Small note: I see you moved some fields around. If https://github.com/kubevirt/project-infra/pull/2169 gets in before this PR we need to rebase and be sure new folders are included

oshoval:
> /lgtm Thanks @oshoval!
> 
> Small note: I see you moved some fields around. If #2169 gets in before this PR we need to rebase and be sure new folders are included

Ye, will do whats need to be done ;]
Thanks

oshoval:
rebased

brianmcarey:
It should work as expected - we do something similar with project-infra presubmits - https://github.com/kubevirt/project-infra/blob/main/github/ci/prow-deploy/files/jobs/kubevirt/project-infra/project-infra-presubmits.yaml#L22-L24

brianmcarey (approved):
/approve

oshoval:
> It should work as expected - we do something similar with project-infra presubmits - https://github.com/kubevirt/project-infra/blob/main/github/ci/prow-deploy/files/jobs/kubevirt/project-infra/project-infra-presubmits.yaml#L22-L24

You mean with this PR right ?
because it doesn't work yet as expected, note that from what i understood from prow team, since we are adding `always_run: false`, an implicit `optional: true` is added.

brianmcarey:
> 
> You mean with this PR right ? because it doesn't work yet as expected, note that from what i understood from prow team, since we are adding `always_run: false`, an implicit `optional: true` is added.

Yes with this PR

oshoval:
Thanks

note that we might want to do this change on more jobs, as they also suffer from the same thing

PR #2174, opened 2022-06-29
Add optional CDI Ceph lane with DV GC enabled

arnongilboa:


awels:
So we want to keep setting env variables out of project-infra and instead put them in a script that is called in the CDI repo. Now if we change something we have 2 PRs we need to do instead of just updating the CDI repo. Can you do with the matching PR in CDI create a new script that sets up the environment variable and then calls the other script? And change this to just call your new script?

PR #2175, opened 2022-06-29
kubevirtci kubevirt bump: Increase periodic rate

oshoval:
Increase to every 3h instead every 12h


dhiller (commented):
/approve

PR #2192, opened 2022-07-07
kubevirt: Add manual lane to bump kubevirtci at kubevirt

oshoval:
Sometimes we want to refresh kubevirtci bumps at kubevirt manually.
This lane can be used for it.
The lane doesn't check labels of the bump PR if exists, as the periodic
does.

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
@@ -241,6 +241,38 @@ presubmits:
     decorate: true
     decoration_config:
       grace_period: 5m0s
+      timeout: 1h
+    labels:
+      preset-dind-enabled: "true"
+      preset-docker-mirror: "true"
+      preset-gcs-credentials: "true"
+      preset-github-credentials: "true"
+    max_concurrency: 1
+    name: pull-kubevirtci-bump-kubevirt
+    skip_branches:
+    - release-\d+\.\d+
+    spec:
+      containers:
+      - image: quay.io/kubevirtci/pr-creator:v20210920-0ef32ec
+        command: ["/bin/sh", "-ce"]
+        args:
+        - |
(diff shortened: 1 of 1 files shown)


oshoval:
The image we are using doesn't have the latest git-pr.sh

oshoval:
@dhiller 
Can you please take another look ?

dhiller:
> The image we are using doesn't have the latest git-pr.sh (the one on the project-infra repo)
> 
> We can use if desired (but i guess it is not a good idea, and will also be different from what the periodic does)
> 
> ```
>           curl -LO https://raw.githubusercontent.com/kubevirt/project-infra/main/hack/git-pr.sh
>           chmod 755 git-pr.sh
> ```
> 
> Changed to use the one in the image, and remove the `-p` as the default is pwd which is right (already kubevirt) so the `cd` won't fail

IIRC the situation for presubmits is different as there's always the mounted checkout whereas in periodics you always have to say what you want to attach using `extra_refs`

dhiller:
@oshoval I have updated the image builds for pr-creator, obviously the run_if_changed missed the other pathes for `hack/` and `label-checker`: #2198

oshoval:
> IIRC the situation for presubmits is different as there's always the mounted checkout whereas in periodics you always have to say what you want to attach using `extra_refs`

IIUC It uses `git-pr.sh` not `hack/git-pr.sh` so it takes the one of `/usr/local/bin` of the container image
right ?

what holds us from merging this PR, having the new image on both this and the periodic ?

dhiller:
> > IIRC the situation for presubmits is different as there's always the mounted checkout whereas in periodics you always have to say what you want to attach using `extra_refs`
> 
> IIUC It uses `git-pr.sh` not `hack/git-pr.sh` so it takes the one of `/usr/local/bin` of the container image right ?
> 
> what holds us from merging this PR, having the new image on both this and the periodic ?

Yeah, I was wanting to ask whether you wanted to update the image, but we can merge this one anyway. We can do the rest separately, you are right.

/approve

oshoval:
Thanks, would you lgtm as well please ?

brianmcarey (commented):
/lgtm

PR #2193, opened 2022-07-07
Make pull-kubevirt-verify-go-mod lane required

iholder-redhat:
Since people sometime forget to `make deps-sync`, this lane sometimes fails constantly just because of human-error. Instead, this lane should be required to ensure our dependencies are always synced.

As in [this comment](https://github.com/kubevirt/kubevirt/pull/8062#issuecomment-1176180982), the reason this lane was optional is because of a worry that this lane may be unstable since it requires access to quite some external resources. But since the lane proved itself being stable, it can be required now.

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
@@ -933,7 +933,6 @@ presubmits:
       preset-dind-enabled: "true"
       preset-docker-mirror-proxy: "true"
     name: pull-kubevirt-verify-go-mod
-    optional: true
     run_if_changed: ^vendor/.*|^staging/.*|go\.mod$|go\.sum$
     skip_branches:
     - release-\d+\.\d+


github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
```diff
@@ -937,7 +937,6 @@ presubmits:
       preset-dind-enabled: "true"
       preset-docker-mirror-proxy: "true"
     name: pull-kubevirt-verify-go-mod
-    optional: true
```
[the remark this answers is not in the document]
iholder-redhat: Makes sense, thanks!

github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
```diff
@@ -933,7 +933,6 @@ presubmits:
...
     run_if_changed: ^vendor/.*|^staging/.*|go\.mod$|go\.sum$
```
oshoval, line 936: oops my bad, since we have this you can't / don't need to set always_run

iholder-redhat: Changed again. Anything else?

oshoval (approved):
Thanks

github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
rmohr: This line needs to be removed. Otherwise the job is not required. We can only have one of the two: either required, or selective run.

oshoval: @rmohr IIRC there is a bug with prow that it won't show it as required in the list at the PR bottom
but if it fails, it should show you "required" in the table of the comment that it adds with the fails summary

rmohr: it will not be considered in the merge queue as required (tide) and also not on the github page. The meaning of optional and run_if_changes are a little bit fuzzy, agreed, but it is not required in the parts which we care about :)

rmohr: @oshoval here the context: https://github.com/kubevirt/kubevirt/pull/8467

PR #2196, opened 2022-07-08
Add podman to bootstrap images

maya-r:
This is so we can merge the code in https://github.com/kubevirt/hostpath-provisioner/pull/115
(The golang image used by hostpath-provisioner presubmits is based on the bootstrap image)


brianmcarey (approved):
Sorry for the delay in looking at this - I wanted to make sure it wouldn't cause any confusion for the e2e workloads. Tested it locally with phaino and it looks good. 

/approve

PR #2220, opened 2022-07-21
manual bump job: Fix Running job from within kubevirt

oshoval:
In case we run the manual job `pull-kubevirtci-bump-kubevirt` from the bump PR itself
it will update the description with just the new differences
since the previous bump if exists.
Because the changes are already there and it calculate the difference.

In case we run it from a PR of kubevirt that is not related,
It will include in the bump those commits which are not related,
since it checkouts the code of that PR.

Adding reset --hard to kubevirt main in order to avoid those.


brianmcarey (commented):
/lgtm

brianmcarey (approved):
/approve

PR #2231, opened 2022-07-31
container images: Mirror needed images to quay

oshoval:
Those images are used by kubevirtci and are still on `docker.io`.
Mirror them to quay, so we can use them without rate limits.

Once they are mirrored, we need one time to make the new repos public.


enp0s3 (commented):
/approve

oshoval:
Thanks @enp0s3 
Can you lgtm as well please ?

enp0s3 (approved):

PR #2232, opened 2022-08-01
Single-node increase timeout

xpivarc:
This periodic seems to timeout
regularly. Let's see if an additional
hour will be enough.


maiqueb (approved):
Fingers crossed ...

PR #2233, opened 2022-08-01
capk: sonobuoy conformance lane

rhrazdil:


davidvossel (commented):
what would you think about using the `optional: false` setting initially. Just until we feel comfortable that this set of tests is stable?

rhrazdil:
>what would you think about using the optional: true setting initially. Just until we feel comfortable that this set of tests is stable?

Sure, added the optional attribute.

rhrazdil:
I'm not sure why the rehearsal job isn't running now

davidvossel (approved):
/lgtm
/approve

PR #2235, opened 2022-08-02
Add presubmit unit test job for csi-driver

brybacki:
Add a job that runs unit tests

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/csi-driver/csi-driver-presubmits.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/csi-driver/csi-driver-presubmits.yaml
@@ -1,7 +1,7 @@
 presubmits:
   kubevirt/csi-driver:
     - name: pull-csi-driver-unit-test
-      cluster: prow-workloads
+      cluster: ibm-prow-jobs
       skip_branches:
         - release-\d+\.\d+
       annotations:
--- a/github/ci/prow-deploy/files/jobs/kubevirt/csi-driver/csi-driver-presubmits.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/csi-driver/csi-driver-presubmits.yaml
@@ -1,5 +1,41 @@
 presubmits:
   kubevirt/csi-driver:
+    - name: pull-csi-driver-unit-test
+      cluster: prow-workloads
+      skip_branches:
+        - release-\d+\.\d+
+      annotations:
+        fork-per-release: "true"
+        testgrid-create-test-group: "false"
+      always_run: true
+      optional: false
+      skip_report: false
+      decorate: true
+      decoration_config:
+      timeout: 1h
+      grace_period: 5m
+      max_concurrency: 11
+      labels:
+        preset-dind-enabled: "true"
+        preset-docker-mirror-proxy: "true"
+        preset-shared-images: "true"
+      spec:
+        containers:
+          - image: quay.io/kubevirtci/golang:v20220630-ded5dcd
+            command:
+              - "/usr/local/bin/runner.sh"
+              - "/bin/sh"
+              - "-c"
+              - "make test"
+            env:
+              - name: GIMME_GO_VERSION
+                value: "1.18"
+            # docker-in-docker needs privileged mode
+            securityContext:
+              privileged: true
+            resources:
+              requests:
+                memory: "4Gi"
     - name: pull-csi-driver-e2e-k8s-1.20
       skip_branches:
         - release-\d+\.\d+


brybacki:
Is that allright?
Which cluster should I select for unit test job:
      cluster: prow-workloads
      cluster: ibm-prow-jobs

github/ci/prow-deploy/files/jobs/kubevirt/csi-driver/csi-driver-presubmits.yaml
```diff
@@ -1,5 +1,41 @@
 presubmits:
   kubevirt/csi-driver:
+    - name: pull-csi-driver-unit-test
+      cluster: prow-workloads
```
awels, line 4: I would use ibm-prow-jobs

github/ci/prow-deploy/files/jobs/kubevirt/csi-driver/csi-driver-presubmits.yaml
```diff
@@ -1,5 +1,41 @@
...
+        preset-shared-images: "true"
+      spec:
+        containers:
+          - image: quay.io/kubevirtci/golang:v20220630-ded5dcd
+            command:
+              - "/usr/local/bin/runner.sh"
+              - "/bin/sh"
+              - "-c"
+              - "make test"
+            env:
```
awels, line 30: maybe don't put the env variable in here, and maybe put those in a script in the repo. This allows us to change the go version without having to change the project-infra presubmit

davidvossel: is there an example of another project doing this?

awels: Yes, HPP for instance. https://github.com/kubevirt/hostpath-provisioner/blob/main/hack/test.sh#L23-L27 it allows you to set it through env variables, but defaults to the value set in the script. This is about not having to modify the PROW job for changes to library versions (k8s) or go versions. CDI does something similar for just the k8s version and not the golang version.

github/ci/prow-deploy/files/jobs/kubevirt/csi-driver/csi-driver-presubmits.yaml
brybacki: ok, and what is the difference?

awels: The prow-workloads is a cluster full of BM machines that are used for the kubevirtci tests that need a kubevirtci cluster to be created. The ibm-prow-jobs is not bare metal, so we should run jobs that don't need BM there to not use capacity on the BM cluster.

brybacki: good idea

github/ci/prow-deploy/files/jobs/kubevirt/csi-driver/csi-driver-presubmits.yaml
brybacki: The hostpath unit test script has some strange command for tests, is that all needed?

https://github.com/kubevirt/hostpath-provisioner/blob/main/hack/run-unit-test.sh
```

if [[ -v PROW_JOB_ID ]] ; then
  useradd prow -s /bin/bash
  chown prow:prow -R /home/prow
  echo "Run go test -v in $PWD"
  sudo -i -u prow /bin/bash -c 'cd /home/prow/go/src/github.com/kubevirt/hostpath-provisioner && /usr/local/go/bin/go test -v ./cmd/... ./controller/... ./pkg/...'
 go get -u golang.org/x/lint/golint
else
```

awels: Actually I linked the wrong file, this is the right one. https://github.com/kubevirt/hostpath-provisioner/blob/main/hack/common.sh#L17-L22 and the call to this function https://github.com/kubevirt/hostpath-provisioner/blob/main/hack/run-unit-test.sh#L20 all of that stuff is to run the unit tests inside a container.

brybacki: can I experiment with this lane on csi-kubevirt - while still in-progress?

awels: No, this has to be merged before you can call the lane. Maybe it makes sense to leave the env variable for now, and see how cumbersome it is to switch golang versions. It might happen infrequent enough that this suggestion is over engineering.

github/ci/prow-deploy/files/jobs/kubevirt/csi-driver/csi-driver-presubmits.yaml
brybacki: done

PR #2236, opened 2022-08-02
mirror: Update istio images naming so we can use them with istio operator

oshoval:
Istio is deployed with Istio operator, which can only be configured to use different registry repository
but will always use container image names: pilot, proxyv2, etc.
Therefore the images should not have the "istio-" prefix
this way the istio operator will use `quay.io/kubevirtci/pilot:v1` instead of `docker.io/istio/pilot:v1`
see https://github.com/kubevirt/kubevirtci/pull/851

Additional changes:
According https://github.com/rook/rook/issues/7213
docker.io/rook/ceph doesn't have rate limits.
We are also using its docker.io version, not the quay one.
Removing it from the mirroring list.

### DIFF
--- a/hack/images_to_mirror.csv
+++ b/hack/images_to_mirror.csv
@@ -1,6 +1,4 @@
 docker.io,library/registry:2.7.1,quay.io/kubevirtci
-docker.io,rook/ceph:v1.5.8,quay.io/kubevirtci
-docker.io,rook/ceph:v1.8.9,quay.io/kubevirtci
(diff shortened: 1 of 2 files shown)


enp0s3 (commented):
/approve

enp0s3 (approved):

oshoval:
Thanks Igor

@jean-edouard can you please make those public ?

quay.io/kubevirtci/install-cni
quay.io/kubevirtci/operator
quay.io/kubevirtci/pilot
quay.io/kubevirtci/proxyv2

PR #2243, opened 2022-08-05
Fix and enable release blocker plugin

xpivarc:
/cc @brianmcarey @rmohr

### DIFF
--- a/external-plugins/release-blocker/main.go
+++ b/external-plugins/release-blocker/main.go
@@ -64,16 +64,15 @@ func main() {
 	logrus.SetLevel(logrus.DebugLevel)
 	log := logrus.StandardLogger().WithField("plugin", pluginName)
 
-	secretAgent := &secret.Agent{}
-	if err := secretAgent.Start([]string{o.github.TokenPath, o.webhookSecretFile}); err != nil {
+	if err := secret.Add(o.github.TokenPath, o.webhookSecretFile); err != nil {
 		logrus.WithError(err).Fatal("Error starting secrets agent.")
 	}
 
-	githubClient, err := o.github.GitHubClient(secretAgent, o.dryRun)
+	githubClient, err := o.github.GitHubClient(o.dryRun)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


github/ci/prow-deploy/kustom/base/configs/current/plugins/plugins.yaml
```diff
@@ -427,11 +427,13 @@ external_plugins:
     events:
       - pull_request
   - name: cherrypicker
+  - name: release-blocker
```
brianmcarey, line 430: Do we need to build a new image and update the deployment manifest before adding this back in? 

https://github.com/kubevirt/project-infra/blob/f96e43d0b8d6aec90ef48b8b02b49b3266d9be74/github/ci/prow-deploy/kustom/base/manifests/local/release-blocker_deployment.yaml#L20

xpivarc:
/hold
@brianmcarey good point. Thanks!

external-plugins/release-blocker/main.go
```diff
@@ -98,7 +98,7 @@ func main() {
 
 	server := &Server{
 		tokenGenerator: secretAgent.GetTokenGenerator(o.webhookSecretFile),
-		botName:        botUserData.Name,
+		botName:        botUserData.Login,
```
rmohr, line 101: big impact :)

rmohr (commented):
Like @brianmcarey pointed out, we need a new image first. Fix looks great.

xpivarc:
@brianmcarey Can I ask you for the image push? (I don't have access)

xpivarc:
/hold 
I don't see the error locally. Will have a look.

brianmcarey:
> @brianmcarey Can I ask you for the image push? (I don't have access)

I've built and pushed the updated release blocker image - `quay.io/kubevirtci/release-blocker:v20220815-ab4bb7e6`

xpivarc:
/hold cancel
@brianmcarey Thanks!

brianmcarey (commented):
/lgtm

PR #2248, opened 2022-08-11
Enable nonroot sriov

xpivarc:
/cc @brianmcarey @dhiller

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
@@ -426,7 +426,7 @@ presubmits:
       nodeSelector:
         type: bare-metal-external
       priorityClassName: windows
-  - always_run: false
+  - always_run: true
     annotations:
       fork-per-release: "true"
       k8s.v1.cni.cncf.io/networks: multus-cni-ns/sriov-passthrough-cni,multus-cni-ns/sriov-passthrough-cni
@@ -443,7 +443,7 @@ presubmits:
       sriov-pod: "true"
     max_concurrency: 10
     name: pull-kubevirt-e2e-kind-1.22-sriov-nonroot
-    optional: true
+    optional: false
     skip_branches:
     - release-\d+\.\d+
     spec:
--- a/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
@@ -508,7 +508,7 @@ presubmits:
           path: /dev/vfio/
           type: Directory
         name: vfio
-  - always_run: true
+  - always_run: false
     annotations:
       fork-per-release: "true"
       k8s.v1.cni.cncf.io/networks: multus-cni-ns/sriov-passthrough-cni,multus-cni-ns/sriov-passthrough-cni
@@ -527,6 +527,7 @@ presubmits:
       sriov-pod: "true"
     max_concurrency: 10
     name: pull-kubevirt-e2e-kind-1.22-sriov
+    optional: true
     skip_branches:
     - release-\d+\.\d+
     spec:


xpivarc:
FYI @EdDev @ormergi

dhiller (commented):
Looking at the history, I see it failing sometimes: https://prow.ci.kubevirt.io/job-history/gs/kubevirt-prow/pr-logs/directory/pull-kubevirt-e2e-kind-1.22-sriov-nonroot

Also I don't see any data on the flakefinder reports: https://storage.googleapis.com/kubevirt-prow/reports/flakefinder/kubevirt/kubevirt/flakefinder-2022-08-10-168h.html

So are we sure that it will work every time and not confuse users?

xpivarc:
> Looking at the history, I see it failing sometimes: https://prow.ci.kubevirt.io/job-history/gs/kubevirt-prow/pr-logs/directory/pull-kubevirt-e2e-kind-1.22-sriov-nonroot
> 
> Also I don't see any data on the flakefinder reports: https://storage.googleapis.com/kubevirt-prow/reports/flakefinder/kubevirt/kubevirt/flakefinder-2022-08-10-168h.html
> 
> So are we sure that it will work every time and not confuse users?

The lane was disabled due to CI issues(well this is my understanding). We don't have any data because it was disabled for 4/5 months and nobody cared about this or didn't know this happened.

This is a huge step back as we want to default nonroot in Kubevirt and we should definitely test this. This PR is just fixing the current state.

EdDev:
My original opinion on the number of jobs we run has not changed.
I find it unrealistic to run such a large number of jobs, attempting to cover a never ending matrix of setup options.

The SR-IOV job is also heavily dependent on HW resources.
We do not run it on more than one K8S version and do not test special setup cases (e.g. single stack).
I see no special reason to add this as a presubmit job.

I would recomend adding it as a periodic and switch to a presubmit once we formally release the rootless feature.
At which time I would expect wither to stop testing the rootfull one or move it to a periodic as well.

xpivarc:
> My original opinion on the number of jobs we run has not changed. I find it unrealistic to run such a large number of jobs, attempting to cover a never ending matrix of setup options.
> 
> The SR-IOV job is also heavily dependent on HW resources. We do not run it on more than one K8S version and do not test special setup cases (e.g. single stack). I see no special reason to add this as a presubmit job.
> 
> I would recomend adding it as a periodic and switch to a presubmit once we formally release the rootless feature. At which time I would expect either to stop testing the rootfull one or move it to a periodic as well.

I don't see why we should be running rootfull mode rather than rootless. I can disable it(rootfull) if you think we have too many jobs.

EdDev:
> I don't see why we should be running rootfull mode rather than rootless. I can disable it(rootfull) if you think we have too many jobs.

We should disable/enable based on what we support in our deployments today (by default).
Or if you prefer, what we are supposed to support today, that is what needs to be tested.
If we have declared the rootless option as our favorite one we support, that is what needs to run indeed, if we are just planning to do that once other things are resolved, then I guess we should stick with the rootful one.

xpivarc:
@EdDev Done

EdDev:
@xpivarc , thank you.
Could you please elaborate in the PR/commit about the reasoning?

If I understand correctly, we expect the next release to be deployed with rootless virt-launchers by default, right?

dhiller:
> > Looking at the history, I see it failing sometimes: https://prow.ci.kubevirt.io/job-history/gs/kubevirt-prow/pr-logs/directory/pull-kubevirt-e2e-kind-1.22-sriov-nonroot
> > Also I don't see any data on the flakefinder reports: https://storage.googleapis.com/kubevirt-prow/reports/flakefinder/kubevirt/kubevirt/flakefinder-2022-08-10-168h.html
> > So are we sure that it will work every time and not confuse users?
> 
> The lane was disabled due to CI issues(well this is my understanding). We don't have any data because it was disabled for 4/5 months and nobody cared about this or didn't know this happened.
> 
> This is a huge step back as we want to default nonroot in Kubevirt and we should definitely test this. This PR is just fixing the current state.

Makes sense, thanks for the clarification.

/approve

/hold to wait for the discussion to settle

Unhold when that is done. Thank you!

brianmcarey (commented):
/lgtm

EdDev (commented):
Looking at the `automation/test.sh`, when specifying the `KUBEVIRT_NONROOT` env, it will run the tests under `tests/nonroot_test.go`.
Why do we need to run these tests in the context of SR-IOV?

EdDev:
Please sync with the sig-network maintainer before proceeding with merging this change.
I think it is appropriate to ack such changes only after informing the relevant SIG and getting its approve, otherwise, there will be a support issue. To me, it seems also like a project wide decision (i.e. declaring that our main support is now focuses on the non-root deployment). If we have not sent a wide notification about our focus change, we should do it.

Except of the above, I am all good to start focusing in the non-root deployment.
And as a following step, to drop the root-ful lanes (at least the network ones).
I am voting against supporting both, not only due to CI load, but due to engineering resources.

xpivarc:
> 

@AlonaKaplan PTAL

@EdDev The notification was sent and I also raised it at the community meetings if I remember correctly. There was no negative response.

AlonaKaplan:
> > 
> 
> @AlonaKaplan PTAL
> 
> @EdDev The notification was sent and I also raised it at the community meetings if I remember correctly. There was no negative response.

I'm ok with the change. But it should be documented in the sriov documentation that it is supported only on nonroot VMs.

ormergi:
I am afraid that as soon as the non-root jobs will be executed the same frequency as  root-full it may introduce more flakes due to the nature of this complex setup, yet I don't have an idea to prevent it, just sharing my concerns.
Have you tackled some issues with non-root before when running SR-IOV tests we should know about?

I have to agree with Edy about supporting both (root and non-root), we merely have enough resources to run and support one of them. I suggest setting only one of them as a voting lane.

xpivarc:
> > > 
> > 
> > 
> > @AlonaKaplan PTAL
> > @EdDev The notification was sent and I also raised it at the community meetings if I remember correctly. There was no negative response.
> 
> I'm ok with the change. But it should be documented in the sriov documentation that it is supported only on nonroot VMs.

/hold cancel
Can you point me to this documentation?

AlonaKaplan:
> /hold cancel Can you point me to this documentation?

https://github.com/kubevirt/user-guide/blob/main/docs/virtual_machines/interfaces_and_networks.md#sriov

PR #2249, opened 2022-08-11
Skip large e2e tests when only docs are updated

brianmcarey:
Currently the full set of e2e presubmits run for simple doc updates.
This can be quite costly in CI time and should not be a requirement for
documentation updates.

Using the skip-if-only-changed configuration[1] to skip e2e jobs when there
is only updates to files under docs/ or only to .md files.

[1] https://github.com/kubernetes/test-infra/blob/master/prow/jobs.md#triggering-jobs-based-on-changes

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
@@ -136,6 +136,7 @@ presubmits:
     name: pull-kubevirt-e2e-k8s-1.23-sig-compute-nonroot
     skip_branches:
     - release-\d+\.\d+
+    skip_if_only_changed: "^docs/|\\.md$"
     spec:
       containers:
       - command:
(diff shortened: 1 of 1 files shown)


brianmcarey:
/hold 

I want to suggest this change in the mailing list before merging. Also have to investigate a potential clash with the branch protector. 

It should be ready for initial review though.

/cc @dhiller

rmohr:
I can vaguely remember that this is conflicting then with the branch protector from github since skipped jobs are not reported at all but they are "required" from the github perspective. Would be great if we can clarify that initially.

brianmcarey:
> 
> [The docs say](https://github.com/kubernetes/test-infra/blob/master/prow/jobs.md#protecting-status-contexts): _The branch protection rules will only enforce the presence of jobs that run unconditionally and have required status contexts. As conditionally-run jobs may or may not post a status context to GitHub, they cannot be required through this mechanism._

Ok reading that would suggest that if say `pull-kubevirt-e2e-k8s-1.23-sig-compute-cgroupsv2` failed it would not stop a merge from happening as it is conditionally-run? 
https://github.com/kubevirt/project-infra/blob/53b25a02464502162e1ae7cc06db93d56ffc245b/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml#L19-L20

PR #2251, opened 2022-08-11
Added fossa presubmit job for cdi

awels:
Allow CDI to run fossa check, not required or automatic
for now, until we can configur the entire setup.


dhiller (approved):
/approve

PR #2254, opened 2022-08-16
publish-kubevirtci: Increase timeout

oshoval:
Since we are adding slim mode, increase the timeout,
to have more margin.


brianmcarey (approved):
/approve

As we are going to increase the number of images published by this job (https://github.com/kubevirt/kubevirtci/pull/852) - the runtime of the job is going to increase .

PR #2258, opened 2022-08-17
podman: Use the default podman socket path\

oshoval:
This is also the path that kubevirtci and kubevirt expects.
This way podman itself will be selected, and not docker.


oshoval:
Tested locally, by replacing the script in the current container, the expected socket was created

oshoval:
/cc @brianmcarey please take a look

Tested locally, by replacing the script in the current container, the expected socket was created

brianmcarey (approved):
/approve

oshoval:
Do we really need to build all images as part of `build-multiarch-bootstrap-image` ?
It just build, not publish, and takes around 4h

PR #2259, opened 2022-08-17
Update podman socket path in bootstrap image to match kubevirtci defaults

brianmcarey:
kubevirtci was updated to look for an environment variable
`KUBEVIRTCI_PODMAN_SOCKET` which defaults to `/run/podman/podman.sock` [1]

The bootstrap image should set up the podman socket at the same path as
the default so that it is always found.

[1] https://github.com/kubevirt/kubevirtci/pull/833

/cc @oshoval @dhiller 


brianmcarey:
Duplicates  - https://github.com/kubevirt/project-infra/pull/2258

PR #2265, opened 2022-08-18
Update kubevirt.github.io jobs to pull from repo

cwilkers:

We need ruby and other files from the kubevirt.github.io repo during build.


cwilkers:
/cc @brianmcarey 
It looks like a chicken/egg problem, I can't fix the pre-submit because I can't pass the pre-submit.

dhiller (approved):
/approve

dhiller:
Since the presubmit is not yet required it doesn't prohibit automatic merging, thus we should be good to go.

As a followup we should make the presubmit required (aka optional: false or remove optional in job spec) once it succeeds.

dhiller:
Strange, tide complained about the job failing, maybe there's some config problem. We need to revisit that.

PR #2266, opened 2022-08-19
Replaced CWD reference in COPY

cwilkers:


dhiller (approved):
Tested the presubmit locally with phaino , looks good!

/approve

PR #2267, opened 2022-08-21
kubevirtci, presubmit: Increase max_concurrency

oshoval:
Waiting for results when max_concurrency is 1 takes a lot of time,
increase it to 2.


dhiller (approved):
/approve

PR #2271, opened 2022-08-22
Update multiarch publish jobs to use legacy bootstrap image

brianmcarey:
The version of podman in the latest bootstrap image does not have the
`--amend` flag availabe when using `podman manifest create`. This causes the
publish_multiarch_image.sh to fail [1].

This flag looks like it will be available in the next podman release[2] so
just moving these jobs back to using DinD until the flag is available

[1] https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/logs/publish-bootstrap-image/1561665301555187712
[2] https://github.com/containers/podman/pull/15350

/cc @dhiller 


dhiller (approved):
/approve

PR #2273, opened 2022-08-23
Switch kubevirtci prowjobs to use podman in container setup

brianmcarey:
Update the kubevirtci prowjobs to use the new podman based bootstrap image. 

There is an outstanding issue with running sonobuoy on the kind based lanes that are deployed with podman so these lanes have been updated to use the legacy image for now.


dhiller (approved):
/approve

xpivarc (approved):

brianmcarey:
/override rehearsal-check-provision-k8s-1.25

1.25 provider is not available in kubevirtci yet

PR #2278, opened 2022-08-23
Set always_run: true for DV GC lane

arnongilboa:

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
@@ -243,7 +243,7 @@ presubmits:
           privileged: true
       nodeSelector:
         type: bare-metal-external
-  - always_run: false
+  - always_run: true
     annotations:
       fork-per-release: "true"
       testgrid-dashboards: kubevirt-presubmits


github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
```diff
@@ -260,7 +260,6 @@ presubmits:
       preset-shared-images: "true"
     max_concurrency: 11
     name: pull-kubevirt-e2e-k8s-1.24-sig-storage-dv-gc
-    optional: true
```
dhiller, line 263: I would not yet make it required, it seems like the lane is not yet stable enough: https://prow.ci.kubevirt.io/job-history/gs/kubevirt-prow/pr-logs/directory/pull-kubevirt-e2e-k8s-1.24-sig-storage-dv-gc

Please do not forget that a failing required lane will block merging a PR.

dhiller (changes requested):

github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
arnongilboa: No problem, I'll battle the lane flakiness in https://github.com/kubevirt/kubevirt/pull/8351

dhiller (approved):
/approve

fabiand:
@dhiller did we consider the impact on our ci resources?
Also what was the reasoning for enabling it always?

IOW it's a big change for CI and there is no discussion at all ...

arnongilboa:
> @dhiller did we consider the impact on our ci resources? Also what was the reasoning for enabling it always?
> 
> IOW it's a big change for CI and there is no discussion at all ...

Having GC enabled affects the behavior of all created DVs, so we must have at least one lane with GC.

In the near future, GC will be enabled by default, so this specific lane may get replaced by a single lane with no-GC.

PR #2294, opened 2022-08-25
Keep make in the final image as well as git

cwilkers:

Need make for website build


dhiller (approved):
/approve

PR #2301, opened 2022-08-29
Build CDI builder using buildah

maya-r:
Right now CDI will only use buildah as a build command if podman is also
used, but in that case we won't build multi-arch images.

Force buildah so we can get arm64 images as well.


brianmcarey (approved):
/approve

PR #2318, opened 2022-09-06
Use legacy image for jobs that push images using bazel

brianmcarey:
Pushing images using bazel fails when run from the podman bootstrap
image. It fails with a 400 bad request[1]. Updating these jobs back to
using the legacy image until issue is resolved.

[1] https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/logs/periodic-kubevirt-push-nightly-build-main/1566954990545670144#1:build-log.txt%3A3

/cc @xpivarc @enp0s3 @dhiller 


dhiller (approved):
/approve

Thank you!

PR #2321, opened 2022-09-07
Update kubevirt prowjobs to use the new podman in container setup

brianmcarey:

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-postsubmits.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-postsubmits.yaml
@@ -15,14 +15,14 @@ postsubmits:
       timeout: 3h
       grace_period: 5m
     labels:
-      preset-dind-enabled: "true"
+      preset-podman-in-container-enabled: "true"
       preset-docker-mirror-proxy: "true"
       preset-gcs-credentials: "true"
       preset-github-credentials: "true"
       preset-kubevirtci-quay-credential: "true"
     spec:
       containers:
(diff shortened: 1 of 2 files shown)


brianmcarey:
@xpivarc @dhiller all of the rehearsals passed apart from the following:

`rehearsal-pull-kubevirt-e2e-arm64` - looks to be a clash in the rehearse plugin as the project-infra repo is mentioned as an extra ref in the job

`rehearsal-pull-kubevirt-e2e-k8s-1.23-single-node` - I believe the periodic version of this job has been failing consistently and requires investigation. 

I believe this is good to go in - what do you think?

xpivarc:
/lgtm 
Yes, the single node seems like it needs a little bit of attention.

dhiller:
> /lgtm Yes, the single node seems like it needs a little bit of attention.

I had a look recently, and it ran ok then: https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/logs/periodic-kubevirt-e2e-k8s-1.23-single-node/1573515281160998912
Latest run is bad again. What I am seeing on this is that either it runs in around 3h or it times out after 5h. I think we should remove the lane, since it brings no value. WDYT>

dhiller (approved):
/approve

dhiller:
> > /lgtm Yes, the single node seems like it needs a little bit of attention.
> 
> I had a look recently, and it ran ok then: https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/logs/periodic-kubevirt-e2e-k8s-1.23-single-node/1573515281160998912 Latest run is bad again. What I am seeing on this is that either it runs in around 3h or it times out after 5h. I think we should remove the lane, since it brings no value. WDYT?

https://github.com/kubevirt/project-infra/pull/2367

PR #2341, opened 2022-09-21
Fix yaml format for ThanosCompactHalted alert

brianmcarey:
/cc @dhiller 


dhiller (approved):
/approve

Maybe we should create a presubmit that checks the validity of the rules.yaml?

PR #2343, opened 2022-09-22
Move publish-kubevirtci prowjob back to using dind

brianmcarey:
Clusters have been failing to come up on the hco test lanes[1]. There looks to be an issue with the providers that have been published with podman which breaks scp to the nodes.

[1] https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_hyperconverged-cluster-operator/2092/pull-hyperconverged-cluster-operator-e2e-k8s-1.23/1572554341003300864

/cc @oshoval @dhiller 

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/kubevirtci/kubevirtci-postsubmits.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/kubevirtci/kubevirtci-postsubmits.yaml
@@ -15,7 +15,7 @@ postsubmits:
         repo: project-infra
         base_ref: main
       labels:
-        preset-podman-in-container-enabled: "true"
+        preset-dind-enabled: "true"
         preset-docker-mirror-proxy: "true"
         preset-gcs-credentials: "true"
         preset-github-credentials: "true"
@@ -43,7 +43,7 @@ postsubmits:
             type: Directory
           name: devices
         containers:
-        - image: quay.io/kubevirtci/golang:v20220906-4f60dd3
+        - image: quay.io/kubevirtci/golang-legacy:v20220810-a8f2e6c
           command:
           - "/usr/local/bin/runner.sh"
           - "/bin/bash"


dhiller (approved):
/approve
/hold to get a view of what others think.

Unhold when you see fit.

oshoval:
Thanks
 Do we want instead to use ssh.sh instead of scp ? (the scp is broken when trying to spin with docker a provider that was provisioned with podman)
i.e use something like this instead scp
`./cluster-up/ssh.sh node01 sudo cat /etc/kubernetes/admin.conf`

it works, and also might fix the problem that after the first reboot of the machine,
and provider that was created with docker and cluster-up with podman have the same problem ?

brianmcarey:
> Thanks Do we want instead to use ssh.sh instead of scp ? (the scp is broken when trying to spin with docker a provider that was provisioned with podman) i.e use something like this instead scp `./cluster-up/ssh.sh node01 sudo cat /etc/kubernetes/admin.conf`

We could try this but I still think we need to understand why scp has stopped working in this case as it might have an effect on other test lanes that use the providers. 
> 
> it works, and also might fix the problem that after the first reboot of the machine, a provider that was created with docker and cluster-up with podman has the same problem ?
> 
> creating a PR to show what i mean

Ok thanks - will take a look.

oshoval:
> > Thanks Do we want instead to use ssh.sh instead of scp ? (the scp is broken when trying to spin with docker a provider that was provisioned with podman) i.e use something like this instead scp `./cluster-up/ssh.sh node01 sudo cat /etc/kubernetes/admin.conf`
> 
> We could try this but I still think we need to understand why scp has stopped working in this case as it might have an effect on other test lanes that use the providers.

Tried it, it solved one step but there are more problems there, so lets go with this PR for now
and investigate offline of course whatever needed

oshoval (approved):
Thanks

oshoval:
We also need please to either delete the previous head providers, so it will choose (need to double check)
the news ones you created yetsterday, or to push some commit to kubevirtci so it will re trigger the build
otherwise it still choose the broken providers atm

brianmcarey:
> We also need please to either delete the previous head providers (2209211054-72ad28f), so it will choose (need to double check) the news ones you created yetsterday, or to push some commit to kubevirtci so it will re trigger the build otherwise it still choose the broken providers atm

This would be a good candidate to merge for kubevirtci - https://github.com/kubevirt/kubevirtci/pull/879

oshoval:
Btw while trying to fix it, it seems not only scp is broken but also the API port 6443, port-forwarding of the containers
so the WA I suggested won't work

PR #2366, opened 2022-10-04
Increase thanos compact PVC to 40Gi

brianmcarey:
The compact process is hitting an error due to a full disk which is causing the compaction to halt and the data retention limits not to be enforced.

Increasing the size of the PVC to 40Gi.

/cc @dhiller @xpivarc 


dhiller (approved):
/approve

PR #2372, opened 2022-10-06
Increase thanos-compact pvc to 100Gi due to out space error

brianmcarey:
The compaction process gets halted due to the persistent volume running out of space. Increasing the volume size to 100Gi should help this issue.

/cc @dhiller 


dhiller (approved):
/approve

PR #2377, opened 2022-10-10
create CI job for test-benchmarks

alicefr:
Enable CI and run unit test for the repo:  https://github.com/kubevirt/test-benchmarks


github/ci/prow-deploy/files/jobs/kubevirt/test-benchmarks/test-benchmarks-presubmits.yaml
```diff
@@ -0,0 +1,27 @@
+presubmits:
+  kubevirt/libguestfs-appliance:
```
dhiller, line 2: This should be different, no?
```suggestion
  kubevirt/test-benchmarks:
```

github/ci/prow-deploy/files/jobs/kubevirt/test-benchmarks/test-benchmarks-presubmits.yaml
```diff
@@ -0,0 +1,27 @@
...
+  - name: pull-test-benchmarks
+    branches:
+      - main
+    always_run: true
+    optional: false
```
dhiller, line 7: For this first one it should be ok, but please bear in mind that in general it is advised to introduce new jobs with the config changed like this
```suggestion
    always_run: false
    optional: true
    skip_report: true
```
This implies that at first jobs are triggered manually until they are stable and also do not report in order to not confuse users with a possibly bad ci signal.

dhiller (changes requested):

alicefr:
@brianmcarey can you please give me a hint why the `check-prow-config` job is failing?

brianmcarey:
> @brianmcarey can you please give me a hint why the `check-prow-config` job is failing?

It looks like it is missing the trigger plugin - I think you need to add something like the following:
https://github.com/kubevirt/project-infra/blob/ed0b9ba5c23cd4101ddf192675174a0a7e0c50ce/github/ci/prow-deploy/kustom/base/configs/current/plugins/plugins.yaml#L294-L299

brianmcarey (approved):
/approve

PR #2392, opened 2022-10-17
Mirror cri-o version 1.25 to be used by 1.25 provider

brianmcarey:
/cc @xpivarc @dhiller 


dhiller (approved):
/approve

nit: I think we should automate this into the script to get rid of the change being required for every new release

PR #2394, opened 2022-10-18
Update nmstate presubmit to use podman bootstrap

brianmcarey:
/cc @qinqon @rhrazdil 

### DIFF
--- a/github/ci/prow-deploy/files/jobs/nmstate/nmstate/nmstate-presubmits.yaml
+++ b/github/ci/prow-deploy/files/jobs/nmstate/nmstate/nmstate-presubmits.yaml
@@ -11,7 +11,7 @@ presubmits:
         grace_period: 5m
       max_concurrency: 6
       labels:
-        preset-dind-enabled: "true"
+        preset-podman-in-container-enabled: "true"
         preset-docker-mirror-proxy: "true"
         preset-shared-images: "true"
       cluster: prow-workloads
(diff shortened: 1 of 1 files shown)


brianmcarey:
@qinqon I guess `pull-nmstate-integ_tier1-k8s` is no longer used? Will i go ahead and remove the config?

qinqon:
> @qinqon I guess `pull-nmstate-integ_tier1-k8s` is no longer used? Will i go ahead and remove the config? 

It is it was fixed by nmstate team a week ago, we have to check why is failing.

brianmcarey:
> 
> It is it was fixed by nmstate team a week ago, we have to check why is failing.

It might just be an issue with the rehearsal plugin - the job is looking for a `main` branch which doesn't exist in the nmstate project - the default branch is called `base`. I will see if I can test this another way.

brianmcarey:
Yes - ran succesfully here with podman so it looks good - https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/nmstate_nmstate/2070/pull-nmstate-integ_tier1-k8s/1583372208846147584

brianmcarey:
> > @qinqon I guess `pull-nmstate-integ_tier1-k8s` is no longer used? Will i go ahead and remove the config?
> 
> It is it was fixed by nmstate team a week ago, we have to check why is failing.

Its an issue with the version of the rehearsal plugin that is deployed - it is quite old (`quay.io/kubevirtci/rehearse:v20200923-8930f83`) and doesn't contain the change to detect the default branch.

I am trying to get a newer version deployed but I don't think that we need to wait for that here?

PR #2405, opened 2022-10-20
Ad Shelly as a velero-plugin admin

brybacki:

### DIFF
--- a/github/ci/prow-deploy/files/orgs.yaml
+++ b/github/ci/prow-deploy/files/orgs.yaml
@@ -584,6 +584,7 @@ orgs:
           - tomob
           - aglitke
           - brybacki
+          - shellyka13
         repos:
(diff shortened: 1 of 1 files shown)


github/ci/prow-deploy/files/orgs.yaml
```diff
@@ -584,6 +584,7 @@ orgs:
           - tomob
           - aglitke
           - brybacki
+          - shellyka13
```
akalenyu, line 587: just double-checking if case matters here, because it's spelled @ShellyKa13

akalenyu: right, I do see some usernames in this file that are different, but that doesn't matter I guess

brianmcarey (approved):
/approve

PR #2406, opened 2022-10-20
Switch rootful vgpu lane to be optional

brianmcarey:
Currently there are two vgpu lanes running for each PR - running just the non-root lane should be enough for this testing - similar is done for the sriov lane.

/cc @xpivarc @dhiller 


xpivarc:
/lgtm
FYI @vladikr

dhiller (commented):
/approve

PR #2303, opened 2022-08-30
add new repository test-benchmarks

alicefr:
Create new repository test-benchmarks in KubeVirt to store the benchmark tests.

Currently, the code is hosted in my github account: 
  https://github.com/alicefr/kubevirt-test-vm

Moving the repository offers a better visibility and hopefully other users can find this useful and contribute to it.

PR #2415, opened 2022-10-24
Add common-instancetypes repo to KubeVirt

lyarwood:
/cc @ksimon1 

This project will hopefully eventually replace common-templates. It has been residing under my personal account for a while now and is ready to start the migration to the KubeVirt namespace.

### DIFF
--- a/github/ci/prow-deploy/files/orgs.yaml
+++ b/github/ci/prow-deploy/files/orgs.yaml
@@ -270,6 +270,9 @@ orgs:
         description: This is a demonstration of how KubeVirt could be displayed in Cockpit
         has_projects: false
         has_wiki: false
+      common-instancetypes:
+        description: Instancetypes and preferences for running VMs on KubeVirt
+        has_projects: true
       common-templates:
         description: Templates for running VMs on KubeVirt
         has_projects: true
@@ -861,3 +864,11 @@ orgs:
           demo: write
           katacoda-scenarios: admin
           kubevirt-tutorial: write
+      common-instancetypes-maintainers:
+        description: "Maintainers of common-instancetypes"
+        privacy: closed
+        maintainers:
+          - lyarwood
+          - ksimon1
+        repos:
+          common-instancetypes: admin


github/ci/prow-deploy/files/orgs.yaml
```diff
@@ -270,6 +270,9 @@ orgs:
         description: This is a demonstration of how KubeVirt could be displayed in Cockpit
         has_projects: false
         has_wiki: false
+      common-instancetypes:
```
ksimon1, line 273: I am wondering if name common-instance-types would be better :)

lyarwood: hmmm the term isn't hyphenated or even separated in our API or user docs, the latter might be a mistake on my part given AWS does refer to their version as `instance types`. I'm on the fence about the additional hyphen here tbh but I'll let others chime in.

dominikholler: I vote also for keeping all realted naming consistent with the API, which means using `common-instancetypes` here.

lyarwood:
@dhiller @brianmcarey would either of you mind looking at this and helping me move it forward. I've essentially copied the approach taken in #2303 so please let me know if this is wrong.

brianmcarey (approved):
/approve

fabiand:
FWIW - I'd even have considered to call the repo just `instancetypes`. The benefit of it would have been a pretty nice URL: github.com/kubevirt/instancetypes which is easy to remember and descriptive.
After all I'm expceting (hoping!) that this URL will be commonly used to create custom derivatives.

Worth to change it?

lyarwood:
> FWIW - I'd even have considered to call the repo just `instancetypes`. The benefit of it would have been a pretty nice URL: github.com/kubevirt/instancetypes which is easy to remember and descriptive. After all I'm expceting (hoping!) that this URL will be commonly used to create custom derivatives.
> 
> Worth to change it?

To me this would suggest that these are the only instancetypes available instead of a suggested common set so I'd rather stick with the current naming.

PR #2416, opened 2022-10-25
Revert prow due to issue with applying sriov labels to pods

brianmcarey:
Theres an issue[1] with sriov pods not getting the labels applied correctly. This causes a clash with sriov pods being scheduled together as the anti-affinity can no longer work correctly.

https://storage.googleapis.com/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/8597/pull-kubevirt-e2e-kind-1.22-sriov/1584691345816031232/podinfo.json

[1] https://github.com/kubevirt/kubevirt/issues/8656

/cc @dhiller @oshoval @rmohr

### DIFF
--- a/github/ci/prow-deploy/kustom/base/configs/current/config/config.yaml
+++ b/github/ci/prow-deploy/kustom/base/configs/current/config/config.yaml
@@ -9,10 +9,10 @@ plank:
       timeout: 2h
       grace_period: 15s
       utility_images:
-        clonerefs: "gcr.io/k8s-prow/clonerefs:v20221020-6d9d0a69a3"
-        initupload: "gcr.io/k8s-prow/initupload:v20221020-6d9d0a69a3"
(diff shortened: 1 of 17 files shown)


oshoval (approved):
Thanks
maybe worth to add the PR desc that the annotation is missing as well
hence the CNI isn't triggered and the PF don't exist
which was the visible syndrome

rmohr:
> Thanks
> maybe worth to add the PR desc that the annotation is missing as well
> hence the CNI isn't triggered and the PF don't exist
> which was the visible syndrome

Seems like this is a general regression where annotations and labels are not passed through in tide jobs. That breaks a lot of things in our CI :) I am surprised that we did not experience more issues.

dhiller (approved):
/approve

We should also investigate whether there's an issue on test-infra for this already.

PR #2417, opened 2022-10-25
Switch non-root sriov lane back to required

brianmcarey:
This reverts commit a4b49f65c2f9f82cd28c0a2467106a323a074b43.

The issue was not related to nonroot but to pods being mislabelled by prow. Switch the non-root lane back to being required.

/cc @EdDev @oshoval 

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
@@ -428,7 +428,7 @@ presubmits:
       nodeSelector:
         type: bare-metal-external
       priorityClassName: windows
-  - always_run: false
+  - always_run: true
     annotations:
       fork-per-release: "true"
       k8s.v1.cni.cncf.io/networks: multus-cni-ns/sriov-passthrough-cni,multus-cni-ns/sriov-passthrough-cni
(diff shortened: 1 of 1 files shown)


brianmcarey:
> > /hold
> 
> why? waiting to see all is fine or something was found ?

Yes just waiting to see things stabilize - also introducing a new required job while there is a large merge queue will leave a lot of PRs waiting for a job that hasn't been triggered.

github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
```diff
@@ -529,7 +528,7 @@ presubmits:
       sriov-pod: "true"
     max_concurrency: 10
     name: pull-kubevirt-e2e-kind-1.22-sriov
-    optional: false
+    optional: true
```
oshoval, line 531: can we omit the line as this will be the default when always_run is false maybe ?

oshoval:
Is there maybe a flag that say operational: yes/no , so we can put it on both of the lanes and just by changing this only flag toggle between them upon needs?
(less fields to change)
or maybe another flag that selects the operational one in some higher level or so (just an idea)

brianmcarey:
/close 

This will be handled by https://github.com/kubevirt/project-infra/pull/2418

PR #2419, opened 2022-10-27
Remove DV GC lanes, as it is enabled by default

arnongilboa:


brianmcarey (approved):
/approve

PR #2430, opened 2022-11-03
Add missing common-instancetypes prow configuration

lyarwood:


github/ci/prow-deploy/kustom/base/configs/current/config/config.yaml
```diff
@@ -351,6 +353,10 @@ branch-protection:
           branches:
             main:
               protect: true
+        common-instancetypes:
+          branches:
+            master:
```
dhiller, line 358: Are you sure that the branch is still called `master`?

dhiller (changes requested):
I think you need to correct the branch name in branch-protection, other that that looks fine to me.

github/ci/prow-deploy/kustom/base/configs/current/config/config.yaml
lyarwood: Ah apologies, nice catch!

dhiller (approved):
/approve

PR #2435, opened 2022-11-04
Add optional presubmit to test cluster-sync against perf cluster

brianmcarey:
/cc @xpivarc @dhiller 

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
@@ -2358,3 +2358,80 @@ presubmits:
           privileged: true
       nodeSelector:
         type: bare-metal-external
+  - always_run: false
+    cluster: ibm-prow-jobs
+    decorate: true
+    decoration_config:
+      grace_period: 5m0s
+      timeout: 1h0m0s
+    labels:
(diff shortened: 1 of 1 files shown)


github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
```diff
@@ -2358,3 +2358,80 @@ presubmits:
           privileged: true
       nodeSelector:
         type: bare-metal-external
+  - always_run: false
```
xpivarc, line 2361: Don't we need `optional: true`?

dhiller: IIRC if `always_run: true` is set, this job is conditional anyway, so it is implicitly optional. Can't find the doc link for this right now, though.

dhiller: There it is: https://docs.prow.k8s.io/docs/legacy-snapshot/prow/jobs/#requiring-jobs-for-auto-merge-through-tide

So in short: if neither `run_if_changed` nor `skip_if_only_changed` are set this job is already conditional and doesn't match any set, so it's optional.

dhiller (approved):
/approve

PR #2443, opened 2022-11-09
Update ovs-cni postsubmit jobs to use podman

brianmcarey:
/cc @oshoval @dhiller 


oshoval (approved):
Thanks
Wonder why i didn't change it

Note please that it is better to not run it as it will make versions
we better wait for a PR to be merged first on the repo that will trgger it

dhiller (commented):
/approve

@oshoval could not follow completely, thus not sure whether you mean that we should hold this PR. But since you LGTMed I guess we are good.

oshoval:
> /approve
> 
> @oshoval could not follow completely, thus not sure whether you mean that we should hold this PR. But since you LGTMed I guess we are good.

We are ok thanks,
we will just need to keep in mind to check the versions when there is a new release
(but not to trigger postsubmit until a real PR is merged)

PR #2461, opened 2022-11-15
Add additional image store to podman conf

brianmcarey:
An additional read-only image store will allow podman to run regularly used images such as the kubevirtci node images without having to pull them everytime. This will greatly reduce the amount of data being pulled over the network and written to the disks.

Initially the additonal image store will be empty until there is automation in place to populate and manage this image store. This update just enables the bootstrap image to use the new image store.

The storage.conf is based on the default config[1] with an update for the additional image store.

[1] https://github.com/containers/podman/blob/6c37cf332439989148e408af49afaf7c0cb6e166/vendor/github.com/containers/storage/storage.conf

/cc @dhiller  @xpivarc @rmohr 

### DIFF
--- a/images/bootstrap/Dockerfile
+++ b/images/bootstrap/Dockerfile
@@ -84,6 +84,9 @@ RUN dnf -y update; dnf -y reinstall shadow-utils && \
 
 VOLUME /var/lib/containers
 
+COPY storage.conf /etc/containers/storage.conf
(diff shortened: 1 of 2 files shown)


rmohr:
/lgtm
/approve

/hold

feel free to unhold if you think it is ready.

brianmcarey:
thanks @rmohr !

/unhold

PR #2465, opened 2022-11-17
Add files to build shared-images-controller image

brianmcarey:
This image is based off of the bootstrap image as the podman go bindings require a running podman system socket to connect to.

Currently this image will just periodically download the latest kubevirtci node images that are required by kubevirt/kubevirt main to the shared images store. It also removes any old images that do not match this latest kubevirtci tag.

The storage.conf is based off of the default config just with the read/write location of container storage updated to match the path of the shared images location.

/cc @xpivarc @dhiller 


images/shared-images-controller/main.go
```diff
@@ -0,0 +1,166 @@
...
+	for _, version := range versions {
+		log.Infoln("Kubevirt Provider version: ", version)
+		name := fmt.Sprintf("%sk8s-%s:%s", kubevirtci_repo, version, tag)
+		if checkImagePresent(imageNames, name) {
+			log.Infoln("Image already present:", name)
+		} else {
+			log.Infoln("Pulling image: ", name)
+			_, err := images.Pull(ctx, name, nil)
+			if err != nil {
+				log.WithError(err).Errorf("Failed to pull image:", name)
```
dhiller, line 98: Maybe directly return the error, since we are logging it twice, once here and a second time in the main func?

```suggestion
```

images/shared-images-controller/main.go
```diff
@@ -0,0 +1,166 @@
...
+		err = pullRequiredImages(connText, tag)
+		if err != nil {
+			log.WithError(err).Errorf("Failed to fetch image names")
+		}
+		err = cleanOldImages(connText, tag)
+		if err != nil {
+			log.WithError(err).Errorf("Failure occured when deleting old images")
+		}
+		log.Infof("Waiting for %d hours before checking again", period)
+		time.Sleep(period * time.Hour)
```
xpivarc, line 163: When do we kill the job?

brianmcarey:
> Mostly nits, please see inline comments.

@dhiller  - thanks for the good feedback. I have addressed your comments here - https://github.com/kubevirt/project-infra/pull/2465/commits/d8b7b72f373e067ee11668b28bce91a9b577bb8b

images/shared-images-controller/main.go
```diff
@@ -0,0 +1,156 @@
...
+		for _, imageName := range i.RepoTags {
+			imageNames[imageName] = struct{}{}
+		}
+	}
+
+	for _, version := range versions {
+		log.Infoln("Kubevirt Provider version: ", version)
+		name := fmt.Sprintf("%sk8s-%s:%s", kubevirtci_repo, version, tag)
+		_, exists := imageNames[name]
+		if exists {
```
dhiller, line 84: nit: you can do this in one line:

```suggestion
		if _, exists := imageNames[name]; exists {
```

dhiller (approved):
/approve

Looks good to me. One nit inline.

/hold to decide whether you first want to fix the nit

images/shared-images-controller/main.go
```diff
@@ -0,0 +1,155 @@
...
+
+	if os.Getenv("XDG_RUNTIME_DIR") != "" {
+		sock_dir := os.Getenv("XDG_RUNTIME_DIR")
+		socket = "unix:" + sock_dir + "/podman/podman.sock"
+	}
+	connText, err := bindings.NewConnection(context.Background(), socket)
+	if err != nil {
+		log.WithError(err).Fatalf("Could not connect to podman socket %d", socket)
+	}
+	for {
```
dhiller, line 138: I hope I am not overthinking this, but what if the process gets killed in the middle of an action like deleting images here? Do we need to care about broken image downloads? I guess not, since podman takes care of that? But what about if the deletion is only half way done? Does podman take care of that also?

However, I've been thinking about a more elegant way of running a periodic process, and I've found [Golang Ticker](https://pkg.go.dev/time#Ticker).

I don't think we need to do this now, but maybe in a later step. WDYT?

brianmcarey: > I hope I am not overthinking this, but what if the process gets killed in the middle of an action like deleting images here? Do we need to care about broken image downloads? I guess not, since podman takes care of that? But what about if the deletion is only half way done? Does podman take care of that also?

We would be relying on podman here - podman will try to download the image again when the loop comes around again so it should be ok on the next iteration. We would have to rely on podman for the image deletion too. 
> 
> However, I've been thinking about a more elegant way of running a periodic process, and I've found [Golang Ticker](https://pkg.go.dev/time#Ticker).
> 
> I don't think we need to do this now, but maybe in a later step. WDYT?

That does look good - we could definitely look into that later.

dhiller (approved):
OK then, please create an issue to use Golang Ticker as discussed in a later iteration.

images/shared-images-controller/main.go
xpivarc: Ok, I got confused by the addition of pre/post submit jobs. Do you want to use them for testing? If so would it make sense to implement an option to indicate you want one run and then exit?

brianmcarey: the pre/post submit jobs are just used for building / publishing the image that will be used by the daemonset.

xpivarc: My bad. I see it now. Should have looked before.

xpivarc:
+1

PR #2474, opened 2022-11-21
Revert "Add newline to commit msg for bump prow jobs (#2466)"

brianmcarey:
This reverts commit 8fe75668f8088e3c3a37daba279a4e0e04a382ec.

The newline character did not result in a new line in the PR description so it did not have the desired effect[1]. Revert back to previous commit message.

/cc @dhiller 

[1] https://github.com/kubevirt/project-infra/pull/2472#issue-1456714984


dhiller (approved):
/approve

I was thinking a bit about this. I saw two options:
 * make this a command line option, i.e. `-C @kubevirtci-taskforce`
 * use a special env var that you then set in the job config 

Both options would require adjustments inside the script and the jobconfig. Not sure what would be the best here. I'd lean towards the command line option since it's more explicit.

PR #2486, opened 2022-11-25
Switch check-provision-1.22 to optional

brianmcarey:
The 1.22 provider is in the process of being retired - we should not be running an extra presubmit per PR.

/cc @dhiller 


dhiller (approved):
/approve

PR #2498, opened 2022-12-02
Make the performance job voting and running per commit

rthallisey:
https://groups.google.com/g/kubevirt-dev/c/zdgrOmJcodc/m/72-MqVwzGQAJ

The performance job has been stable for many months so I think its time to make it voting.

This job is designed to catch the following issues on a new PR:
1) Performance regression (slow phase transitions)
2) Scale regression (too many http requests)
3) control-plane mem increase

Catching these issues in an MR would allow the author to add a release note documenting a performance change.

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
@@ -240,7 +240,7 @@ presubmits:
           privileged: true
       nodeSelector:
         type: bare-metal-external
-  - always_run: true
+  - always_run: false
     annotations:
       fork-per-release: "true"
       testgrid-dashboards: kubevirt-presubmits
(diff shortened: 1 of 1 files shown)


rthallisey:
/cc @brianmcarey 

Relates-to: https://github.com/kubevirt/kubevirt/issues/8749

github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
```diff
@@ -257,7 +257,6 @@ presubmits:
       preset-shared-images: "true"
     max_concurrency: 2
     name: pull-kubevirt-e2e-k8s-1.22-sig-performance
-    optional: true
     run_if_changed: ^pkg/virt-api/.*|^pkg/virt-handler/.*|^pkg/virt-controller/.*|^pkg/virt-launcher/.*
```
brianmcarey, line 260: `run_if_changed` clashes with `always_run: true` - has to be removed in this case.

github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
```diff
@@ -257,7 +257,6 @@ presubmits:
       preset-shared-images: "true"
     max_concurrency: 2
```
brianmcarey, line 258: If this is to run and be required on every kubevirt PR - we may need to increase the max concurrency to 4 or 6

brianmcarey (commented):
Hi @rthallisey  - thanks for the PR - just a couple of comments inline. 

/cc @dhiller

github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
brianmcarey: Sorry - I wasn't clear enough - in this case I meant that `run_if_changed` should be removed.

brianmcarey: It would - i think this has come up previously - the branch protector in prow won't stop merges if conditionally run jobs fail so even with the latest `run_if_changed` it doesn't guarantee that it will stop the merge in a failed case. https://github.com/kubevirt/project-infra/pull/2249#issuecomment-1212089928

rthallisey:
@brianmcarey updated based on our last discussion

brianmcarey (approved):
/lgtm
/approve

PR #2511, opened 2022-12-09
Add daemonset for shared images to prow-workloads cluster

brianmcarey:
This daemonset is responsible for maintaining the shared image store on
each node in the workloads cluster. It will download the latest
kubevirtci node images and remove any old images that are no longer
required. The e2e jobs will mount this directory as a read only shared
image store.

/cc @enp0s3 @xpivarc 

### DIFF
--- a/github/ci/prow-deploy/kustom/overlays/workloads-production/resources/shared-images-controller.yaml
+++ b/github/ci/prow-deploy/kustom/overlays/workloads-production/resources/shared-images-controller.yaml
@@ -1,0 +1,38 @@
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: shared-images-controller
+  namespace: kubevirt-prow
+  labels:
+    name: shared-images-controller
+spec:
+  selector:
+    matchLabels:
+      name: shared-images-controller
+  updateStrategy:
+    type: RollingUpdate
+  template:
+    metadata:
+      labels:
+        name: shared-images-controller
+    spec:
+      nodeSelector:
+        type: bare-metal-external
+      terminationGracePeriodSeconds: 1
+      containers:
+      - name: shared-images-controller
+        image: quay.io/kubevirtci/shared-images-controller:v20221208-a669d6f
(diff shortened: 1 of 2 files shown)


brianmcarey:
/unhold

Should be ready now.

github/ci/prow-deploy/kustom/overlays/workloads-production/resources/shared-images-controller.yaml
```diff
@@ -0,0 +1,38 @@
...
+      terminationGracePeriodSeconds: 1
+      containers:
+      - name: shared-images-controller
+        image: quay.io/kubevirtci/shared-images-controller:v20221208-a669d6f
+        command: [ "/usr/local/bin/runner.sh", "/shared-images-controller"]
+        resources:
+          requests:
+            memory: 2Gi
+        securityContext:
+          privileged: true
```
xpivarc, line 30: Does it need to be privileged?

xpivarc (commented):
cc @dhiller

github/ci/prow-deploy/kustom/overlays/workloads-production/resources/shared-images-controller.yaml
brianmcarey: It requires a running podman socket to manage the images - for now I think we require it to be privileged in order to start podman in the container. 
It would be cool to see if we could move this and the e2e jobs to not require privileged but I think we would need to deploy the fuse-device-plugin to try rootless without privileged - https://www.redhat.com/sysadmin/podman-inside-kubernetes

xpivarc: Got it. I think we can do without but its not blocking issue

xpivarc:
+1

brianmcarey:
/cc @dhiller 

can you please take a look?

dhiller (approved):
/approve

PR #2512, opened 2022-12-09
Increase thanos data retention for 1hr resolution to 1 year

brianmcarey:
Data retained in the thanos bucket has been greatly reduced so the data retention period for the 1hr resolution can be increased to 1 year

This will allow better visibility of trends over time.

/cc @dhiller 


dhiller (approved):
/approve

PR #2514, opened 2022-12-11
single stack: Switch to k8s-1.25-ipv6

oshoval:
Switch to k8s-1.25-ipv6 (from k8s-1.24-ipv6)

Additional change:-
Bump the non required lanes to future lane k8s-1.26-ipv6
instead of deleting them and recreating next time.

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
@@ -1135,7 +1135,7 @@ presubmits:
       preset-podman-in-container-enabled: "true"
       preset-shared-images: "true"
     max_concurrency: 11
-    name: pull-kubevirt-e2e-k8s-1.24-ipv6-sig-network
+    name: pull-kubevirt-e2e-k8s-1.25-ipv6-sig-network
     skip_branches:
     - release-\d+\.\d+
     spec:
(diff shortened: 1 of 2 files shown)


oshoval:
/cc @brianmcarey 

Once we check the lane
Will create also a PR that remove the old lane

github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
```diff
@@ -1147,7 +1147,7 @@ presubmits:
         - automation/test.sh
         env:
         - name: TARGET
-          value: k8s-1.24-ipv6-sig-network
+          value: k8s-1.25-ipv6-sig-network
```
oshoval, line 1150: @brianmcarey 
Is there a way to create env var (either in the yaml of in the bash scripts) that depend on the job name ?
this way we will need to change only the `name` field

oshoval: Thanks, do you like this idea ?
if so we can try on follow PR

oshoval:
https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/8926/pull-kubevirt-e2e-k8s-1.25-ipv6-sig-network/1602212686857244672

Passed so removed the WIP here
/cc @brianmcarey 
Please take a look

brianmcarey (approved):
/approve

PR #2524, opened 2022-12-21
Update bootstrap images to use Fedora 37

brianmcarey:
Fedora 37 has been released for over a month now[1] and looks stable enough
so we can update the bootstrap images to F37.

The kubevirt builder image is using golang version 1.19.2 [2] - this golang
bootstrap image should default to the same.

[1] https://fedoramagazine.org/announcing-fedora-37/
[2] https://github.com/kubevirt/kubevirt/blob/297edbe21961c7a3ef462f94f4eeead0520ec043/hack/builder/Dockerfile#L7

/cc @dhiller @enp0s3


dhiller (approved):
/approve

PR #2538, opened 2023-01-05
Fix bootstrap image version for arm64 workloads

brianmcarey:
The image tag of the arm64 based jobs were updated to an image that does not support arm64. This updates theses jobs to the latest bootstrap image that supports arm64.

/cc @zhlhahaha @dhiller 

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml
@@ -640,7 +640,7 @@ periodics:
         value: "1"
       - name: KUBEVIRT_E2E_FOCUS
         value: arm64
-      image: quay.io/kubevirtci/bootstrap:v20221222-8f66e7e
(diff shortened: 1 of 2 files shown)


zhlhahaha:
/lgtm
Thanks, Brian

dhiller:
> The image tag of the arm64 based jobs were updated to an image that does not support arm64. This updates theses jobs to the latest bootstrap image that supports arm64.

How come that a later version does not support arm any more?

dhiller:
> > The image tag of the arm64 based jobs were updated to an image that does not support arm64. This updates theses jobs to the latest bootstrap image that supports arm64.
> 
> How come that a later version does not support arm any more?

Also, from what I can tell, we are only [building the arm image](https://github.com/kubevirt/project-infra/blob/a0161b75068f7f3a4f8fe1f00f5ce7c108d4b11e/github/ci/prow-deploy/files/jobs/kubevirt/project-infra/project-infra-postsubmits.yaml#L107) with the postsubmit, but with the bootstrap-legacy? The regular is build with a periodics, which might be the reason why the arm image is overwritten?

dhiller (commented):
/approve

To unblock the builds

PR #2540, opened 2023-01-05
Build multiarch bootstrap image as part of postsubmit job

brianmcarey:
Only building the multiarch bootstrap image periodically causes inconsistencies which can cause issues with image mismatches.

This change removes the periodic build job and enables the multiarch build in the postsubmit job.

Also updates publish script to use podman rather than docker.

/cc @zhlhahaha @dhiller 


brianmcarey:
/hold 

Testing locally with phaino

brianmcarey:
Local test run completed successfully 

/unhold

dhiller (approved):
/approve

Thanks for taking this, @brianmcarey !

PR #2548, opened 2023-01-12
ksd: remove the release-note plugin from ksd

AlonaKaplan:
KubeSecondaryDNS is already using a builtin github action to create the release notes based on PR titles.
In this repository we don't need to manually add another custom collectible note to the PR description,
hence remove the plugin that expects it.

### DIFF
--- a/github/ci/prow-deploy/kustom/base/configs/current/config/config.yaml
+++ b/github/ci/prow-deploy/kustom/base/configs/current/config/config.yaml
@@ -135,6 +135,7 @@ tide:
   - repos:
     - kubevirt/libguestfs-appliance
     - kubevirt/project-infra
+    - kubevirt/kubesecondarydns
     - kubevirt/cluster-api-provider-kubevirt
     - kubevirt/machine-remediation-operator
     - kubevirt/kubevirtci
@@ -171,7 +172,6 @@ tide:
     - k8snetworkplumbingwg/ovs-cni
     - kubevirt/macvtap-cni
     - kubevirt/cluster-network-addons-operator
-    - kubevirt/kubesecondarydns
     - kubevirt/bridge-marker
     - nmstate/kubernetes-nmstate
     - k8snetworkplumbingwg/kubemacpool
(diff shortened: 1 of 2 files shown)


oshoval:
It is not the same release note as this PR removes, but the PR descriptions
so might better to change please the description accordingly?

oshoval:
> > It is not the same release note as this PR removes, but the PR descriptions so might better to change please the description accordingly?
> 
> What do you mean by "it is not same release note as this PR removes?". Same as what? This PR removes the plugin that adds `do-not-merge/release-note-label-needed` to a PR wih no release-note markdown block.

Right, this (what you just described) is the release-note that appears on the bottom of a PR desc
versus the one that we have of github that adds the PR title only and named the same

AlonaKaplan:
> > > It is not the same release note as this PR removes, but the PR descriptions so might better to change please the description accordingly?
> > 
> > 
> > What do you mean by "it is not same release note as this PR removes?". Same as what? This PR removes the plugin that adds `do-not-merge/release-note-label-needed` to a PR wih no release-note markdown block.
> 
> Right, this (what you just described) is the release-note that appears on the bottom of a PR desc versus the one that we have of github that adds the PR title only and named the same. Just to not confused between the two please.

Sorry. but I don't understand what you're trying to say. What description/title do you suggest to this PR?

oshoval:
There are those two:
1. https://github.com/kubevirt/kubesecondarydns/blob/main/.github/workflows/image-push.yaml#L69
2. https://github.com/kubevirt/project-infra/pull/2548/files#diff-3a088ffbe61dd30f930cd1855ab72a652fcea4a33c4b047eeb6df38f7a79ad51L137

They are not the same, even that both are named release notes
So something like this maybe:

KubeSecondaryDNS is already using a builtin github action to create the release notes based on PR titles.
In this repository we don't need to manually add another custom collectible note to the PR description,
hence remove the plugin that expects it.

oshoval:
btw just for the record the one we remove is pretty much "extended release note" as the template says

Thanks
/lgtm

PR #2557, opened 2023-01-17
Enable PSA on 1.26 lanes

acardace:
Since https://github.com/kubevirt/kubevirt/pull/8649 KubeVirt can run VMs as `restricted` PSA pods, we should run all the latest lanes with PSA enabled to make sure we don't break this.

Running with PSA enabled translates in testing KubeVirt in the most restrictive Kubernetes environment and thus it's only more beneficial if we run every PR against PSA lanes.

Signed-off-by: Antonio Cardace <acardace@redhat.com>

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
@@ -1375,6 +1375,49 @@ presubmits:
           privileged: true
       nodeSelector:
         type: bare-metal-external
+  - always_run: false
+    skip_report: true
+    annotations:
+      fork-per-release: "true"
+      testgrid-dashboards: kubevirt-presubmits
+    cluster: prow-workloads
+    decorate: true
+    decoration_config:
+      grace_period: 5m0s
+      timeout: 4h0m0s
+    labels:
+      preset-bazel-cache: "true"
+      preset-bazel-unnested: "true"
+      preset-docker-mirror-proxy: "true"
+      preset-podman-in-container-enabled: "true"
+      preset-podman-shared-images: "true"
+      preset-shared-images: "true"
+    max_concurrency: 11
+    name: pull-kubevirt-e2e-k8s-1.26-ipv6-psa-sig-network
+    optional: true
+    skip_branches:
+    - release-\d+\.\d+
+    spec:
+      containers:
+      - command:
+        - /usr/local/bin/runner.sh
+        - /bin/sh
+        - -c
+        - automation/test.sh
+        env:
+        - name: TARGET
+          value: k8s-1.26-ipv6-sig-network
+        - name: KUBEVIRT_PSA
+          value: "true"
+        image: quay.io/kubevirtci/bootstrap:v20221222-8f66e7e
+        name: ""
+        resources:
+          requests:
+            memory: 29Gi
+        securityContext:
+          privileged: true
+      nodeSelector:
+        type: bare-metal-external
   - always_run: true
     annotations:
       fork-per-release: "true"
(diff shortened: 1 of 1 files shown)


xpivarc:
@acardace Did we check these lanes pass?

dhiller:
Regarding ipv6 failure: the lane seems to be broken, since a test run on it showed the same symptoms: https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/9066/pull-kubevirt-e2e-k8s-1.26-ipv6-sig-network/1615638876666204160

github/ci/prow-deploy/files/jobs/kubevirt/kubevirtci/kubevirtci-presubmits.yaml
```diff
@@ -491,7 +491,7 @@ presubmits:
         - /usr/local/bin/runner.sh
         - /bin/sh
         - -c
-        - cd cluster-provision/k8s/1.26 && ../provision.sh
+        - cd cluster-provision/k8s/1.26 && KUBEVIRT_PSA='true' ../provision.sh
```
dhiller, line 494: Sorry for this late remark, but I think you should do this in a separate PR, since we can't change the kubevirt/kubevirt lanes to enable psa before the psa is working inside kubevirt.

acardace: Gotcha, removed it!

dhiller: Thanks! LMK when you have created the other PR.

@brianmcarey I think we might set the psa check-provisions to `always_run: true` so we get any failures. WDYT?

brianmcarey: Yes I think we should have it set to always run. It is an important provider for kubevirt.

dhiller:
https://github.com/kubevirt/kubevirtci/pull/950 has passed tests, now we wait for publish of kubevirtci to succeed, then the bump to happen and get merged.

dhiller:
https://github.com/kubevirt/kubevirt/pull/9081 is waiting for merge.

dhiller:
@acardace I wasn't expecting so many failures.

I think it's not the best idea to take this in right away, since it will create a lot of friction.

I'd suggest to duplicate the lane definitions and make them triggerable manually by setting `always_run: false`, `skip_report: true` (see docs [here](https://docs.prow.k8s.io/docs/jobs/)), so that other PR authors aren't blocked by these changes right away.

This would enable people to work on PSA, since they then can trigger the lanes by commenting with `/test pull-kubevirt-e2e-k8s-1.26-psa-sig-network` as usual.

dhiller:
What do others think? @xpivarc @enp0s3 @brianmcarey

acardace:
> @acardace I wasn't expecting so many failures.
me too

> 
> I think it's not the best idea to take this in right away, since it will create a lot of friction.
> 
> I'd suggest to duplicate the lane definitions and make them triggerable manually by setting `always_run: false`, `skip_report: true` (see docs [here](https://docs.prow.k8s.io/docs/jobs/)), so that other PR authors aren't blocked by these changes right away.
> 
> This would enable people to work on PSA, since they then can trigger the lanes by commenting with `/test pull-kubevirt-e2e-k8s-1.26-psa-sig-network` as usual.

What about enforcing the PSA compute lane and making the others optional as you described?

xpivarc:
> > @acardace I wasn't expecting so many failures.
> 
> me too
> 
> > I think it's not the best idea to take this in right away, since it will create a lot of friction.
> > I'd suggest to duplicate the lane definitions and make them triggerable manually by setting `always_run: false`, `skip_report: true` (see docs [here](https://docs.prow.k8s.io/docs/jobs/)), so that other PR authors aren't blocked by these changes right away.
> > This would enable people to work on PSA, since they then can trigger the lanes by commenting with `/test pull-kubevirt-e2e-k8s-1.26-psa-sig-network` as usual.
> 
> What about enforcing the PSA compute lane and making the others optional as you described?

Sounds good.  Looking at the failures it should be pretty straightforward to fix them. The only problem will be the operator lane. There I think what is happening is the following: We downgrade Kubevirt but we have PSA enabled. The n-1 version is not PSA compliant and therefore it fails to run any VM.

dhiller (approved):
/approve

acardace:
@dhiller can you remove the hold?

PR #2571, opened 2023-01-26
sriov: Use podman image

oshoval:


oshoval:
"Command Output: Error: statfs /var/log/audit: no such file or directory"
will look on it later

oshoval:
weird, it pass on other PRs but failed here even with docker

oshoval:
@brianmcarey 
testing this again sriov + podman due to the recent fixes of the netavark

(we might need to use latest image if it doesnt work, just for the chance that it has affect)

oshoval:
once we need we can reopen

PR #2575, opened 2023-01-26
Add workaround for cri-o version 1.24

brianmcarey:
The CentOS Stream 8 builds for cri-o 1.24 look to be failing.
The repo only has up to 1.24.1[1] which still hits this intermittent issue[2] which is fixed in 1.24.2.

Use the CentOS 8 build as a workaround for this.

Also remove cri-o 1.23 from kubevirt mirror. 

[1] https://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable:/cri-o:/1.24/CentOS_8_Stream/x86_64/
[2] https://github.com/cri-o/cri-o/issues/5889

/cc @dhiller


dhiller (approved):
/approve

PR #2579, opened 2023-01-27
Restore 1.26 compute lane name

acardace:
Due to https://github.com/kubevirt/project-infra/pull/2577#issuecomment-1406544238

Signed-off-by: Antonio Cardace <acardace@redhat.com>


dhiller:
@brianmcarey @acardace I think due to current lane status we really need to keep an eye on the automation, since it wasn't built for this kind of complex situation we are in currently. :confused:

PR #2583, opened 2023-01-30
Temporarily move pinc jobs to prow-workloads cluster

brianmcarey:
The docker image proxy is currently down on the ibm-prow-jos cluster - temporarily move the jobs that rely on this proxy to the prow-workloads cluster.

This should be changed back once the docker image proxy is back up.

/cc @dhiller 


dhiller (approved):
/approve

PR #2598, opened 2023-02-07
Remove PSA optional jobs

xpivarc:
There are voting jobs already and therefore
there is no need for optional ones.


acardace (approved):

brianmcarey (approved):
/approve

PR #2605, opened 2023-02-10
Remove unused presubmit lane for testing against perf cluster

brianmcarey:
This lane was introduced to test issues with the old perf cluster however it was never really used. Removing as no longer needed. 

/cc @dhiller @enp0s3 


dhiller (approved):
/approve

PR #2613, opened 2023-02-14
common-instancetypes: rework run_if_changed one final time

lyarwood:
/cc @0xFelix 
/cc @brianmcarey 

The previous attempt didn't work, so here's one final attempt with actually valid regex [1].

[1] https://regex101.com/r/dcAdWM/1


brianmcarey (approved):
/lgtm 

/approve

PR #2615, opened 2023-02-15
Set check-provision-k8s-1.26-centos9 to required

brianmcarey:
As the 1.26 CentOS 9 provider is the provider that testing will be carried out against in kubevirt[1] - the check-provision job should be set to required.

The 1.26 CentOS 8 provider will be removed shortly so a check-provision job should not be required there. 

[1] https://github.com/kubevirt/project-infra/pull/2610

/cc @dhiller @enp0s3


dhiller (approved):
/approve

PR #2627, opened 2023-02-20
Add SYS_CHROOT capability to podman in bootstrap image

brianmcarey:
A number of CI jobs require this capability to complete successfully[1]

It was recently removed from the default set of podman capabilities[2]

[1] https://github.com/kubevirt/kubevirt/issues/9279
[2] https://blog.podman.io/2022/12/dropping-capabilities-making-containers-more-secure/

### DIFF
--- a/images/bootstrap/containers.conf
+++ b/images/bootstrap/containers.conf
@@ -4,6 +4,7 @@ userns="host"
 ipcns="host"
 cgroupns="host"
 cgroups="disabled"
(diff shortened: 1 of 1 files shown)


images/bootstrap/containers.conf
```diff
@@ -4,6 +4,7 @@ userns="host"
 ipcns="host"
 cgroupns="host"
 cgroups="disabled"
+default_capabilities = ["CHOWN",  "DAC_OVERRIDE", "FOWNER",  "FSETID",  "KILL",  "NET_BIND_SERVICE",  "SETFCAP",  "SETGID",  "SETPCAP",  "SETUID", "SYS_CHROOT"]
```
dhiller, line 7: How can we make sure that we are adding only the capabilities we actually require?

@xpivarc @jean-edouard can we even do this?

xpivarc: Trial and error. I am afraid I don't have a better answer.

PR #2628, opened 2023-02-21
Update kubevirtci-bumper to bump minor version of alt providers

### LINKED ISSUE #2340: kubevirtci bumper missing variants of kubevirtci providers
The automated kubevirtci bumper which bumps the minor versions of the kubevirtci providers is currently missing variants of the major version providers such as `1.24-psa` & `1.24-ipv6`. The automation should be able to handle the bumping of these variants too.

An example automated PR: https://github.com/kubevirt/kubevirtci/pull/877

brianmcarey:
Previously alternative providers such as `k8s-1.26-centos9` were being skipped by the kubevirtci bumper for minor version bumps.

This update will bump the minor versions of these alternative providers.

Fixes #2340 

/cc @dhiller @xpivarc

### DIFF
--- a/robots/pkg/kubevirtci/kubevirtci.go
+++ b/robots/pkg/kubevirtci/kubevirtci.go
@@ -17,7 +17,7 @@ import (
 )
 
 var SemVerRegex = regexp.MustCompile(`^[v]?([0-9]+)\.([0-9]+)\.([0-9]+)$`)
-var SemVerMinorRegex = regexp.MustCompile(`^([0-9]+)\.([0-9]+)*`)
+var ProviderFolderRegex = regexp.MustCompile(`^([0-9]+)\.([0-9]+)(-[a-z0-9-]+)?$`)
 
 func BumpMinorReleaseOfProvider(providerDir string, minors []*github.RepositoryRelease) error {
 	// Update the latest three minor k8s versions
@@ -78,7 +78,7 @@ func isProviderDirectory(entry os.DirEntry) bool {
 	if !entry.IsDir() {
 		return false
 	}
-	return SemVerMinorRegex.MatchString(entry.Name())
+	return ProviderFolderRegex.MatchString(entry.Name())
 }
 
 func isSupportedProvider(entry os.DirEntry, supportedReleases []*github.RepositoryRelease) bool {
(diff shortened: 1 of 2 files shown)


robots/pkg/kubevirtci/kubevirtci_test.go
```diff
@@ -59,7 +59,6 @@ func TestReadExistingProviders(t *testing.T) {
 			name: "should ignore strange folders",
 			existing: []querier.SemVer{
 				newSemVer("v1x", "2", "3"),
-				newSemVer("1", "3-multus", "4"),
```
oshoval, line 62: maybe worth to add please a test for 1.26-centos9 template ?

robots/pkg/kubevirtci/kubevirtci.go
```diff
@@ -17,7 +17,7 @@ import (
 )
 
 var SemVerRegex = regexp.MustCompile(`^[v]?([0-9]+)\.([0-9]+)\.([0-9]+)$`)
-var SemVerMinorRegex = regexp.MustCompile(`^([0-9]+)\.([0-9]+)$`)
+var SemVerMinorRegex = regexp.MustCompile(`^([0-9]+)\.([0-9]+)*`)
```
dhiller, line 20: I think this change hides what is really going on here. We should make the possible extension of the path name explicit like so

```suggestion
var SemVerMinorRegex = regexp.MustCompile(`^([0-9]+)\.([0-9]+)(-[a-z0-9-]+)?$`)
```

While we are at this change, we should also rename the `var`  

```suggestion
var ProviderFolderRegex = regexp.MustCompile(`^([0-9]+)\.([0-9]+)*`)
```

I can't remember out of my head, but isn't this regex also used in other places? Maybe we need to create a separate one for `isProviderDirectory` ?

brianmcarey: I made these updates here - https://github.com/kubevirt/project-infra/pull/2628/commits/cf91037faad3869f4cb554ee300f61aff3ba461a

dhiller (approved):
/approve

/hold to double check for the exported variable being used elsewhere

brianmcarey:
> /approve
> 
> /hold to double check for the exported variable being used elsewhere

This exposed variable is not being used elsewhere. 

There is another `SemVerMinorRegex` that is used elsewhere.
https://github.com/kubevirt/project-infra/blob/5b8ee3f1b9e4415aa0993b1e6da4bc58fdfb4f46/robots/pkg/querier/queries.go#L15

/unhold

PR #2634, opened 2023-02-27
k3d-sriov: Replace kind with k3d for SR-IOV jobs

oshoval:
Add k3d-sriov presubmit and periodic jobs.
Switch kubevirt presubmit kind with k3d as the gating job.
Disable gating of kubevirtci kind-1.23-sriov presubmit (k3d-sriov is gating already).
Fix name typo of kubevirt kind periodic

Kept the kind-1.23-sriov kubevirt periodic for now.

Depends on 
https://github.com/kubevirt/kubevirtci/pull/972 (need to be [bumped](https://github.com/kubevirt/kubevirt/pull/9384) as well)
https://github.com/kubevirt/kubevirt/pull/9326

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml
@@ -502,7 +502,7 @@ periodics:
     preset-dind-enabled: "true"
     preset-docker-mirror-proxy: "true"
     preset-shared-images: "true"
-  name: periodic-kubevirt-e2e-k8s-1.23-sriov
+  name: periodic-kubevirt-e2e-kind-1.23-sriov
   reporter_config:
     slack:
       job_states_to_report: []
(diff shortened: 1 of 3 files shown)


EdDev (commented):
Thank you for this!

I think we need an intermediate period on which this jobs runs periodically and without voting from the PR/s, to assure stability.
Maybe a periodic is enough.

After a week or so, we can do the final switch.

oshoval:
> Thank you for this!
> 
> I think we need an intermediate period on which this jobs runs periodically and without voting from the PR/s, to assure stability. Maybe a periodic is enough.
> 
> After a week or so, we can do the final switch.

Periodic will run just a few times, unless we do it more frequent as periodic
otherwise a week with just 1 per day, isn't much, wdyt?
another option is to run it manually X times, 20 is fine ? more ?
Thought about just switching, and if needed we can switch back

oshoval:
@EdDev 
you can see here that since we speaked yesterday it ran the lane 3 times as part of 
https://prow.ci.kubevirt.io/?job=check-up-k3d-1.25-sriov
which is gating already on kubevirtci

brianmcarey (commented):
/lgtm

oshoval:
Thanks

oshoval:
Thanks

Edy wdyt?

oshoval:
https://github.com/kubevirt/kubevirt/pull/9326 was merged

@brianmcarey 
Can you please approve as well ?

Thanks

brianmcarey (approved):
/approve

PR #2651, opened 2023-03-07
[kubevirtci] Set check-provision-k8s-1.26-centos9 to required

brianmcarey:
k8s-1.26-centos9 is our main kubernetes 1.26 test target - the presubmit to check the provisioning of this provider should be set to required.

/cc @dhiller @enp0s3


dhiller (approved):
/approve

PR #2659, opened 2023-03-14
Update 1.26 kubevirtci image name to match provider name

brianmcarey:
Currently the `shared-images-controller` is failing to pull the k8s-1.26 provider image as the provider[1] was removed and is no longer published. 

Updated to temporarily handle the new provider name for 1.26 - `k8s-1.26-centos9`

Also removed the return - as if it fails to pull the first image, we still want it to continue to try to download the rest. 

[1] https://github.com/kubevirt/kubevirtci/pull/969

/cc @dhiller @xpivarc

### DIFF
--- a/images/shared-images-controller/main.go
+++ b/images/shared-images-controller/main.go
@@ -79,7 +79,13 @@ func pullRequiredImages(ctx context.Context, tag string) error {
 
 	for _, version := range versions {
 		log.Infoln("Kubevirt Provider version: ", version)
+
 		name := fmt.Sprintf("%sk8s-%s:%s", kubevirtci_repo, version, tag)
+		// k8s-1.26 kubevirtci provider was renamed to k8s-1.26-centos9
+		// to track the change to CentOS Stream 9
+		if version == "1.26" {
+			name = fmt.Sprintf("%sk8s-%s-centos9:%s", kubevirtci_repo, version, tag)
+		}
 		if _, exists := imageNames[name]; exists {
 			log.Infoln("Image already present:", name)
 			continue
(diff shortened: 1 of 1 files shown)


images/shared-images-controller/main.go
```diff
@@ -79,15 +79,21 @@ func pullRequiredImages(ctx context.Context, tag string) error {
...
+		}
 		if _, exists := imageNames[name]; exists {
 			log.Infoln("Image already present:", name)
 			continue
 		}
 		log.Infoln("Pulling image: ", name)
 		_, err := images.Pull(ctx, name, nil)
 		if err != nil {
-			return err
+			log.WithError(err).Errorf("Failed to pull image '%s'", name)
```
xpivarc, line 96: Let's aggregate the errors and let the caller log it as we do now.

images/shared-images-controller/main.go
```diff
@@ -79,15 +79,21 @@ func pullRequiredImages(ctx context.Context, tag string) error {
 
 	for _, version := range versions {
 		log.Infoln("Kubevirt Provider version: ", version)
+
 		name := fmt.Sprintf("%sk8s-%s:%s", kubevirtci_repo, version, tag)
+		// k8s-1.26 kubevirtci provider was renamed to k8s-1.26-centos9
+		// to track the change to CentOS Stream 9
+		if version == "1.26" {
```
dhiller, line 86: This fixes the issue for exactly this version.

How about we first query quay about what provider versions are there and then just pull them all? This way we would never have this issue occur again, regardless of what suffix is present? WDYT?

images/shared-images-controller/main.go
brianmcarey: @xpivarc - I was hoping to just log the error when it happens and move on to pull the next image - why would I aggregate the errors? If it fails to pull an image - we still want it to try to pull the rest.

xpivarc: Ok, fair enough. I thought we might be doing some logic based on returned error. We are not doing it.

images/shared-images-controller/main.go
dhiller: OK, we might take this as an improvement for the future.

PR #2670, opened 2023-03-22
kubevirt, periodics, SR-IOV: Reduce rate

oshoval:
Currently we are running both k3d and kind periodics, each one twice a day.
We even run them on the same time so they might [fall](https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/logs/periodic-kubevirt-e2e-k8s-k3d-1.25-sriov/1638206139675971584) to schedule.

Reduce the rate, as we are running k3d on every presubmit,
in order to save resources during work hours.
Increase the delta between the two periodics in order to reduce
collisions.

Kind was changed to `"At 23:00 on Saturday”`.
K3d was changed to `“At 21:00”`.

Note:
We can even consider to remove kind periodic, 
or maybe bump it from 1.23 to 1.25 but without cpu manager
tests and config, and keep the periodic of it.

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml
@@ -489,7 +489,7 @@ periodics:
     testgrid-dashboards: kubevirt-periodics
     testgrid-days-of-results: "60"
   cluster: prow-workloads
-  cron: 0 22, 11 * * *
+  cron: 0 23 * * 6
   decorate: true
   decoration_config:
     grace_period: 30m0s
@@ -571,7 +571,7 @@ periodics:
     testgrid-dashboards: kubevirt-periodics
     testgrid-days-of-results: "60"
   cluster: prow-workloads
-  cron: 0 22, 12 * * *
+  cron: 0 21 * * *
   decorate: true
   decoration_config:
     grace_period: 30m0s


github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml
```diff
@@ -489,7 +489,7 @@ periodics:
     testgrid-dashboards: kubevirt-periodics
     testgrid-days-of-results: "60"
   cluster: prow-workloads
-  cron: 0 22, 11 * * *
+  cron: 0 23 * * 5,6
```
orelmisan, line 492: @oshoval Could you please explain the meaning of these numbers?

orelmisan: I know they are the cron format, but what it the meaning to humans?

oshoval: https://crontab.guru
adding comments are prone to get outdated so it doesn't make sense
"“At 23:00 on Friday and Saturday.”

orelmisan: > adding comments are prone to get outdated so it doesn't make sense

It was not my intention to request addition of a comment.
But I think adding a human readable explanation to the commit message could be beneficial.

oshoval: We are using cron all over, I think it is an overkill

orelmisan: I know the cron format (not by heart though)
I just think that when something is not straight-forward, its better to write what it does.
It will also help the reviewer to verify if what you did is equal to what you meant to do.

oshoval: To ease the reviewer I supplied the string as mentioned above. and not just pasted the site.
I could just add it but imo it isn't a good practice,
it isn't something we are doing and I don't think we should start.

oshoval:
Added the cron explanation

orelmisan:
> Added the cron explanation

Thank you.

github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml
orelmisan, line 492: Please consider spreading it across the week, so it will have a better chance of catching problems earlier.
AFAIK Less code is being pushed during the weekend so it will probably test the same code on Friday and Saturday.

oshoval: We don't want to run it when there are other jobs, it just take resources
It is a lane that is going to be deprecated, it is just step before.

oshoval: I reduced it to one time a week, there is no need for more, and then we can agree maybe easier about this discussion

brianmcarey (commented):
/approve

orelmisan (approved):
Thank you @oshoval

PR #2682, opened 2023-03-27
Include podman log in output when podman socket fails to start

brianmcarey:
There are occasions when the podman socket fails to start within the specified timeout period. This looks to happen when a number of jobs start all at once on a node and there is high load. This is very difficult to reproduce locally.

Including the podman log will help to troubleshoot issues with the podman socket starting in time.

/cc @xpivarc @dhiller


dhiller (approved):
/approve

PR #2686, opened 2023-03-28
Remove deprecated options from npm command

cwilkers:
Current Dockerfile fails in user-guide `make build_img` due to changes in npm's tolerance for deprecated options. Removing those deprecated options.


cwilkers:
CC @dhiller 
Hi, we have a session Thursday where we will be using this workflow, can you help?

dhiller (approved):
/approve

PR #2690, opened 2023-03-30
Remove mykaul from kubevirt org

brianmcarey:
mykaul is no longer involved with the KubeVirt community so removing from the github org

/cc @dhiller @dankenigsberg


dhiller:
> We should at least ping @mykaul for his awareness and to give him the opportunity to weigh in.

BTW this opens a pandoras box. How do we decide whom to remove from the org? Who is even allowed to remove people from the org?

dhiller (approved):
User requested to be removed on another channel.

/approve

PR #2692, opened 2023-03-30
Make the 1.27 check-provision presubmit required

brianmcarey:
The k8s-1.27 provider has been merged[1] and the check-provision presubmit is passing. We should make this job required so that we can keep this provider in a ready state for when it is needed.

[1] https://github.com/kubevirt/kubevirtci/pull/988

/cc @dhiller


dhiller (approved):
/approve

Thank you @brianmcarey !

PR #2701, opened 2023-04-12
Remove check provision lane for 1.25 PSA as no longer needed

brianmcarey:
/cc  @xpivarc  @dhiller

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/kubevirtci/kubevirtci-presubmits.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/kubevirtci/kubevirtci-presubmits.yaml
@@ -264,32 +264,6 @@ presubmits:
           privileged: true
       nodeSelector:
         type: bare-metal-external
-  - always_run: false
-    cluster: prow-workloads
-    decorate: true
-    decoration_config:
-      timeout: 3h0m0s
-    labels:
-      preset-docker-mirror-proxy: "true"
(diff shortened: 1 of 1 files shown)


github/ci/prow-deploy/files/jobs/kubevirt/kubevirtci/kubevirtci-presubmits.yaml
```diff
@@ -264,32 +264,6 @@ presubmits:
...
-      preset-podman-in-container-enabled: "true"
-    max_concurrency: 2
-    name: check-provision-k8s-1.25-psa
-    spec:
-      containers:
-      - command:
-        - /usr/local/bin/runner.sh
-        - /bin/sh
-        - -c
-        - cd cluster-provision/k8s/1.25 && KUBEVIRT_PSA='true' ../provision.sh
```
xpivarc, line 283: Do we need to add `&& KUBEVIRT_PSA='true'` into `check-provision-k8s-1.25`?

brianmcarey: From a quick look - PSA is only enabled on the 1.26 kubevirt presubmits so I don't think we should enable it for the 1.25 provider check.

xpivarc: We don't need to

PR #2704, opened 2023-04-13
Update 1.26 cgroupsv2 lanes to use 1.26-centos9 provider

brianmcarey:
The old 1.26 provider is no longer available which causes the 1.26 cgroupsv2 presubmit lanes to fail[1] - update these to use the new 1.26-centos9 provider.

[1] https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/9591/pull-kubevirt-e2e-k8s-1.26-sig-storage-cgroupsv2/1646409132577656832#1:build-log.txt%3A51

/cc @dhiller @enp0s3


xpivarc (approved):

xpivarc:
/lgtm cancel
Actually isn't centos9 on v2 by default. We probably can just drop these.

xpivarc:
Ok, I just checked and it is by default.
@brianmcarey We can just remove these.

brianmcarey:
> Ok, I just checked and it is by default. @brianmcarey We can just remove these.

Updated to remove the lanes instead.

xpivarc:
/lgtm
Thanks

brianmcarey (commented):
/approve

PR #2709, opened 2023-04-17
Move project-infra prowjobs from old workloads cluster

brianmcarey:
This change moves the lighter jobs to the managed control plane cluster and the heavier jobs to the new workloads cluster. 

/cc @dhiller


dhiller (approved):
/approve

PR #2716, opened 2023-04-19
Move kubevirtci vm provider check provisions back to old cluster

brianmcarey:
These lanes appear to be a bit flaky on the new cluster - move back to old cluster while issue is investigated.

/cc @dhiller


dhiller (approved):
/approve

PR #2664, opened 2023-03-15
Add files to deploy bazel cache and image cache to new workloads cluster

brianmcarey:
/cc none

PR #2719, opened 2023-04-19
sriov-passthru-cni: Enable allocation restriction of system reserved PFs

ormergi:
Some nodes external network interfaces have SR-IOV capabilities, including the cluster network interface.
Nothing prevent the CNI from allocating such interfaces, which if they do get allocated, the node will have network disconnection.

To prevent such cases, the 'SYSTEM_RESERVED_PFS' env var is introduced to enable users specify the interfaces the CNI should not allocate.

The CNI will not allocate interfaces that specified in the 'SYSTEM_RESERVED_PFS' env var.

The 'SYSTEM_RESERVED_PFS' env var value is expected to be in the form of comma separated list: "eno1,eno2,eno3".

### DIFF
--- a/cni-plugins/sriov-passthrough-cni/README.md
+++ b/cni-plugins/sriov-passthrough-cni/README.md
@@ -68,3 +68,12 @@ requests:
 limits:
   prow/sriov: "2"
 ```
+
+# Reserved system SR-IOV PFs
+Some nodes external network interfaces have SR-IOV capabilities, including the cluster network interface.
+The CNI may allocate such interfaces and cause network disconnection to the node.
+
+In order to prevent from the CNI to allocate such interfaces, set the `SYSTEM_RESERVED_PFS` with the interfaces that CNI should not allocate in a comma seperated form:
+```bash
+SYSTEM_RESERVED_PFS="eno1,eno2,eno3"
+```
--- a/cni-plugins/sriov-passthrough-cni/plugin/sriov-passthrough-cni
+++ b/cni-plugins/sriov-passthrough-cni/plugin/sriov-passthrough-cni
@@ -6,6 +6,8 @@ readonly GW_IP=169.255.0.1
 readonly CONTAINER_NETNS_LINK="/var/run/netns/$CNI_CONTAINERID"
 readonly LOGFILE=/var/log/pfcni.log
 readonly PF_COUNT=1
+# SYSTEM_RESERVED_PFS comma seperated list of interfaces that will not be allocated, e.g: "enp1f1,enp1f2,.."
+readonly SYSTEM_RESERVED_PFS=""
 
 function main() {
     case "$CNI_COMMAND" in
@@ -38,6 +40,11 @@ function setns_sriov_pfs() {
   for pf in "${sriov_pfs[@]}"; do
     local pf_name="${pf%%/device/*}"
     pf_name="${pf_name##*/}"
+
+    if reserved_pf "${pf_name}"; then
+      continue
+    fi
+
     if ip link set "$pf_name" netns "$CNI_CONTAINERID"; then
       if timeout 5s bash -c "until ip netns exec $CNI_CONTAINERID ip link show $pf_name > /dev/null; do sleep 1; done"; then
         log "Added $pf_name to netns $CNI_CONTAINERID"
@@ -52,6 +59,13 @@ function setns_sriov_pfs() {
 
   log "FATAL: Could not allocate enough PFs"
   exit 1
+}
+
+function reserved_pf() {
+    local -r pf=$1
+    if [ -n "${SYSTEM_RESERVED_PFS}" ]; then
+        grep -w -q ${pf} <<< "${SYSTEM_RESERVED_PFS}"
+    fi
 }
 
 function gen_result_config() {


brianmcarey (commented):
/lgtm

cni-plugins/sriov-passthrough-cni/README.md
```diff
@@ -68,3 +68,12 @@ requests:
...
   prow/sriov: "2"
 ```
+
+# Reserved system SR-IOV PFs
+Some nodes external network interfaces have SR-IOV capabilities, including the cluster network interface.
+The CNI may allocate such interfaces and cause network disconnection to the node.
+
+In order to prevent from the CNI to allocate such interfaces, set the `SYSTEM_RESERVED_PFS` with the interfaces that CNI should not allocate in a comma seperated form:
+```bash
+SYSTEM_RESERVED_PFS="eno1,eno2,eno3"
```
oshoval, line 78: We can consider a configmap that is mounted as a file
then we don't need to change the image
and can have empty config map where not needed (as the mount would fail otherwise)

We can also consider using MAC address in case we see that naming aren't good enough
due to ungraceful detach, that cause nics to be renamed - but we can start without it, and act only if
it happens.

cni-plugins/sriov-passthrough-cni/plugin/sriov-passthrough-cni
```diff
@@ -38,6 +40,12 @@ function setns_sriov_pfs() {
   for pf in "${sriov_pfs[@]}"; do
     local pf_name="${pf%%/device/*}"
     pf_name="${pf_name##*/}"
+
+    if reserved_pf "${pf_name}"; then
+      log "PFs ${pf_name} is system reserved, skipping.."
```
oshoval, line 45: I don't think we need this log, it will print it a lot and it isn't helpful

brianmcarey: In the case of the new cluster - this will be 4 extra log lines which will provide useful information and only prints if we have `SYSTEM_RESERVED_PFS` set.

oshoval: for each CNI run / each job which is a lot
open please the file and see why it is not needed imo
It doesn't add info as this is the same thing again and again

cni-plugins/sriov-passthrough-cni/README.md
oshoval: so we need to remove the `SYSTEM_RESERVED_PFS=""` isnt it ?

cni-plugins/sriov-passthrough-cni/plugin/sriov-passthrough-cni
brianmcarey: Yes makes sense to me now - we don't need to log this each time a test pod is deployed.

cni-plugins/sriov-passthrough-cni/README.md
oshoval: > > We can consider a configmap that is mounted as a file then we don't need to change the image and can have empty config map where not needed (as the mount would fail otherwise)
> 
> We achieve the same with the env var, I prefer to start with an env var and if we see that a config-map is necessary we can change it later.
> 
> > Each machine will have its own configmap, with the same image.
> 
> Since the CNI is deployed using a daemonSet, the same config map will be mounted to all CNI pods, similar to the env var.

The configmap can be different on each machine in case needed,
but as long as we don't need we can start with env var.
Note that the old cluster will not support the repo, it is just need to be known.
 
> 
> > We can also consider using MAC address in case we see that naming aren't good enough due to ungraceful detach, that cause nics to be renamed - but we can start without it, and act only if it happens. The MAC can also use a configmap in case needed, so it will be the easiest way to add a file to the CNI pod.
> 
> Note that the interfaces specified in the env var will not be allocated by the CNI, and their names are note prone to be renamed.

Of course, I meant the real ones, they will be detached and might change their name into a prefix that is bad,
but if you use a full name it won't happen, unless a real one comes and goes with different name (due to bug or so) and then it won't be filtered, we can start without and handle of course only if it happens.

cni-plugins/sriov-passthrough-cni/plugin/sriov-passthrough-cni
ormergi: Is there a shortage of storage for logs? if we do, we have bigger problems then this log message..
If not, I don't think we need to be hard on this log message, which could eventually help with debugging.

Note that removing this log message will make debugging harder when something goes wrong.

brianmcarey:
The daemonset you would have to change is part of another PR - here https://github.com/kubevirt/project-infra/pull/2664/commits/96b6f37ab30bdf6891d355762c09eb376824f8ca#diff-2eeb451e294b99f6885a77ec2172d26d41e04a2da0d5409456ac7a0163156081R12

brianmcarey:
> > The daemonset you would have to change is part of another PR - here [96b6f37#diff-2eeb451e294b99f6885a77ec2172d26d41e04a2da0d5409456ac7a0163156081R12](https://github.com/kubevirt/project-infra/commit/96b6f37ab30bdf6891d355762c09eb376824f8ca#diff-2eeb451e294b99f6885a77ec2172d26d41e04a2da0d5409456ac7a0163156081R12)
> 
> Why not on the same PR ? It is in this repo isn't it ? it will be create a full context

It hasn't been merged into the repo yet - still an active PR. 
https://github.com/kubevirt/project-infra/pull/2664

oshoval:
> > > The daemonset you would have to change is part of another PR - here [96b6f37#diff-2eeb451e294b99f6885a77ec2172d26d41e04a2da0d5409456ac7a0163156081R12](https://github.com/kubevirt/project-infra/commit/96b6f37ab30bdf6891d355762c09eb376824f8ca#diff-2eeb451e294b99f6885a77ec2172d26d41e04a2da0d5409456ac7a0163156081R12)
> > 
> > 
> > Why not on the same PR ? It is in this repo isn't it ? it will be create a full context
> 
> It hasn't been merged into the repo yet - still an active PR. #2664

Ok, mixed with the old one `github/ci/prow-deploy/kustom/overlays/workloads-production/resources/multus-cni.yaml`

cni-plugins/sriov-passthrough-cni/README.md
ormergi: > The configmap can be different on each machine in case needed,
but as long as we don't need we can start with env var.
Note that the old cluster will not support the repo, it is just need to be known.

IIUC it means that there will be config-map object per each node, all config-maps should be mounted to the CNI pod, and the CNI should check against all files before allocation.
I think it's harder to maintain.

> Of course, I meant the real ones, they will be detached and might change their name into a prefix that is bad,
but if you use a full name it won't happen unless a real one comes and goes with different name (due to bug or so) and then it won't be filtered, we can start without and handle of course only if it happens.

If you meant for interfaces that will get allocated, in case their name change, the  CNI will still not allocate the reserved interfaces (which are specified in the env var).

cni-plugins/sriov-passthrough-cni/plugin/sriov-passthrough-cni
ormergi: How about printing the reserved interface names (the env var value) before starting the loop?

This way it won't repeat if there is more than one iteration.
We will see this message per ADD command.

oshoval: We don't need it, it is fixed in the PR of the daemonset

cni-plugins/sriov-passthrough-cni/README.md
oshoval: > > The configmap can be different on each machine in case needed,
> > but as long as we don't need we can start with env var.
> > Note that the old cluster will not support the repo, it is just need to be known.
> 
> IIUC it means that there will be config-map object per each node, all config-maps should be mounted to the CNI pod, and the CNI should check against all files before allocation. I think it's harder to maintain.
> 

You will have configmap per SR-IOV BM.
In case each BM need other list it won't able to use correct solution.
The prow job that deploy the cluster will know to set those configmaps.
The configmap will be mounted as a file, and then the script will read it to build the black list.
If the current static list would be enough there is no need to.

> > Of course, I meant the real ones, they will be detached and might change their name into a prefix that is bad,
> > but if you use a full name it won't happen unless a real one comes and goes with different name (due to bug or so) and then it won't be filtered, we can start without and handle of course only if it happens.
> 
> If you meant for interfaces that will get allocated, in case their name change, the CNI will still not allocate the reserved interfaces (which are specified in the env var).

This what I wrote "but if you use a full name it won't happen..."
beside the little corner case, which is just theoretical atm, so we can ignore it unless it happens.

oshoval (approved):
Thanks

nit, you can use markdown for the env var names on the PR desc

brianmcarey (approved):
/approve

brianmcarey:
Image built: `quay.io/kubevirtci/sriov-passthrough-cni:v1.2`

PR #2723, opened 2023-04-20
sriov-passtru-cni: Do not override SYSTEM_RESERVED_NICS env var

ormergi:
Following #2719, make the CNI to not override the `SYSTEM_RESERVED_NICS` env var.

The `SYSTEM_RESERVED_NICS` should be set in the context where the CNI runs, if it is deployed using a  DaemonSet it should be specified in the spec.


oshoval (approved):
Please add in the PR desc that it will be set on the daemonSet

PR #2740, opened 2023-05-02
Update bootstrap image to make podman in container setup more resilient

brianmcarey:
- Increase wait for podman socket to be ready - When a large numbers of jobs start on a node at once, there is high disk
IO which causes the podman socket to take longer to reach a ready
state[1] - increase the timeout to allow for this.

- Add upstream dns server for podman in container network - This helps to reduce the dependency on the host CI clusters DNS and
reduces the number of search domains that is included in the podman in
podman containers.

[1] https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/9268/pull-kubevirt-e2e-k8s-1.24-sig-storage/1653301111404105728

/cc @dhiller

### DIFF
--- a/images/bootstrap/containers.conf
+++ b/images/bootstrap/containers.conf
@@ -1,5 +1,5 @@
 [containers]
-netns="host"
+netns="private"
 userns="host"
 ipcns="host"
 cgroupns="host"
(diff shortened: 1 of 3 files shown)


dhiller (approved):
/approve

PR #2746, opened 2023-05-03
Move kubevirtci prowjobs to new workloads cluster

brianmcarey:
/cc @dhiller


brianmcarey:
There looks to be an issue with the centos8 providers and another issue deleting a test namespace on the 1.27 provider. These issues are seen on both clusters.

brianmcarey:
@dhiller the issues here are not host cluster related as they are seen on both clusters - are we ok to take this in?

dhiller:
> @dhiller the issues here are not host cluster related as they are seen on both clusters - are we ok to take this in?

Yes, agreed.

/lgtm
/approve

PR #2747, opened 2023-05-03
Increase max concurrency of vpgu e2e lane to 2

brianmcarey:
There are two nodes that have GPUs for vgpu testing. The anti-affinity should stop the jobs from being scheduled on the same node so the max concurrency should be increased from 1 to make use of our GPU nodes.

/cc @jean-edouard @fossedihelm @dhiller


dhiller (approved):
/approve

PR #2750, opened 2023-05-05
Move containerdisks e2e jobs to new workloads cluster

brianmcarey:
Theses jobs should be moved to the new workloads cluster as the old cluster has been reduced in size to prepare for decommissioning.

/cc @0xFelix


0xFelix (approved):
/lgtm
/approve

PR #2763, opened 2023-05-10
Move sig-network lanes for 0.58 back to old workloads cluster

brianmcarey:
The 0.58 sig network lanes are hitting DNS issues on the new cluster

Move them to the old cluster until the issue is resolved

/cc @xpivarc @dhiller


brianmcarey (commented):
/approve 

Needed to unblock some backports.

Will look at what is required to move these to the new cluster.

PR #2766, opened 2023-05-11
Skip reporting on the failing fossa presubmit check

brianmcarey:
The fossa check has been failing constantly for the past couple of months so it is not adding any useful feedback to PRs - skip reporting the presubmit result until the issue is resolved.

/cc @dhiller @xpivarc


dhiller (approved):
@brianmcarey thank you for this! :tada: 

/approve

PR #2772, opened 2023-05-16
kubevirt, presubmit: Bump kind SR-IOV to k8s-1.27 and re-enable instead of k3d

oshoval:
Depends on kind-1.27-sriov lands on kubevirt https://github.com/kubevirt/kubevirt/pull/9616


github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
```diff
@@ -508,11 +429,11 @@ presubmits:
       - command:
         - /usr/local/bin/runner.sh
         - /bin/bash
-        - -c
+        - -ce
```
oshoval, line 432: Note:

This is a bug with github, I didn't change it, it links between the deleted job (had `-c` and the updated job (has and still have `-ce`)

Both are fine because there is only one command (`-ce` is better because it is robust to commands addding and this what we have)

brianmcarey (approved):
/approve

PR #2775, opened 2023-05-17
mirror: Add kindest/node:v1.27.1

oshoval:
Mirror mentioned image to quay.


oshoval:
/cc @brianmcarey @dhiller 

We need it for new kind SR-IOV provider please

oshoval:
Thanks

Note that the mirrored image doesn't work, something wasn't done correct

oshoval:
Note that there are no rate limits anymore for kind node, so we can remove those from the list

PR #2777, opened 2023-05-18
setup sig-scale owners for perf-report-creator

alaypatel07:
Copying the sig-scale owners configuration from kubevirt/kubevirt.


robots/cmd/perf-report-creator/OWNERS
```diff
@@ -0,0 +1,6 @@
+reviewers:
+  - sig-scale-reviewers
```
alaypatel07, line 2: @dhiller I wonder if `- ci-maintainers` should be an explicit item here or will it be inherited from top level permissions?

dhiller: The [docs say that they are inherited](https://docs.prow.k8s.io/docs/components/plugins/approve/approvers/#overview), but looking at the top level files, I think the setup is not correct.

dhiller: Should be fixed with https://github.com/kubevirt/project-infra/pull/2779

alaypatel07: looks like #2779 merged, I'll rebase and push

dhiller (approved):
/approve

PR #2785, opened 2023-05-25
Increase sig-monitoring e2e test timeout

0xFelix:
This increases the timeout of the sig-monitoring e2e tests to 2h0m0s so the whole testsuite has enough time to finish.

Fixes CI failures in https://github.com/kubevirt/kubevirt/pull/9724

/cc @brianmcarey


brianmcarey (approved):
/approve

PR #2787, opened 2023-05-25
add scrape.sh to automate generating weekly perf-report

alaypatel07:
Add a script to scrape results of two jobs and aggregate it into weekly folder
as well as copy it into performance benchmark directory


alaypatel07:
@dhiller this will make #2773 easier to maintain long term. The only thing now pending is to create a graph every week

PR #2790, opened 2023-05-26
Remove 1.24 kubevirt test prow jobs

brianmcarey:
The 1.27 test lanes have been introduced so it is time to remove the 1.24 lanes. 

/cc @dhiller @xpivarc 

Closes #2789


dhiller (approved):
Thanks @brianmcarey , especially because the automation doesn't handle the old operator naming well!

Also thanks for the removal of the cgroupsv2 and realtime lanes.

/approve

/hold

@xpivarc do you agree that we can remove them?

PR #2792, opened 2023-05-29
Remove old prowjobs that target the old workloads cluster

brianmcarey:
This change removes the jobs that were used to deploy the old workloads cluster. This cluster is due to be removed in the near future so these jobs are no longer required.

Also removed are leftover jobs from the old ARM cluster which are no longer used.

/cc @dhiller @enp0s3


enp0s3 (approved):
/approve

PR #2796, opened 2023-05-31
Remove old workload clusters deployment files

brianmcarey:
- Remove the ansible files that were used for bootstrapping and installing
kubernetes to the old workloads cluster and the old ARM cluster. The ARM
cluster no longer exists and the old prow workloads cluster is due to be
removed soon. The automated jobs that use these files were removed in a
previous PR[1]

- The old workloads cluster is no longer taking any prow jobs and will be
removed soon. Deploying updates to the prow components in this cluster
is no longer required so these files can be removed. These are deployed to the new cluster as part of this PR [2]

[1] https://github.com/kubevirt/project-infra/pull/2792
[2] https://github.com/kubevirt/project-infra/pull/2664

/cc @enp0s3 @dhiller


dhiller (approved):
/approve

PR #2805, opened 2023-06-05
WIP: try to use kind for sig-network

oshoval:


oshoval:
Doesnt worth the effort

PR #2813, opened 2023-06-07
Move check-provision-manager to kubevirt-prow-workloads cluster

brianmcarey:
The `prow-workloads` cluster no longer exists so this job will never get scheduled. Update to target `kubevirt-prow-workloads`.

/cc @oshoval @dhiller


oshoval (approved):
Thanks

oshoval (approved):

oshoval (approved):

PR #2814, opened 2023-06-07
Add KubeVirt fuzzing presubmit

rmohr:
Optional fuzzing job for https://github.com/kubevirt/kubevirt/pull/9866.


rmohr:
Nice improvements on rehearse :)

brianmcarey:
> Hm, the output is not very convincing of the test. Let me see that I can ensure that reasonable test output is printed. My suspicion is that a permutation is failing because of slowness due to not enough cpu.

If you wanted test with plenty of cpu - you could rehearse against the workloads cluster - `kubevirt-prow-workloads`.

PR #2815, opened 2023-06-08
Update e2e presubmits to run etcd in memory

brianmcarey:
When there is high disk IO on a node, we can sometimes see jobs fail due to etcd timeouts[1][2]. Update these jobs to run etcd in memory to reduce reliance on node disks.

`Error from server: error when creating "/home/prow/go/src/github.com/kubevirt/kubevirt/_out/manifests/release/kubevirt-cr.yaml": etcdserver: request timed out`

[1] https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/9848/pull-kubevirt-e2e-k8s-1.27-sig-operator/1666679307365257216#1:build-log.txt%3A2065
[2] https://search.ci.kubevirt.io/?search=etcdserver%3A+request+timed+out&maxAge=336h&context=1&type=build-log&name=&excludeName=&maxMatches=5&maxBytes=20971520&groupBy=job

/cc @dhiller @EdDev @xpivarc

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
@@ -33,6 +33,8 @@ presubmits:
           value: k8s-1.25-sig-compute
         - name: KUBEVIRT_CGROUPV2
           value: "true"
+        - name: KUBEVIRT_WITH_ETC_IN_MEMORY
+          value: "true"
         image: quay.io/kubevirtci/bootstrap:v20230502-bfaa042
         name: ""
         resources:
@@ -110,6 +112,8 @@ presubmits:
         - name: TARGET
           value: k8s-1.25-sig-storage
         - name: KUBEVIRT_CGROUPV2
+          value: "true"
+        - name: KUBEVIRT_WITH_ETC_IN_MEMORY
           value: "true"
         image: quay.io/kubevirtci/bootstrap:v20230502-bfaa042
         name: ""
@@ -915,6 +919,8 @@ presubmits:
           value: k8s-1.25-sig-network
         - name: KUBEVIRT_SINGLE_STACK
           value: "true"
+        - name: KUBEVIRT_WITH_ETC_IN_MEMORY
+          value: "true"
         image: quay.io/kubevirtci/bootstrap:v20230502-bfaa042
         name: ""
         resources:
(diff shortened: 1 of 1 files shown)


xpivarc:
Did we measure the required memory?

brianmcarey:
> Did we measure the required memory?

When testing the sig-compute lane last night - it got to about 400M - the default size is 512M

xpivarc (approved):

brianmcarey:
Rehearsals look good here. 

/hold

Just putting a hold to allow for further discussion on this.

github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
```diff
@@ -33,6 +33,8 @@ presubmits:
           value: k8s-1.25-sig-compute
         - name: KUBEVIRT_CGROUPV2
           value: "true"
+        - name: KUBEVIRT_WITH_ETC_IN_MEMORY
```
dhiller, line 36: > > Did we measure the required memory?
> 
> When testing the sig-compute lane last night - it got to about 400M - the default size is 512M

Do we want to set a bigger default value for this to be safe? We could hide that default in a new preset, this way we would need to change it only once?

brianmcarey:
> 

The overall job times of the rehearsals look to be about the same as what is currently running so no major improvement there. 

This change would just help avoid hitting that etcd timeout flake which can sometimes fail batches etc.

github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
brianmcarey: I was going to suggest to stick with the default value for now and only increase it if we see a need to as memory in the CI cluster is an expensive resource.

dhiller:
> > Did we measure the required memory?
> 
> When testing the sig-compute lane last night - it got to about 400M - the default size is 512M

Do we want to set a bigger default value for this to be safe? We could hide that default in a new preset, this way we would need to change it only once?

> > I would really want to see a comparison about how the overall job performance is affected. However if you have looked at the results and they look better, everything that relieves the cluster should be good.
> 
> The overall job times of the rehearsals look to be about the same as what is currently running so no major improvement there.
> 
> This change would just help avoid hitting that etcd timeout flake which can sometimes fail batches etc.

Sounds good to me then.

github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
dhiller: SGTM.

dhiller (approved):
/approve

PR #2827, opened 2023-06-13
Add ppc64le support in kubevirt hostpath provisioner

awels:
Had a request for a ppc64le build, they verified that setting the GOARCH was enough to make it work. So it seems relatively straight forward to add it to the post submit jobs to help them out.


aglitke (commented):
/lgtm

PR #2852, opened 2023-06-26
Use --all flag when cleaning up podman containers in bootstrap pod

brianmcarey:
We should rely on podman to cleanup all of the created containers as part of the e2e tests.

`podman rm --all --force`

The above should remove all containers and we should not hit the dependent containers issue[1] that we can hit with the current cleanup.

[1] https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/9974/pull-kubevirt-e2e-k8s-1.25-sig-storage-1.0/1673289162947563520#1:build-log.txt%3A3103
[2] https://docs.podman.io/en/latest/markdown/podman-rm.1.html#all-a

/cc @enp0s3 @xpivarc


enp0s3 (approved):
/approve

PR #2854, opened 2023-06-27
Update preset bazel version to match kubevirt builder image

brianmcarey:
As the e2e jobs run the builds unnested - the bazel version set in the preset should match that of the kubevirt builder image[1]

[1] https://github.com/kubevirt/kubevirt/blob/d89207bd558440af2c273870540093951a82e97c/hack/builder/Dockerfile#L6

/cc @xpivarc @enp0s3


xpivarc (approved):

brianmcarey:
/hold 

Just want to do run a test with an e2e lane first.

PR #2855, opened 2023-06-27
common-instancetypes: Add release-0.3 presubmits

lyarwood:


github/ci/prow-deploy/files/jobs/kubevirt/common-instancetypes/common-instancetypes-0.3.yaml
```diff
@@ -0,0 +1,59 @@
+presubmits:
+  kubevirt/common-instancetypes:
+  - name: pull-common-instancetypes
+    branches:
+      - release-0.3
+    always_run: true
+    optional: true
```
brianmcarey, line 7: Are you sure you want this as optional?

lyarwood: That's the behaviour on main at the moment, I'll confirm there and will backport a change to this branch if we can drop it.

PR #2857, opened 2023-06-28
CDI linter lane use golang container

awels:
The linter lane needs to compile some go code soon, modifying the used container to be a golang container


brianmcarey (approved):
/approve

PR #2866, opened 2023-07-03
Switch CDI postsubmit jobs to use podman

maya-r:
It works fine with buildah/podman


brianmcarey (commented):
/lgtm

akalenyu:
Is there a way to rehearse these? I think we had some sort of issue with these image pushing jobs and podman

maya-r:
If not, one of the jobs runs on the latest commit, so we should be able to see a failure pretty much right away after merging

brianmcarey:
> Is there a way to rehearse these? I think we had some sort of issue with these image pushing jobs and podman

We don't have a way of rehearsing postsubmits currently as we don't want rehearsals actually publishing any artifacts etc.

akalenyu (approved):
lgtm

brianmcarey (approved):
/approve

PR #2868, opened 2023-07-04
Configure thanos-query to use thanos-receive as a store for queries

brianmcarey:
Also includes some updates to the thanos-receive configuration to improve performance

/cc @dhiller


dhiller (approved):
/approve

PR #2871, opened 2023-07-06
Increase size of persistent volume attached to thanos receive

brianmcarey:
The previous volume ran out of space
`panic: write header: write /var/thanos/receive/default-tenant/chunks_head/000067.tmp: no space left on device`

/cc @dhiller @xpivarc


dhiller (approved):
/approve

PR #2876, opened 2023-07-07
Fix csi-driver post submit

awels:
Make target had changed and the environment variable indicating which repo to push to has changed as well. Converging on kubevirt-csi-driver repo.


davidvossel (commented):
/lgtm
/approve

PR #2878, opened 2023-07-09
sriov cni: Return sriov pfs to root ns

oshoval:
Upon DEL command (the previous command was wrong), gracefully detach the SR-IOV nics.
The reason is if it isn't done as such, it might
take time until it be available again by the kernel.

Depends on https://github.com/kubevirt/kubevirtci/pull/1040 (merged)


oshoval:
Closing this, in case we will want and have time we can back to it in the future

PR #2880, opened 2023-07-11
Remove sippy deployment from control plane cluster

brianmcarey:
The sippy deployment has not been updated in over two years and I believe that nobody has used it during this time.

It should be removed from the cluster since it has gone unused for so long.

/cc @dhiller @xpivarc @enp0s3


dhiller (approved):
/approve

As a follow up we need to remove the sippy deployment manually.

brianmcarey:
> /approve
> 
> As a follow up we need to remove the sippy deployment manually.

The sippy deployment has been removed from the cluster.

PR #2884, opened 2023-07-13
cnao, presubmit: set multus-dynamic-networks-functests as optional

oshoval:
The lane is flaky, lets set as optional until it is fixed, it will reduce noise.


maiqueb (approved):
Thank you.

PR #2888, opened 2023-07-14
Enable prow for the new kubernetes.kubevirt repository

0xFelix:
$subject


dhiller (approved):
Changes look good to me.

/approve

One thing to consider might be to add a section for your repository in `branch-protection`, so `tide` can deduce which test lanes to be required for merge: https://docs.prow.k8s.io/docs/components/core/tide/config/#context-policy-options

This config will (as the name says) enable protecting the main branch from force-pushes, which might or might not be what you want.

If you want to configure required test lanes manually, you can do as described in the above doc link in the `context_options`, however you might find the first option easier.

/hold can be removed, if you are good with the changes for now, we can still work on this later.

0xFelix:
@dhiller Thanks for the heads up, I actually missed that.

Updated the PR to include the branch protection.

dhiller (approved):
Thanks! Looks good!

PR #2894, opened 2023-07-18
Increase hostpath provisioner timeout to 3 hours

awels:
because doing qemu compile of arm64 and ppe releases. That takes some time and it is currently timeing out during the build while it is progressing.


brianmcarey (commented):
you may need to move this to the workloads cluster too
`cluster: kubevirt-prow-workloads`

but the new timeout might be enough - lets see how it goes.

awels:
Good point, let me do that, that should help

brianmcarey (commented):
/lgtm

PR #2899, opened 2023-07-20
Add CDI ceph waitforfirstconsumer lane

akalenyu:
More storage providers use WaitForFirstConsumer not just for local volumes,
but also generally to do things like availability zone distribution and so forth.
Let's make sure we integrate well with those.

Currently we only test HPP with WaitForFirstConsumer and
that leaves some open integration points like snapshots/csi clones

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/containerized-data-importer/containerized-data-importer-presubmits.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/containerized-data-importer/containerized-data-importer-presubmits.yaml
@@ -305,6 +305,44 @@ presubmits:
           resources:
             requests:
               memory: "29Gi"
+  - name: pull-containerized-data-importer-e2e-ceph-wffc
+    skip_branches:
+      - release-v1.28
+      - release-v1.34
+      - release-v1.38
+      - release-v1.43
+      - release-v1.49
+    annotations:
+      fork-per-release: "true"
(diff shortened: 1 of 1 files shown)


github/ci/prow-deploy/files/jobs/kubevirt/containerized-data-importer/containerized-data-importer-presubmits.yaml
```diff
@@ -305,6 +305,44 @@ presubmits:
           resources:
             requests:
               memory: "29Gi"
+  - name: pull-containerized-data-importer-e2e-ceph-wffc
+    skip_branches:
```
brianmcarey, line 309: Could we shorten this `skip_branches:` with some like we have for the kubevirt jobs? 

https://github.com/kubevirt/project-infra/blob/d346545770114ef120b3d89be58c339b33401038/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml#L23C4-L23C4

akalenyu: I believe we can't - we don't want to skip all of them, just this specific set

brianmcarey: Ok thanks for clearing that up for me.

brianmcarey (approved):
/approve

PR #2903, opened 2023-07-21
Always run the check provision job for 1.28

brianmcarey:
The 1.28 provider was merged[1] so lets enable this lane to always run so that we prevent issues being introduced.

[1] https://github.com/kubevirt/kubevirtci/pull/1059

/cc @dhiller


dhiller (approved):
/approve

PR #2908, opened 2023-07-27
Remove check-provision jobs for fedora test vm images

brianmcarey:
These check-provision jobs do not work when run against the workloads cluster as the cluster nodes are missing btrfs[1] support which is required by newer versions of fedora.

[1] https://github.com/kubevirt/kubevirtci/issues/1004#issuecomment-1649469766

/cc @zhlhahaha @xpivarc


brianmcarey (commented):
/approve

PR #2916, opened 2023-08-02
common-instancetypes: Increase node memory in functest

lyarwood:
This doubles the node memory from the default of 4096 to 8192 so we can start VirtualMachines using the u1.medium instance type that itself requires 4Gi.


lyarwood:
/cc @0xFelix 

Can you approve this when you get a chance?

brianmcarey (approved):
/approve

brianmcarey:
/hold

Feel free to unhold

PR #2921, opened 2023-08-08
Disable required status checks in kubevirt.core

0xFelix:
This is required so the Ansible Middleware release automation is able to create and push the necessary commits.


dhiller (approved):
/approve

PR #2922, opened 2023-08-08
Add usb emulation for 1.25

victortoso:
Hi,

I don't remember why we have not added this in #2872 so I'm starting this thread with a PR :)
Functional tests with 1.26 and 1.27 [worked great!](https://github.com/kubevirt/kubevirt/pull/10015#issuecomment-1669212331) 

Cheers,


brianmcarey (commented):
I believe support for this was not added to the centos stream 8 based providers (so just 1.25) - it was just added to the centos stream 9 providers[1]

With 1.28 around the corner it may not be worth adding to the stream 8 provider. 

@xpivarc - what do you think?


[1] https://github.com/kubevirt/kubevirtci/pull/996

xpivarc:
@brianmcarey 
Might as well be true :)

PR #2944, opened 2023-08-17
Skip ceph wffc lane on release branches

akalenyu:
Whoops - never meant to run this on any release branches. just main is enough


enp0s3 (approved):
/approve
@awels @akalenyu if this lane to some extent does duplicated functionality testing, you can make this lane optional in main as well.

PR #2950, opened 2023-08-23
Generalize flakefinder for future gerrit usage

rmohr:
Move github structs and clients behind interfaces which allow working on gerrit and gerrit Change Requests in the future as well.

* `PullRequest` is now hidden behind an `api.Change` interface
* GitHub client is hidden behind a `api.Query` interface

Checking if a prow job was executed on the latest change was moved behind the `api.Query` interface.

### DIFF
--- a/robots/pkg/flakefinder/downloader_test.go
+++ b/robots/pkg/flakefinder/downloader_test.go
@@ -3,9 +3,8 @@ package flakefinder_test
 import (
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
-
-	"github.com/google/go-github/v28/github"
 	"kubevirt.io/project-infra/robots/pkg/flakefinder"
+	ghapi "kubevirt.io/project-infra/robots/pkg/flakefinder/github"
 )
 
 var _ = Describe("downloader.go", func() {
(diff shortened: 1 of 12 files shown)


robots/pkg/flakefinder/flakefinder.go
```diff
@@ -184,57 +184,26 @@ type ReportBaseData struct {
...
+	changes, err := q.Query(ctx, startOfReport, endOfReport)
+	if err != nil {
+		logrus.Fatal(err)
 	}
-	logrus.Infof("%d pull requests found.", len(prs))
 
 	var reports []*JobResult
-	for _, pr := range prs {
+	var changeNumbers []int
+	for _, pr := range changes {
```
dhiller, line 197: nit
```suggestion
	for _, change := range changes {
```

robots/pkg/flakefinder/downloader_test.go
```diff
@@ -3,9 +3,8 @@ package flakefinder_test
 import (
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
-
-	"github.com/google/go-github/v28/github"
 	"kubevirt.io/project-infra/robots/pkg/flakefinder"
+	github2 "kubevirt.io/project-infra/robots/pkg/flakefinder/github"
```
dhiller, line 7: nit: I find the `github2` ugly and a bit confusing, since the reader has to keep in mind that they are looking at usage of the GitHub implementation for the "repo query api" so to say. I understand that you want to keep it consistent though.

So, since I can't come up with a better solution for this right now myself I guess this improvement is for a later PR.

dhiller (approved):
Just two nits inline, in general these changes look fine to me. I'll leave it up to you if you want to address them. Just unhold in the latter case.

/approve
/hold

robots/pkg/flakefinder/downloader_test.go
rmohr: Done. Renamed to `ghapi`.

robots/pkg/flakefinder/flakefinder.go
rmohr: Done.

rmohr:
/unhold

PTAL

dhiller:
> /unhold
> 
> PTAL

@rmohr was out for a couple of days, will take another look soon!

dhiller (approved):
@rmohr looks great, thanks!

PR #2958, opened 2023-08-28
Set check-provision-k8s-1.25 to always-run false

brianmcarey:
The k8s-1.25 provider will be removed soon. We no longer need to check the provisioning of the provider against every kubevirtci PR.

/cc @dhiller @xpivarc


dhiller (approved):
/lgtm
/approve

PR #2963, opened 2023-08-29
CDI push builder images on release branches

awels:
Sometimes we need to create a specific builder for a release branch (maybe new version of go). Right now it is not possible to push a new builder from a branch This modifies the job to also push from release branches if there is a new builder.


maya-r:
I guess it's not going to break anything since the builder is updated manually to a checksum, but it's hard to tell whether the builder image comes from the release branch or the main branch

awels:
Yes you are correct about that, the naming of the builder doesn't take the release branch into account (since before it didn't actually generate builders in release branches). This should be a rare occasion though.

PR #2965, opened 2023-08-30
Remove PR pending from reports doc as flake stats merged

brianmcarey:
Includes updating the descriptions to all start with an uppercase letter just to match.

/cc @dhiller


dhiller (approved):
@brianmcarey thank you for the fix!

/approve

PR #2968, opened 2023-08-31
Switch docker to podman in prowjob scripts

brianmcarey:
The jobs are really using podman rather than docker so it is more clear for everyone to swap to using podman in the scripts.

`docker` is just symlinked to podman in the bootstrap image[1].

[1] https://github.com/kubevirt/project-infra/blob/33c4878341d4274b58052616bbebd61b5513716e/images/bootstrap/Dockerfile#L84

/cc @enp0s3 @dhiller


dhiller (approved):
/approve

Thank you @brianmcarey :tada:

PR #2978, opened 2023-09-05
Add cri-o version 1.28 to the kubevirt mirror

brianmcarey:
cri-o v1.28.0 has been released[1] - this should be added to the mirror so that it can be installed to the 1.28 kubevirtci provider

[1] https://github.com/cri-o/cri-o/releases/tag/v1.28.0

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/kubevirtci/kubevirtci-periodics.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/kubevirtci/kubevirtci-periodics.yaml
@@ -30,7 +30,7 @@ periodics:
           - name: BASE_REPOID
             value: devel_kubic_libcontainers_stable
           - name: CRIO_VERSIONS
-            value: "1.22,1.23,1.24,1.25,1.26,1.27"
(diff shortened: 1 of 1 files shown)


dhiller (approved):
/approve

@brianmcarey should we extend the automation to automatically mirror new releases?

brianmcarey:
> /approve
> 
> @brianmcarey should we extend the automation to automatically mirror new releases?

I think thats a good idea - we could just query the latest releases on GH. I will create an issue.

dhiller:
> /approve
> 
> @brianmcarey should we extend the automation to automatically mirror new releases?

Maybe we should also look at which versions we still need mirrored. Would it be reasonable to update only the latest 3 + 1 ?

brianmcarey:
> > /approve
> > @brianmcarey should we extend the automation to automatically mirror new releases?
> 
> Maybe we should also look at which versions we still need mirrored. Would it be reasonable to update only the latest 3 + 1 ?

We have hit issues in the past with removing older versions from the mirror - https://github.com/kubevirt/project-infra/pull/2784

PR #2972, opened 2023-09-01
Update kubevirtci to use podman based golang image

brianmcarey:
The go version in the legacy image is getting quite old and can cause the publish job to fail[1].

Lets update this job to use the podman based golang image which has a more up to date version of go.

[1] https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/logs/publish-kubevirtci/1697339447776907264#1:build-log.txt%3A8

/cc @dhiller @xpivarc

PR #2984, opened 2023-09-11
Switch publish kubevirtci prowjob back to docker

brianmcarey:
Providers that have been published since the switch to publishing with podman[1] are failing to start up correctly when run with docker so we need to change back to publishing with docker for now.

[1] https://github.com/kubevirt/project-infra/pull/2972

/cc @dhiller @xpivarc


brianmcarey:
Running a new publish here with docker - https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/logs/publish-kubevirtci/1701137504704401408

dhiller (approved):
/approve

PR #2989, opened 2023-09-18
Move workloads from old control plane cluster to the new cluster

brianmcarey:
This change moves all jobs that were running on cluster `ibm-prow-jobs`
to run on the new cluster `kubevirt-prow-control-plane`

This also add /lib/modules as a read-only hostPath to jobs running with
podman in container enabled. Without this jobs would fail on the new
cluster as netavark was unable to find ipv6 modules


dhiller (approved):
/approve

brianmcarey:
/hold 

until scale up of the new cluster is complete - adding extra nodes to handle the new workload.

PR #2990, opened 2023-09-18
Deploy greenhouse and docker-mirror-proxy to new control plane cluster

brianmcarey:
Greenhouse(Bazel cache) and the docker-mirror-proxy are required on the new cluster in order to enable the migration of workloads from the existing control plane cluster.

/cc @xpivarc @dhiller @enp0s3

### DIFF
--- a/github/ci/prow-deploy/kustom/components/docker-mirror-proxy/pv/patches/JsonRFC6902/deployment.yaml
+++ b/github/ci/prow-deploy/kustom/components/docker-mirror-proxy/pv/patches/JsonRFC6902/deployment.yaml
@@ -11,3 +11,8 @@
     name: storage
     persistentVolumeClaim:
       claimName: docker-mirror-proxy
+- op: add
+  path: /spec/template/spec/securityContext
+  value:
+      runAsUser: 0
+      runAsGroup: 0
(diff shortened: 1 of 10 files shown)


github/ci/prow-deploy/kustom/components/docker-mirror-proxy/pv/patches/JsonRFC6902/deployment.yaml
```diff
@@ -11,3 +11,8 @@
     name: storage
     persistentVolumeClaim:
       claimName: docker-mirror-proxy
+- op: add
```
dhiller, line 14: Why is that necessary now?

brianmcarey: By default the new openshift cluster runs the pods as a non privileged user - the docker-mirror-proxy pod ends up in a error as it tries to access files under /etc

https://github.com/rpardini/docker-registry-proxy/blob/51585675e6543f6e9b81a2b15b15e8f0778119e2/entrypoint.sh#L7

dhiller: Thanks for the explanation!

dhiller (commented):
/approve

dhiller (approved):

PR #2992, opened 2023-09-20
Move ci-search service to the new control plane cluster

brianmcarey:
/cc @dhiller 

/hold  

The DNS has to be updated before merge so that the certs get applied correctly


dhiller (approved):
/approve

PR #2995, opened 2023-09-21
Temporarily move prow deploy test lane back to old cluster

brianmcarey:
There seems to be an issue starting the kind cluster on the new control plane cluster[1]. Temporarily moving back to the old cluster while investigating the issue.

[1] https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_project-infra/2994/pull-project-infra-prow-deploy-test/1704811297931005952

/cc @dhiller


dhiller (approved):
/approve

PR #3017, opened 2023-10-11
Remove cert-manager deployment as no longer required

brianmcarey:
The new control plane cluster no longer requires a cert-manager deployment as it uses the openshift operator for cert-manager.

/cc @dhiller


dhiller (commented):
The [docs say](https://docs.openshift.com/container-platform/4.13/security/cert_manager_operator/cert-manager-operator-install.html) that this is not installed by default. Do we need to upgrade our installation docs to include that this is required? Or is this a feature of the managed offering from IBM?

dhiller (approved):
/approve

brianmcarey:
/override  pull-project-infra-cert-manager-deploy-test 

Lane removed as part of this PR

PR #3023, opened 2023-10-16
[performance] Increase performance job memory

tiraboschi:
PR #10110 introduced an additional container
in each virt-launcher pod to stream out the
logs from the serial console of the guest OS.
This additional container is running a custom
tail binary named virt-tail and so it also
requires resources.
The increased amount of requested resources
(35M of memory for each virt-launcher pod)
was making the performance test lane failing
and so the number of VMs involved in the test was
squezed from 100 to 90 but due to that
the results are now not directly comparable.

In order to be able to revert back to 100
test VMs we need 35M*100=3500M more,
let's round to 4Gi more.


fossedihelm:
@tiraboschi https://github.com/kubevirt/kubevirt/pull/10566 will also increase the memory requests. Any possibility to increase of another `100Mi*100 = 10Gi`, thus increasing up to `70Gi`?
This +10 should also be done for recent release tags but it can be done with dedicated PRs.
Thanks

tiraboschi:
> thus increasing up to `70Gi`? 

Yes, I think so.
@dhiller do we have HW capabilities to run with 70Gi?

xpivarc:
/lgtm
@brianmcarey @dhiller please have a look

rthallisey:
The change is fine with me.  We caught the mem increase in the MR, which is good and it will be in the release.  At some point we need to get this going in the other direction ;).

brianmcarey:
> > thus increasing up to `70Gi`?
> 
> Yes, I think so. @dhiller do we have HW capabilities to run with 70Gi?

The nodes are big enough to take jobs that request 70Gi - it might just take a while for the perf jobs to be scheduled to the cluster during busy periods though. (but that would be the same story with 60Gi too)

tiraboschi:
Thanks @brianmcarey do you prefer to bump it to 70Gi in a single step getting ready in advance for https://github.com/kubevirt/kubevirt/pull/10566 ?

brianmcarey:
> Thanks @brianmcarey do you prefer to bump it to 70Gi in a single step getting ready in advance for [kubevirt/kubevirt#10566](https://github.com/kubevirt/kubevirt/pull/10566) ?

Yes a single step would be simplest. Can you update the PR description with the two reasons for the memory increase so we don't lose the information?

tiraboschi:
> Can you update the PR description with the two reasons for the memory increase so we don't lose the information?

Done

brianmcarey (approved):
/approve

thanks @tiraboschi

brianmcarey:
We probably need to increase the kubevirtci node memory alongside the job memory. 

https://github.com/kubevirt/project-infra/pull/3023/files#diff-26f7cd0389085c1da47f919a02c4dbeca5ddd96ca00d8e49c0862f698475aaf9L83

PR #3024, opened 2023-10-18
Increase performance presubmit node size

xpivarc:
Forgotten in https://github.com/kubevirt/project-infra/pull/3023/

Increasing by 3G per 4 nodes = 12G. Now the total is 15G per 4 nodes = 60G and we are running with 70G requests.

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
@@ -80,7 +80,7 @@ presubmits:
         - name: KUBEVIRT_STORAGE
           value: hpp
         - name: KUBEVIRT_MEMORY_SIZE
-          value: 12G
+          value: 15G
         - name: KUBEVIRT_NUM_NODES
           value: "4"
         - name: KUBEVIRT_DEPLOY_PROMETHEUS


fossedihelm:
Can you explain the correlation between `KUBEVIRT_MEMORY_SIZE` and `resources.requests.memory`?

EdDev (approved):

xpivarc:
> Can you explain the correlation between `KUBEVIRT_MEMORY_SIZE` and `resources.requests.memory`?

The `KUBEVIRT_MEMORY_SIZE` is what memory will be used for VMs and the latter is just for scheduling on our CI cluster (making sure you will not run out of it).

xpivarc:
@brianmcarey I do wonder why we have a larger request than total used memory, would you know? Should I match 70G?

brianmcarey:
> @brianmcarey I do wonder why we have a larger request than total used memory, would you know? Should I match 70G?

We need some overhead for the test pod as the build etc happens in the pod itself.

fossedihelm:
> > Can you explain the correlation between `KUBEVIRT_MEMORY_SIZE` and `resources.requests.memory`?
> 
> The `KUBEVIRT_MEMORY_SIZE` is what memory will be used for VMs and the latter is just for scheduling on our CI cluster (making sure you will not run out of it).

Thanks @xpivarc!


> We need some overhead for the test pod as the build etc happens in the pod itself - probably not 10Gi though.

So `KUBEVIRT_MEMORY_SIZE` * `KUBEVIRT_NUM_OF_NODES` must be lower than `resources.requests.memory`?
Thank you, for the explanation!

PR #3026, opened 2023-10-18
Add acardace to OWNERS of project-infra

brianmcarey:
The automation for creating kubevirt releases requires write access to proeject-infra to create the job configs for the release branches.

Antonio is part of the release team and requires write access to project-infra as a result.

/cc @davidvossel @fabiand @xpivarc

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -2,6 +2,7 @@ filters:
   ".*":
     reviewers:
       - ci-maintainers
+      - acardace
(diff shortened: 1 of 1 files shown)


brianmcarey:
We should look at updating the automation to create the release job configs from the user's fork so these permissions would no longer be required.

davidvossel (commented):
/lgtm

PR #3027, opened 2023-10-18
Add files to deploy new grafana test instance with operator

brianmcarey:
This includes the files to deploy a test grafana instance that can be
used in the presubmit lane to verify changes to the grafana deployment
on the control plane

/cc @dhiller @xpivarc


github/ci/services/grafana/hack/deploy.sh
```diff
@@ -0,0 +1,19 @@
...
+
+main() {
+	local environment=${1}
+
+	bazelisk run //github/ci/services/grafana:$environment.apply
+
+	bazelisk run //github/ci/services/common/k8s/cmd/wait -- -namespace monitoring -selector grafana-deployment -kind deployment
+
+	# Allow some time for the grafana operator to create the required pods & services
+	sleep 10
```
dhiller, line 13: This is pretty brittle, no? Shouldn't we use something like a pod readiness check or maybe even the cr state? 

Note: I've tried this one and it seemed to work
```bash
kubectl wait --for=jsonpath='{.status.stage}'=complete -n openshift-operators grafana grafana
```

dhiller (approved):
Looks good to me! One minor nit inline.

/approve
/hold

github/ci/services/grafana/hack/test.sh
```diff
@@ -0,0 +1,45 @@
...
+    protocol: TCP
+EOF
+
+kubectl create ns monitoring
+
+# Install ingress
+kubectl apply -f https://raw.githubusercontent.com/kubernetes/ingress-nginx/main/deploy/static/provider/kind/deploy.yaml
+
+# Allow time for ingress controller to start
+sleep 20
```
dhiller, line 36: Oops, I overlooked this one, maybe we could do something similar like [there](https://github.com/kubevirt/project-infra/pull/3027/files#diff-3eb05a768dd6189427bc8a4e338855d2075e50d365080900780e770d186aa860R12) here also? 

(Could also be done in a follow up, but you decide)

dhiller (approved):
One nit inline, don't want to block on that one, since I overlooked it myself.

github/ci/services/grafana/hack/test.sh
brianmcarey: I've removed the sleep as no longer needed due to the addition of waiting on the grafana operator - but unfortunately this remove the lgtm. 

Rehearsal looks good.

dhiller (approved):
Thank you @brianmcarey !

PR #3032, opened 2023-10-19
update k8s version to 1.27 for perf-report-creator job scraping

alaypatel07:


brianmcarey (approved):
/approve 

thanks @alaypatel07 - ok so we have to keep this in mind when bumping the version of the perf lane. 

Or would there be a way of updating this not to rely on the job name? match to something like `periodic-kubevirt-e2e-k8s-*-sig-performance` ?

PR #3037, opened 2023-10-27
Fix typo in cluster name in 0.49 sig-compute presubmit

brianmcarey:
This causes the job to remain in a triggered state - never getting run. 

/cc @dhiller


dhiller (approved):
/approve

PR #3038, opened 2023-10-27
Update bridge-marker release presubmits to use new workloads cluster

brianmcarey:
These jobs were missed in the initial move - these jobs currently remain in triggered state so need to be updated to move the new cluster

/cc @0xFelix @dhiller


dhiller (approved):
/approve

PR #3056, opened 2023-11-07
CHANGELOG: Store in repo instead of git tag

andreabolognani:
Previously discussed [here](https://groups.google.com/g/kubevirt-dev/c/FvtVEAIVjo0/m/2Ekis79eAQAJ) and [here](https://groups.google.com/g/kubevirt-dev/c/lxqXb2wIqUk/m/yelXQYTUBAAJ).

Marking as draft because I haven't actually tested any of this O:-)

Also, before it can be merged, someone should populate the `CHANGELOG/` directory with release notes for past versions. I can take care of that if the approach proposed here is considered acceptable.

### DIFF
--- a/releng/release-tool/release-tool.go
+++ b/releng/release-tool/release-tool.go
@@ -37,13 +37,14 @@ type releaseData struct {
 	name      string
 	newBranch string
 
-	tagBranch        string
-	tag              string
-	promoteRC        string
-	promoteRCTime    time.Time
-	previousTag      string
-	releaseNotesFile string
-	skipReleaseNotes bool
+	tagBranch          string
+	tag                string
+	promoteRC          string
+	promoteRCTime      time.Time
+	previousTag        string
+	releaseNotesFile   string
+	releaseNotesReadme string
+	skipReleaseNotes   bool
 
 	force bool
 
@@ -92,6 +93,7 @@ func (r *releaseData) generateReleaseNotes() error {
 
 	releaseNotesDir := path.Join(r.repoDir, "CHANGELOG")
 	r.releaseNotesFile = path.Join(releaseNotesDir, fmt.Sprintf("CHANGELOG-%s.md", r.tag))
+	r.releaseNotesReadme = path.Join(releaseNotesDir, "README.md")
 
 	err := os.Mkdir(releaseNotesDir, 0755)
 	if err != nil && !os.IsExist(err) {
@@ -212,6 +214,19 @@ func (r *releaseData) generateReleaseNotes() error {
 	f.WriteString("```\n")
 
 	f.WriteString(additionalResources)
+
+	g, err := os.Create(r.releaseNotesReadme)
+	if err != nil {
+		return err
+	}
+	defer g.Close()
+
+	g.WriteString(fmt.Sprintf("# %s release notes\n", r.name))
+	g.WriteString("\n")
+	g.WriteString(fmt.Sprintf("The most recent release is [%s](CHANGELOG-%s.md).\n", r.tag, r.tag))
+
+	// TODO list past releases too
+
 	return nil
 }
 
@@ -329,6 +344,11 @@ func (r *releaseData) makeTag(branch string) error {
 	}
 
 	_, err = gitCommand("-C", r.repoDir, "add", r.releaseNotesFile)
+	if err != nil {
+		return err
+	}
+
+	_, err = gitCommand("-C", r.repoDir, "add", r.releaseNotesReadme)
 	if err != nil {
 		return err
 	}
--- a/releng/release-tool/release-tool.go
+++ b/releng/release-tool/release-tool.go
@@ -79,18 +79,13 @@ func _gitCommand(arg ...string) (string, error) {
 
 func (r *releaseData) generateReleaseNotes() error {
 	additionalResources := fmt.Sprintf(`
-Additional Resources
---------------------
-- Mailing list: <https://groups.google.com/forum/#!forum/kubevirt-dev>
-- Slack: <https://kubernetes.slack.com/messages/virtualization>
-- An easy to use demo: <https://github.com/%s/demo>
-- [How to contribute][contributing]
-- [License][license]
+## Additional Resources
 
-
-[contributing]: https://github.com/%s/%s/blob/main/CONTRIBUTING.md
-[license]: https://github.com/%s/%s/blob/main/LICENSE
----
+- Mailing list: https://groups.google.com/forum/#!forum/kubevirt-dev
+- Slack: https://kubernetes.slack.com/messages/virtualization
+- An easy to use demo: https://github.com/%s/demo
+- How to contribute: https://github.com/%s/%s/blob/main/CONTRIBUTING.md
+- License: https://github.com/%s/%s/blob/main/LICENSE
 `, r.org, r.org, r.repo, r.org, r.repo)
 
 	tagUrl := fmt.Sprintf("https://github.com/%s/%s/releases/tag/%s", r.org, r.repo, r.tag)
@@ -180,21 +175,24 @@ Additional Resources
 	numContributors := len(contributorList)
 	typeOfChanges = strings.TrimSpace(typeOfChanges)
 
-	f.WriteString(fmt.Sprintf("This release follows %s and consists of %d changes, contributed by %d people, leading to %s.\n", r.previousTag, numChanges, numContributors, typeOfChanges))
+	f.WriteString(fmt.Sprintf("# %s %s\n", r.name, r.tag))
+	f.WriteString("\n")
+	f.WriteString(fmt.Sprintf("This release follows %s and consists of %d changes, leading to %s.", r.previousTag, numChanges, typeOfChanges))
 	if r.promoteRC != "" {
-		f.WriteString(fmt.Sprintf("%s is a promotion of release candidate %s which was originally published %s", r.tag, r.promoteRC, r.promoteRCTime.Format("2006-01-02")))
+		f.WriteString("  \n")
+		f.WriteString(fmt.Sprintf("%s is a promotion of release candidate %s, which was originally published on %s.", r.tag, r.promoteRC, r.promoteRCTime.Format("2006-01-02")))
 	}
 	f.WriteString("\n")
-	f.WriteString(fmt.Sprintf("The source code and selected binaries are available for download at: %s.\n", tagUrl))
 	f.WriteString("\n")
-	f.WriteString(fmt.Sprintf("The primary release artifact of %s is the git tree. The release tag is\n", r.name))
+	f.WriteString(fmt.Sprintf("The primary release artifact of %s is the git tree. The release tag is ", r.name))
 	f.WriteString(fmt.Sprintf("signed and can be verified using `git tag -v %s`.\n", r.tag))
 	f.WriteString("\n")
-	f.WriteString(fmt.Sprintf("Pre-built containers are published on Quay and can be viewed at: <https://quay.io/%s/>.\n", r.org))
+	f.WriteString(fmt.Sprintf("The source code and selected binaries are available for download at: %s  \n", tagUrl))
+	f.WriteString(fmt.Sprintf("Pre-built containers are published on Quay and can be viewed at: https://quay.io/%s/\n", r.org))
 	f.WriteString("\n")
 
 	if len(releaseNotes) > 0 {
-		f.WriteString("Notable changes\n---------------\n")
+		f.WriteString("## Notable changes\n")
 		f.WriteString("\n")
 		for _, note := range releaseNotes {
 			f.WriteString(fmt.Sprintf("- %s\n", note))
@@ -202,12 +200,16 @@ Additional Resources
 	}
 
 	f.WriteString("\n")
-	f.WriteString("Contributors\n------------\n")
-	f.WriteString(fmt.Sprintf("%d people contributed to this release:\n\n", numContributors))
+	f.WriteString("## Contributors\n")
+	f.WriteString("\n")
+	f.WriteString(fmt.Sprintf("%d people contributed to this release:\n", numContributors))
+	f.WriteString("\n")
 
+	f.WriteString("```\n")
 	for _, contributor := range contributorList {
 		f.WriteString(fmt.Sprintf("%s\n", strings.TrimSpace(contributor)))
 	}
+	f.WriteString("```\n")
 
 	f.WriteString(additionalResources)
 	return nil
@@ -408,7 +410,7 @@ func (r *releaseData) getReleaseNote(number int) (string, error) {
 				note = strings.TrimPrefix(note, "-")
 				// best effort at catching "none" if the label didn't catch it
 				if note != "" && !strings.Contains(note, "NONE") && strings.ToLower(note) != "none" && strings.ToLower(note) != "n/a" {
-					note = fmt.Sprintf("[PR #%d][%s] %s", number, *pr.User.Login, note)
+					note = fmt.Sprintf("[#%d](https://github.com/%s/%s/pull/%d) ([@%s](https://github.com/%s)) %s", number, r.org, r.repo, number, *pr.User.Login, *pr.User.Login, note)
 					return note, nil
 				}
 			}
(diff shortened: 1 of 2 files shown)


releng/release-tool/release-tool.go
```diff
@@ -308,7 +308,21 @@ func (r *releaseData) makeTag(branch string) error {
 
 	r.generateReleaseNotes()
 
-	_, err := gitCommand("-C", r.repoDir, "tag", "-s", r.tag, "-F", r.releaseNotesFile)
+	if !r.dryRun {
```
0xFelix, line 311: I think `dryRun` is not needed here, this is only modifying a local copy of the repo. Only the push below should be guarded by `dryRun`.

releng/release-tool/release-tool.go
```diff
@@ -308,7 +308,21 @@ func (r *releaseData) makeTag(branch string) error {
...
+		_, err := gitCommand("-C", r.repoDir, "add", r.releaseNotesFile)
+		if err != nil {
+			return err
+		}
+	}
+
+	if !r.dryRun {
```
0xFelix, line 318: Same.

releng/release-tool/release-tool.go
```diff
@@ -308,7 +308,21 @@ func (r *releaseData) makeTag(branch string) error {
...
+		_, err := gitCommand("-C", r.repoDir, "commit", "-m", fmt.Sprintf("CHANGELOG: Add release notes for %s", r.tag))
+		if err != nil {
+			return err
+		}
+	}
+
+	_, err := gitCommand("-C", r.repoDir, "tag", "-s", r.tag)
```
0xFelix, line 325: `err` is unchecked?

releng/release-tool/release-tool.go
andreabolognani: Makes sense.

releng/release-tool/release-tool.go
```diff
@@ -308,7 +308,21 @@ func (r *releaseData) makeTag(branch string) error {
...
-	_, err := gitCommand("-C", r.repoDir, "tag", "-s", r.tag, "-F", r.releaseNotesFile)
+	if !r.dryRun {
+		_, err := gitCommand("-C", r.repoDir, "add", r.releaseNotesFile)
+		if err != nil {
+			return err
+		}
+	}
+
+	if !r.dryRun {
+		_, err := gitCommand("-C", r.repoDir, "commit", "-m", fmt.Sprintf("CHANGELOG: Add release notes for %s", r.tag))
```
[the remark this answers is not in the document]
andreabolognani: Yeah, looking at other uses of `git commit` in the tool it seems like a good idea to include it.

releng/release-tool/release-tool.go
```diff
@@ -308,7 +308,21 @@ func (r *releaseData) makeTag(branch string) error {
...
+		if err != nil {
```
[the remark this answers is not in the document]
andreabolognani: @0xFelix I'm not sure what you mean, can you please expand on that?

releng/release-tool/release-tool.go
andreabolognani: Pre-existing, but I agree that it's not idea. I'll add the check.

releng/release-tool/release-tool.go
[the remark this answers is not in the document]
0xFelix: It's possible to write it like this:

```
if _, err := gitCommand("-C", r.repoDir, "commit", "-m", fmt.Sprintf("CHANGELOG: Add release notes for %s", r.tag)); err != nil {
    return err
}
```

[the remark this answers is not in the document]
andreabolognani: I see! That style doesn't seem to be used much in the file though: only two checks are inline. I can still change it if you prefer, but personally I'd rather stay consistent with the existing coding style.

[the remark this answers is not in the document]
0xFelix: I'm fine with it, just a nit.

0xFelix:
Looks good from what I can tell, how can we test it?

andreabolognani:
@acardace was the last one to tag a release, so perhaps they can take a look and see if anything looks out of place? If things looks reasonable, I'll start working on the KubeVirt PR that will put all historical release notes into `CHANGELOG/` and mark this one as ready.

andreabolognani:
On second thought. While the release notes produced by the script are Markdown-adjacent, they're not actually valid Markdown that renders sensibly on GitHub. Let me try to improve that situation first.

It's still okay to take a look at the git part of the flow though, that's probably not going to change much.

andreabolognani:
I'm also pretty sure I broke all the tests O:-)

andreabolognani:
Big update.

I've fixed the test suite, and also tested things locally to ensure that the process still works - it does, at least as far as I can tell.

The output format is significantly overhauled with the aim of looking good in GitHub.

This is how the recent v1.1.0 release would have looked like in the new format:

* [the release notes (`CHANGELOG-v1.1.0.md`)](https://github.com/andreabolognani/kubevirt/blob/aeb201e6da611d65511ae3c001635ccd6cd9940a/CHANGELOG/CHANGELOG-v1.1.0.md)
* [browsing the `CHANGELOG/` directory to show off `README.md`](https://github.com/andreabolognani/kubevirt/tree/aeb201e6da611d65511ae3c001635ccd6cd9940a/CHANGELOG)
* [the commit adding those files](https://github.com/andreabolognani/kubevirt/commit/aeb201e6da611d65511ae3c001635ccd6cd9940a)

andreabolognani:
Added a couple of fixes for existing incorrect behaviors.

Made the code that parses the output of `git shortlog` more robust, since I've seen at least a couple of variations in the wild. The `CHANGELOG/` directory is now created if it doesn't exist, which is good both to remove the dependency on https://github.com/kubevirt/kubevirt/pull/10716 and for projects that might want to start using the tool in the future.

I don't think there's any outstanding concern, so I'm going to mark this as ready now.

How to validate the changes: run

```
$ make build && ./release-tool --org kubevirt --repo kubevirt --git-user 'Your Name' --git-email your@email --github-token-file /your/token --new-release v1.1.1
```

then check the `/tmp/release-tool/kubevirt/https-kubevirt` directory.

andreabolognani:
I've decided that I like the "embedded git output as preformatted text block" format better for the contributor list. It's more compact and easier to scan at a glance. I'm open to changing it back if people feel strongly otherwise.

releng/release-tool/release-tool.go
```diff
@@ -76,23 +78,26 @@ func _gitCommand(arg ...string) (string, error) {
...
+- Slack: https://kubernetes.slack.com/messages/virtualization
+- An easy to use demo: https://github.com/%s/demo
+- How to contribute: https://github.com/%s/%s/blob/main/CONTRIBUTING.md
+- License: https://github.com/%s/%s/blob/main/LICENSE
 `, r.org, r.org, r.repo, r.org, r.repo)
 
 	tagUrl := fmt.Sprintf("https://github.com/%s/%s/releases/tag/%s", r.org, r.repo, r.tag)
 
-	r.releaseNotesFile = fmt.Sprintf("%s/%s-release-notes.txt", r.repoDir, r.tag)
+	releaseNotesDir := fmt.Sprintf("%s/CHANGELOG", r.repoDir)
```
0xFelix, line 93: What about using `path.Join` instead of `fmt.Sprintf`?

releng/release-tool/release-tool.go
```diff
@@ -161,40 +176,56 @@ func (r *releaseData) generateReleaseNotes() error {
...
+	if err != nil {
+		return err
+	}
+	defer g.Close()
+
+	g.WriteString(fmt.Sprintf("# %s release notes\n", r.name))
+	g.WriteString("\n")
+	g.WriteString(fmt.Sprintf("The most recent release is [%s](CHANGELOG-%s.md).\n", r.tag, r.tag))
+
+	// TODO list past releases too
```
0xFelix, line 227: Instead of all past releases maybe list only the supported ones (last three IIRC)?

0xFelix (commented):
I like it! Added two comments.

releng/release-tool/release-tool.go
andreabolognani: I've followed what was already there but your suggestion makes sense and, since I'm already touching the code, I might as well improve upon it :)

releng/release-tool/release-tool.go
andreabolognani: k8s lists [all releases](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/README.md), which is nice considering that GitHub uses alphabetical sorting when listing files and so you get 1.10 listed before 1.2, which itself is listed before 1.20.

Either way, I'm not going to tackle that part right now so deciding how many releases to list is left as an exercise for the reader ;)

releng/release-tool/release-tool.go
andreabolognani: Done.

releng/release-tool/release-tool.go
0xFelix: ACK

0xFelix (approved):
/lgtm

Thanks :)

andreabolognani:
Ping.

Is anything missing? I'd like to get this and https://github.com/kubevirt/kubevirt/pull/10716 merged.

0xFelix:
Ping @dhiller @xpivarc @acardace 

What do you think?

dhiller (commented):
/approve

@xpivarc @acardace WDYT?

/hold

xpivarc:
I will bring this up with maintainers. Please hold for now.

aburdenthehand:
Looks really great. Thanks @andreabolognani for pushing forward on this. 

One question: Will this only be generated at the same time when we release a version? Currently we get the pre-release-tagged log ahead of the release which can be modified to categorise the release notes and also used for the user-guide release notes PR. 
(I'm happy as long as there is a way to generate this info ahead of the final release).

andreabolognani:
> One question: Will this only be generated at the same time when we release a version? Currently we get the pre-release-tagged log ahead of the release which can be modified to categorise the release notes and also used for the user-guide release notes PR.
> (I'm happy as long as there is a way to generate this info ahead of the final release).

The process shouldn't change much: right now the release notes are embedded in the tag the moment it's created, after my changes they will be added to the repository in a commit that is created right before the tag.

AFAICT currently the only way to generate the release notes ahead of the actual release is to use the `--dry-run` option, which will cause the tool to skip the `git push` step, and then inspect the temporary repository. That will still work. Or is there another way that I've missed?

I've now looked at the user-guide repository and I've seen that there is some automated processing going on there. That part would obviously have to be adapted. And the task of categorizing entries as changes, bug fixes etc., that's manual, right? If so, I'm not sure where this step would fit into the existing, automated release workflow.

Maybe we could start asking people to explicitly categorize their PRs at creation time? Something like

    ```release-note:deprecation
    Feature foo has been deprecated.
    ```

In any case, I wonder if it still makes sense to have the the release notes as part of user-guide once they're browsable via GitHub. Maybe we can just yank that section?

Sort of unrelated, but I've just realized that the way I've implemented things doesn't really work too well when it comes to tagging alphas, betas and rcs. I need to go back and revisit that part. I'll move the PR back to draft for the moment.

aburdenthehand:
> AFAICT currently the only way to generate the release notes ahead of the actual release is to use the --dry-run option, which will cause the tool to skip the git push step, and then inspect the temporary repository. That will still work. Or is there another way that I've missed?

I'm not sure what the mechanism is but when we prepare the RC I have been able to edit the tag text through github, which has given me two weeks to manually organise.
But running a dry-run locally for me works for now. As I understand it, it wouldn't change the actual release changelog that this script will prepare but means I can prepare the user-guide. 

Re: Why also publish the user guide - Last year when I was wondering the same thing I ran a bunch of findability tests comparing the two and the search is much better for the user guide than in github. I think it's also useful when searching a term in the user guide to have a related release note returned in the search results. 

> I've now looked at the user-guide repository and I've seen that there is some automated processing going on there. That part would obviously have to be adapted. And the task of categorizing entries as changes, bug fixes etc., that's manual, right? If so, I'm not sure where this step would fit into the existing, automated release workflow.

> Maybe we could start asking people to explicitly categorize their PRs at creation time? Something like
We have github labels already defined to handle all the categories (sig: and kind:). I imagine using something like 'release-note:deprecation' would be useful for kubevirt-bot to automatically add a deprecation label but tbh it might just be better to encourage greater label awareness/usage. Especially with the upcoming changes to SIG groups that has been talked about this year. 

With ~consistent labels, I imagine we can add a filtering mechanism to the great work you've done in this PR to automatically handle the categorisation of the release notes into their relevant sections. That would be a dream.

andreabolognani:
> I'm not sure what the mechanism is but when we prepare the RC I have been able to edit the tag text through github

Tags can't be edited directly through GitHub, much as commits can't. So I assume you're talking about the release notes **attached** to the tag, e.g. https://github.com/kubevirt/kubevirt/releases/tag/v1.1.0. If you compare that to the output of `git show v1.1.0`, you'll see that the latter is missing all the nice categorization.

> which has given me two weeks to manually organise.

So is it you taking care of sorting all changes into the appropriate categories for every release? Can you please describe the workflow in some detail? 

> But running a dry-run locally for me works for now. As I understand it, it wouldn't change the actual release changelog that this script will prepare but means I can prepare the user-guide.

IIUC you base your work on the release notes generated for RC, right? If so, you wouldn't even need do do any dry-runs: just update your local clone once the RC has been pushed and grab the release notes from the `CHANGELOG/` directory.

> Re: Why also publish the user guide - Last year when I was wondering the same thing I ran a bunch of findability tests comparing the two and the search is much better for the user guide than in github. I think it's also useful when searching a term in the user guide to have a related release note returned in the search results.

I have no strong objection to publishing the same information in multiple places, but I would like it to be consistent. If for whatever reason the release notes can't be the same everywhere, the version in the repo should IMO be the "best" one, as it's the only one that's cryptographically signed (via git tags) and guaranteed to survive events such as changes in hosting.

> We have github labels already defined to handle all the categories (sig: and kind:). I imagine using something like 'release-note:deprecation' would be useful for kubevirt-bot to automatically add a deprecation label but tbh it might just be better to encourage greater label awareness/usage. Especially with the upcoming changes to SIG groups that has been talked about this year.
> 
> With ~consistent labels, I imagine we can add a filtering mechanism to the great work you've done in this PR to automatically handle the categorisation of the release notes into their relevant sections. That would be a dream.

I really don't think I can justify spending a lot more time branching further into KubeVirt release tooling than I already have :) but since the labels matching each of the release notes categories appear to already exist, I think it would absolutely make a lot of sense to enforce via the usual PR tooling that each PR that contains release notes also has an appropriate label associated to it before being merged.

Once that guarantee is in place, it should be fairly simple to update the tool further so that items are automatically sorted into the various buckets.

aburdenthehand:
> Tags can't be edited directly through GitHub, much as commits can't. So I assume you're talking about the release notes **attached** to the tag, e.g. https://github.com/kubevirt/kubevirt/releases/tag/v1.1.0. If you compare that to the output of `git show v1.1.0`, you'll see that the latter is missing all the nice categorization.

Yes, sorry, that's what I'm referring to. 

> So is it you taking care of sorting all changes into the appropriate categories for every release? Can you please describe the workflow in some detail?

Once the tag is created I can edit the markdown and go through the release notes one by one and sort them (usually locally so I can work on a copy and compare against an original). Those with the kind/, sig/ or bug/ labels are easy; for the rest, some are easy to deduce and the remaining I reach out to a higher authority. 

> IIUC you base your work on the release notes generated for RC, right? If so, you wouldn't even need do do any dry-runs: just update your local clone once the RC has been pushed and grab the release notes from the `CHANGELOG/` directory.

Even better :)

> I have no strong objection to publishing the same information in multiple places, but I would like it to be consistent. If for whatever reason the release notes can't be the same everywhere, the version in the repo should IMO be the "best" one, as it's the only one that's cryptographically signed (via git tags) and guaranteed to survive events such as changes in hosting.

+1

> I really don't think I can justify spending a lot more time branching further into KubeVirt release tooling than I already have :) but since the labels matching each of the release notes categories appear to already exist, I think it would absolutely make a lot of sense to enforce via the usual PR tooling that each PR that contains release notes also has an appropriate label associated to it before being merged.

Oh yes, absolutely. I didn't mean to imply that the filtering should be included in this PR. I was just describing an existing alternative to having a release note type in the field itself.
I very much appreciate the work you've already put into this and don't think it should be blocked by any additional ideas. As you say, these can be added as an additional update.

PR #3068, opened 2023-11-15
Remove kind and molecule from prow-deploy image

brianmcarey:
kind and molecule are no longer required in the prow-deploy image as the presubmit has been moved to use kubevirtci[1] for the test cluster

Also includes bump to base images.

[1] https://github.com/kubevirt/project-infra/pull/3061

/cc @dhiller @xpivarc


dhiller (approved):
/approve

dhiller:
@brianmcarey since the images make an impact, we might add a rehearsal of affected jobs with a temporary image build in order to check whether they would still run with it. Would require additional features on the rehearse plugin though and might be hard...
WDYT?

PR #3106, opened 2023-12-04
When releasing CDI, cross compile arm64 as well

awels:
We have been releasing periodic arm64 containers for some time now. We should make an official release as well.

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/containerized-data-importer/containerized-data-importer-postsubmits.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/containerized-data-importer/containerized-data-importer-postsubmits.yaml
@@ -29,9 +29,10 @@ postsubmits:
         - "-c"
         - |
           cat "$QUAY_PASSWORD" | docker login --username $(cat "$QUAY_USER") --password-stdin=true quay.io &&
-          # Only push images on tags
+          # Only push images on tags, for amd64 and aarch64 (crossbuild, since we have no aarch64 builder)
           [ -z "$(git tag --points-at HEAD | head -1)" ] ||
-          DOCKER_TAG="$(git tag --points-at HEAD | head -1)" make bazel-push-images
+          DOCKER_TAG="$(git tag --points-at HEAD | head -1)-amd64" make bazel-push-images ||
+          DOCKER_TAG="$(git tag --points-at HEAD | head -1)-arm64" BUILD_ARCH="crossbuild-aarch64" make bazel-push-images
         # docker-in-docker needs privileged mode
         securityContext:
           privileged: true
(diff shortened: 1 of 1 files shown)


brianmcarey (commented):
It might be better to move this job to the workloads cluster `kubevirt-prow-workloads` and increase the timeout - I think the cross compile will add a lot of run time and require more performance.

awels:
This is doing essentially what is happening in the periodic and that is running in the `kubevirt-prow-workload` cluster, but the timeout is the same, I can increase it because I am running 2 builds.

nixpanic:
You probably want to correct the subject of the PR, `s/amd64/arm64/`. Increasing the timeout for cross-compiling is probably a good idea indeed. I'd :+1:  the changes, but it won't count for getting this merged.

github/ci/prow-deploy/files/jobs/kubevirt/containerized-data-importer/containerized-data-importer-postsubmits.yaml
```diff
@@ -29,9 +31,10 @@ postsubmits:
         - "-c"
         - |
           cat "$QUAY_PASSWORD" | docker login --username $(cat "$QUAY_USER") --password-stdin=true quay.io &&
-          # Only push images on tags
+          # Only push images on tags, for amd64 and aarch64 (crossbuild, since we have no aarch64 builder)
           [ -z "$(git tag --points-at HEAD | head -1)" ] ||
-          DOCKER_TAG="$(git tag --points-at HEAD | head -1)" make bazel-push-images
+          DOCKER_TAG="$(git tag --points-at HEAD | head -1)-amd64" make bazel-push-images ||
```
akalenyu, line 36: Reassigning `DOCKER_TAG`? Is this what you meant?

awels: yes if you look at what kubevirt does here, they do the same, making a tag for amd and arm and quay seems to correctly re-arrange that.

github/ci/prow-deploy/files/jobs/kubevirt/containerized-data-importer/containerized-data-importer-postsubmits.yaml
```diff
@@ -29,9 +31,10 @@ postsubmits:
...
+          DOCKER_TAG="$(git tag --points-at HEAD | head -1)-arm64" BUILD_ARCH="crossbuild-aarch64" make bazel-push-images
```
brianmcarey, line 37: Would you also want to create a combined manifest? 

something like:
```
podman manifest create --amend $IMAGE_NAME $IMAGE_NAME-amd64 $IMAGE_NAME-arm64
podman manifest push $IMAGE_NAME
```

awels:
/hold
I will likely close this in favor of a script in the CDI repo.

github/ci/prow-deploy/files/jobs/kubevirt/containerized-data-importer/containerized-data-importer-postsubmits.yaml
brianmcarey: Sounds good to me

PR #3121, opened 2023-12-11
Update infra components doc to capture recent changes

brianmcarey:
The cluster contexts have been changed following cluster migrations - this should be updated in the doc

Also added the bot-review plugin to list of external plugins. 

/cc @dhiller @xpivarc


dhiller (approved):
/approve

PR #3126, opened 2023-12-13
release-tool: Update phase 2 jobs to be considered phase 1 when forking a new branch

oshoval:
During new release branch creation, we should convert phase 2 jobs
to be considered phase 1, by setting `always_run=true`,
because phased plugin is used only for main branch.

Note: we might want to consider https://github.com/kubevirt/project-infra/pull/3184 (by removing `run_before_merge` as well)
but since the jobs become phase 1 it shouldn't affect even if the field remains.

### DIFF
--- a/releng/release-tool/release-tool.go
+++ b/releng/release-tool/release-tool.go
@@ -14,10 +14,10 @@ import (
 	"strings"
 	"time"
 
-	"golang.org/x/oauth2"
-
 	"github.com/Masterminds/semver"
 	"github.com/google/go-github/v32/github"
+	"golang.org/x/oauth2"
+	"sigs.k8s.io/yaml"
 )
 
 type blockerListCacheEntry struct {
@@ -414,8 +414,14 @@ func (r *releaseData) forkProwJobs() error {
 
 	// create new prow configs if they don't already exist
 	if _, err := os.Stat(fullOutputConfig); err != nil && os.IsNotExist(err) {
+		updateJobConfig, err := updatePhase2Jobs(r.org+"/"+r.repo, fullJobConfig)
+		if err != nil {
+			return err
+		}
+		defer os.Remove(updateJobConfig)
+
 		log.Printf("Creating new prow yaml at path %s", fullOutputConfig)
-		cmd := exec.Command("/usr/bin/config-forker", "--job-config", fullJobConfig, "--version", version, "--output", fullOutputConfig)
+		cmd := exec.Command("/usr/bin/config-forker", "--job-config", updateJobConfig, "--version", version, "--output", fullOutputConfig)
 		bytes, err := cmd.CombinedOutput()
 		if err != nil {
 			log.Printf("ERROR: config-forker command output: %s : %s ", string(bytes), err)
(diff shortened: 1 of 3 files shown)


oshoval:
This is not good, the output is full of changes that are not needed (and also need to load a real prow config)

releng/release-tool/release-tool.go
```diff
@@ -1271,3 +1277,101 @@ func main() {
...
+			return "", err
+		}
+	}
+
+	updatedContent, err := yaml.Marshal(data)
+	if err != nil {
+		return "", fmt.Errorf("Error marshalling yaml: %v", err)
+	}
+
+	updateJobConfig := os.TempDir() + "/job-config.yaml"
```
oshoval, line 1305: should we use rand name ? if we need to allow reentrancy

releng/release-tool/release-tool.go
```diff
@@ -1271,3 +1277,105 @@ func main() {
...
+		optional, optionalOk := jobMap["optional"].(bool)
+		alwaysRun, alwaysRunOk := jobMap["always_run"].(bool)
+		runIfChanged, runIfChangedOk := jobMap["run_if_changed"].(string)
+		skipIfOnlyChanged, skipIfOnlyChangedOk := jobMap["skip_if_only_changed"].(string)
+
+		if (!alwaysRunOk || !alwaysRun) &&
+			(!optionalOk || !optional) &&
+			(!runIfChangedOk || runIfChanged == "") &&
+			(!skipIfOnlyChangedOk || skipIfOnlyChanged == "") {
+			jobs = append(jobs, jobMap["name"].(string))
```
oshoval, line 1376: maybe we can update it during detection

releng/release-tool/release-tool.go
```diff
@@ -1271,3 +1277,105 @@ func main() {
...
+		if !ok {
+			return fmt.Errorf("Invalid job format in yaml")
+		}
+
+		name, ok := jobMap["name"].(string)
+		if !ok {
+			return fmt.Errorf("Missing or invalid 'name' field for job in yaml")
+		}
+
+		if name == jobName {
```
oshoval, line 1339: basically we can add the detection here
question is if we want to split detection and modification

releng/release-tool/release-tool.go
```diff
@@ -1271,3 +1277,105 @@ func main() {
...
+	if err != nil {
+		return "", err
+	}
+
+	for _, job := range jobs {
+		err := updateJob(orgRepo, job, data)
+		if err != nil {
+			return "", err
+		}
+	}
```
oshoval, line 1302: updateJob can get a list and do it all in one shot as well

oshoval:
/cc @brianmcarey @dhiller 
needed once we start working with phased plugin

releng/release-tool/release-tool.go
```diff
@@ -1271,3 +1277,70 @@ func main() {
 		}
 	}
 }
+
+func updatePhase2Jobs(orgRepo, fullJobConfig string) (string, error) {
+	fileContent, err := ioutil.ReadFile(fullJobConfig)
```
dhiller, line 1282: What is the reason why you don't use [config.ReadJobConfig](https://github.com/kubernetes/test-infra/blob/c5fbbe779eac4753c6c1a13dda1a89a07de65f32/prow/config/config.go#L1674) from kubernetes/test-infra? I think that could simplify the code, i.e. making unnecessary all the type conversion?

Usage example: https://github.com/kubevirt/project-infra/blob/a940953303f34b402f70d6b45c840fce519f75b7/robots/pkg/kubevirt/cmd/require/presubmits.go#L96

oshoval: i think that when i tried something alike it reordered the fields and did a mess
i can try to make sure with this function

dhiller: Yeah, that's the func I meant. I don't know if it matters much that it reorders the fields, since this is from perspective of the forked jobs, which aren't yet committed anyway, no?

oshoval: Forgot how i tested it on the past, and what was the results tbh

Tried now to trigger only the func from the [diff](https://github.com/kubevirt/project-infra/compare/c42e6a20402114f19a49a528539ff35c1359e6b8..ac8a5d6a7e206ff8c35ce4e03c7b0cf4a75e8980#diff-585313f6146df52ade10307f2f7bd15a57cddd496439c6cc2d8c1cdd30eb7562L1290)

```
	fullJobConfig := fmt.Sprintf("%s/github/ci/prow-deploy/files/jobs/%s/%s/%s-presubmits.yaml",
		"/home/oshoval/go/src/github.com/kubevirt/project-infra", r.org, r.repo, r.repo)
	updateJobConfig, err := updatePhase2Jobs(r.org, r.repo, fullJobConfig)
	if err != nil {
		log.Fatalf("ERROR %s", err)
	}

	log.Printf("updateJobConfig: %s", updateJobConfig)
```

The load fails when it validates the presubmits
```
ERROR [invalid presubmit job pull-kubevirt-e2e-k8s-1.29-sig-performance: invalid decoration config: utility image config is not specified, invalid presubmit job pull-kubevirt-e2e-windows2016: invalid decoration config: utility image config is not specified, invalid presubmit job pull-kubevirt-e2e-kind-1.27-vgpu: invalid decoration config:
```
It calls ... validateDecoration which eventually fails about the decoration

I think the current way the PR suggests is easiest

oshoval: Don't remember how i tested it exactly, 
when i test it now it totally mess the file (did some tricks there to tests it quick and dirty)
IIRC, at the past it was either messing it or added tons of nil fields and this was the reason i dropped it
The current solution is the cleanest i found tbh in matter of the output and also the code is pretty straight forward and has unit tests.

oshoval:
@brianmcarey @dhiller 
wdyt about the following idea:
Close this PR and even without phased plugin supporting branches, it will still work and just run the phase 2
by tide due to "run_before_merge".
It is totally safe as long as we use "run_before_merge" flag.
If in the future we fix phased plugin such that this flag can be removed, then we will need to revisit this PR or some other solutions (like supporting phased plugin for branches etc)
Alternative is to continue working on the PR / take as it.

oshoval:
i am closing this PR, i think that everything will work atm out of the box, (no need to change nor the always run nor the run_before_merge)
please see previous comment explanation
if someone is against please let me know

Thanks

/close

oshoval:
reopened upon Brian request

brianmcarey:
> reopened upon Brian request It should solve new release creation imo it is good as is, hope you will agree

thanks @oshoval - this is an option for fixing - https://github.com/kubevirt/project-infra/issues/3478

dhiller:
@xpivarc this is the PR I talked about in today's sig-ci meeting

dhiller:
@oshoval I'd opt to close this one in favor of https://github.com/kubevirt/project-infra/pull/3958

WDYT?

oshoval:
Hi @dhiller 
sure, thanks a lot

one question to consider, wouldn't just updating tools-infra will solve at least the `run_before_merge` issue ?
(since the new pkg includes this field in the structs that are used to unmarshel)
unless it is already updated

dhiller:
yes, the uodate of test-infra deseriales the run_before_merge correctly, however, i'd like to have it as before. we always changed the run_before_merge to always_run.

dhiller:
maybe we can discuss that we only run the remaining lanes directly before merge, but i think that since we have a lot less backports than  main prs it doesn't matter much.

wdyt?

oshoval:
imho lets keep as simple as possible the branches please, i.e just `always_run` all on branches as you said iiuc

PR #3139, opened 2023-12-21
Remove prometheus-stack as it is no longer maintained or deployed

brianmcarey:
Since all of the infra clusters have been migrated to OCP and we are making us of the builtin monitoring stack - this prometheus stack is no longer deployed or maintained.

/cc @dhiller @xpivarc


dhiller (approved):
/approve

dhiller:
Thank you for cleaning up, Brian! :100:

PR #3149, opened 2024-01-03
Bump shared-images-controller dependencies

brianmcarey:
Update the base image for the controller to the latest bootstrap.

Also update go version to v1.21 and podman to version v4.9.0

Podman v4.9[1] included a large number of bug fixes that make it
worthwhile upgrading the podman version in the shared-images-controller

[1] https://github.com/containers/podman/releases/tag/v4.8.0

/cc @xpivarc @dhiller


dhiller (approved):
/approve

PR #3150, opened 2024-01-04
plugins: Introduce checklist plugin

oshoval:
This plugin adds a new label that blocks merging `do-not-merge/incomplete-checklist`.
It monitors PR body [checklist](https://github.com/kubevirt/kubevirt/blob/main/.github/PULL_REQUEST_TEMPLATE.md).
In case any of the items exists and unchecked it will add the label.
It will remove it once all items that do exists, are checked.
Monitors PR Open and PR Edit events.

Note: Currently active on all repos, we might want to limit it for specific repos,
but on the other hand, since those repos don't have the template, the label won't be added.


oshoval:
Will reopen once required

PR #3156, opened 2024-01-08
common-instancetypes: Use bump-common-instancetypes.sh

0xFelix:
Use bump-common-instancetypes.sh in the periodics that update kubevirt/kubevirt to the latest common-instancetypes version.

Also, add the 'periodic-' prefix to the job name and run the job on the
smaller control plane cluster instead.


brianmcarey (approved):
/approve

PR #3166, opened 2024-01-11
Add periodic job to create bump PR for centos version in kubevirtci

brianmcarey:
From time to time old centos stream 9 images are removed and cleaned up,
this can cause publishes of kubevirtci[1] and presubmits to fail.

This change adds a periodic job that runs once a week to create a PR in the kubevirtci repo with the latest
published centos stream version.

[1] https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/logs/publish-kubevirtci/1745004368974843904#1:build-log.txt%3A2263

/cc @dhiller @oshoval @xpivarc

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/kubevirtci/kubevirtci-periodics.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/kubevirtci/kubevirtci-periodics.yaml
@@ -95,3 +95,29 @@ periodics:
       resources:
         requests:
           memory: "200Mi"
+- name: periodic-kubevirtci-bump-centos-version
+  cron: "50 1 * * 0"
+  annotations:
+    testgrid-create-test-group: "false"
+  decorate: true
+  decoration_config:
+    timeout: 1h
+  max_concurrency: 1
+  extra_refs:
+  - org: kubevirt
+    repo: kubevirtci
+    base_ref: main
+    workdir: true
+  labels:
+    preset-docker-mirror: "true"
+    preset-github-credentials: "true"
+  cluster: kubevirt-prow-control-plane
+  spec:
+    containers:
+    - image: quay.io/kubevirtci/pr-creator:v20230103-9f4e101
+      command: ["/bin/sh", "-c"]
+      args:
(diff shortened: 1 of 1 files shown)


brianmcarey:
requires https://github.com/kubevirt/kubevirtci/pull/1114 to be merged first

/hold

github/ci/prow-deploy/files/jobs/kubevirt/kubevirtci/kubevirtci-periodics.yaml
```diff
@@ -95,3 +95,29 @@ periodics:
       resources:
         requests:
           memory: "200Mi"
+- name: periodic-kubevirtci-bump-centos-version
+  cron: "50 1 * * *"
```
oshoval, line 99: we dont need to run it every day,
it means there will be lots of providers that the only thing is changed is the centos
and we would take each and every change

i think once a month or two / or even just manually upon needs is better, as was until today

oshoval:
the reason for all this was just because images become deprecated every now and then
what about lets just mirror them ?

EdDev:
Please provide detailed reasoning in the PR description and commit.
I would also be interesting to understand the implications of this.

CentOS Stream is a development tracking OS, it is expected to change frequently with partially verified packages.
This may in turn cause instability in our e2e tests.

brianmcarey:
> Please provide detailed reasoning in the PR description and commit. I would also be interesting to understand the implications of this.

Updated the PR and commit with more details. 
> 
> CentOS Stream is a development tracking OS, it is expected to change frequently with partially verified packages. This may in turn cause instability in our e2e tests.

This updated version will be tested in two stages - first by the conformance tests in kubevirtci and then by all the kubevirt presubmit lanes - any instability should be identified at these stages. 

This is really no different from what we have been doing manually up to this point.

oshoval:
i still think 1 time a week is way too much,
we dont change kubevirtci lately a lot, and we don't need to rebase changes again and again,
causing potential flakes, images re download manually and by CI, jobs that run on CI

the image dies just once if few months, so we can do it once in few months either manual or automatic
but even if automatic, once in two months is more than enough imo

oshoval:
possible idea - in the periodic, add detection if old image got deleted (and new image exists, to show it is not just the whole host is down temporary)
and only in this case update the image

wdyt ?

PR #3170, opened 2024-01-12
prow-deploy: Remove deployment files that are specific to the old cluster

brianmcarey:
The old control plane cluster has been removed so these files are no longer relevant

/cc @dhiller


dhiller (approved):
/approve

PR #3184, opened 2024-01-17
phased: Enable run_before_merge for phase 2 jobs

oshoval:
Seems tide has a bug, it doesnt run the phase 2 for batches. 
A simple wa until it is fixed is to use `run_before_merge`, which force a batch to run it.

Real fix is to consider manually triggered jobs as well on this line at `presubmitsForBatch`
`shouldRun, err := ps.ShouldRun(baseBranch, c.changedFiles.batchChanges(prs), ps.RunBeforeMerge, false)`
The required logic exists on https://github.com/kubernetes/test-infra/pull/30191

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
@@ -801,6 +801,7 @@ presubmits:
         securityContext:
           privileged: true
   - always_run: false
+    run_before_merge: true
     annotations:
       fork-per-release: "true"
       testgrid-dashboards: kubevirt-presubmits


brianmcarey (approved):
This looks like what we were looking for. 

```
        // RunBeforeMerge indicates that a job should always run by Tide as long as
        // Brancher matches.
        // This is used when a prowjob is so expensive that it's not ideal to run on
        // every single push from all PRs.
        RunBeforeMerge bool `json:"run_before_merge,omitempty"`
```

https://github.com/kubernetes/test-infra/blob/c758e68cdeb67314148a6be992957d092d63fcaa/prow/config/jobs.go#L203C1-L205C57

The phased plugin will still be useful I think though as it will not have to wait for "phase 1" jobs to complete before triggering the "phase 2" jobs. 

/approve

oshoval:
Trying here to fix tide, WIP
https://github.com/kubernetes/test-infra/pull/31687

once it is merged (hopefully), we can remove the WA

PR #3187, opened 2024-01-18
kubevirt,presubmits: remove bump-kubevirtci and sev presubmits

brianmcarey:
The bump-kubevirtci presubmit was not regularly used as there is an existing periodic job. If a bump is required urgently, the CI team will be able to trigger the periodic manually. This job also does not make sense for release branches.

The SEV presubmit has not been used recently - lets rely on the periodic lane for feedback on SEV.

/cc @dhiller @xpivarc @enp0s3


dhiller (approved):
/approve

PR #3195, opened 2024-01-19
kubevirt,presubmits: increase memory requests for prow-control-plane jobs

brianmcarey:
When there is a large number of jobs kicked off at once, we can see that the control-plane cluster gets overwhelmed and jobs fail early to start the podman socket[1].

Screenshot from 2024-01-19 08-56-02

Doubling the memory request will reduce the number of running jobs by half and should help with this problem.

These jobs are generally very short so this should not really impact feedback time.

[1] https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/11041/pull-kubevirt-verify-go-mod/1748262393516920832

/cc @dhiller @xpivarc

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
@@ -322,7 +322,7 @@ presubmits:
         name: ""
         resources:
           requests:
-            memory: 4Gi
+            memory: 8Gi
         securityContext:
           privileged: true
   - always_run: false
@@ -353,7 +353,7 @@ presubmits:
         name: ""
         resources:
           requests:
-            memory: 4Gi
+            memory: 8Gi
         securityContext:
           privileged: true
   - always_run: true
@@ -382,7 +382,7 @@ presubmits:
         name: ""
         resources:
           requests:
-            memory: 4Gi
+            memory: 8Gi
         securityContext:
           privileged: true
   - always_run: false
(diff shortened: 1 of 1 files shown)


github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
```diff
@@ -353,7 +353,7 @@ presubmits:
         name: ""
         resources:
           requests:
-            memory: 4Gi
+            memory: 8Gi
```
dhiller, line 356: This is not run by default, so doubling it's memory request will have no effect.

github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
```diff
@@ -567,7 +567,7 @@ presubmits:
...
```
dhiller, line 570: nit: this is running unnested, so it shouldn't be required to double memory for it.

github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
```diff
@@ -793,7 +793,7 @@ presubmits:
...
```
dhiller, line 796: This is also unnested.

dhiller (commented):
Thanks for taking care of this! One question and two nits inline.

github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits.yaml
brianmcarey: I think it is ok at 8Gi - just so if the test lane is scheduled - it is not squeezed onto a node that already has a number of jobs running.

PR #3202, opened 2024-01-22
Update test-infra dependency to make use of new prowjob config fields

brianmcarey:
As part of the phased plugin work we are introducing a new field `run_before_merge` to some of our prowjobs - this is not available in the version of test-infra that is currently in place and as a result the automation removes this field[1] from our configs.

Update test-infra to the latest version in order to allow using this field.

[1] https://github.com/kubevirt/project-infra/pull/3198#discussion_r1461650975

/cc @oshoval @dhiller


oshoval:
tried to convert quickly git to git v2 here
unit tests pass (beside a test that isn't related that didnt pass also before)
https://github.com/kubevirt/project-infra/pull/3203

maybe worth to combine the PRs and see ?
we are already using v2 on most places, just botreview didn't so if this will be the only problem we should
be fine

brianmcarey:
> I figure the tests are failing since in later test-infra versions bazel has been removed.

Yeah it looks like we have further bazel issues here.

oshoval:
> > I figure the tests are failing since in later test-infra versions bazel has been removed.
> 
> Yeah it looks like we have further bazel issues here.

do we want to try and fix them ? 
or to do some workaround until tide is fixed and the new flag is not used? (started https://github.com/kubernetes/test-infra/pull/31687 but not sure how long it will take)
we can also try to add some ad hoc hack that revert just changes to this line, if the job is kept

oshoval:
a fix can be something like done here
https://github.com/kubevirt/project-infra/pull/3126
agnostic of the struct, in case the job wasn't remove, check the name of it, and take it vanilla using the logic that appears on mentioned PR, this way it won't pass via the unmarsheling that can remove fields iiuc

PR #3309, opened 2024-03-08
Trigger the performance presubmit on monitoring changes

rthallisey:
The perf jobs rely heavily on the monitoring pkg so trigger the performance presubmit job when there are changes in that pkg.

```release-note
NONE
```


brianmcarey (commented):
Looks good @rthallisey  - could you sign the commit?

brianmcarey (commented):
/approve

thanks @rthallisey !

PR #3356, opened 2024-04-19
Explicitly set REGISTRY env in ovs-cni jobs

ykulazhenkov:

Explicitly set REGISTRY env in ovs-cni jobs
This change is required because default registry in the Makefile in the ovs-cni repo was changed to ghcr.io/k8snetworkplumbingwg.

Context: https://github.com/k8snetworkplumbingwg/ovs-cni/pull/307
Example of a failed job https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/logs/main-ovs-cni/1781239529819082752
```release-note
NONE
```

@phoracek Do we want to continue push to quay in addition to GH registry? If yes, then I believe that this patch will fix the push jobs.

### DIFF
--- a/github/ci/prow-deploy/files/jobs/k8snetworkplumbingwg/ovs-cni/ovs-cni-postsubmits.yaml
+++ b/github/ci/prow-deploy/files/jobs/k8snetworkplumbingwg/ovs-cni/ovs-cni-postsubmits.yaml
@@ -18,8 +18,10 @@ postsubmits:
             command:
               - "/usr/local/bin/runner.sh"
               - "/bin/bash"
-              - "-c"
-              - "cat $QUAY_PASSWORD | docker login --username $(cat $QUAY_USER) --password-stdin=true quay.io && make docker-build docker-push"
+              - "-ce"
+              - |
+                cat $QUAY_PASSWORD | docker login --username $(cat $QUAY_USER) --password-stdin=true quay.io
+                make REGISTRY=quay.io/kubevirt docker-build docker-push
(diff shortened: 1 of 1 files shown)


github/ci/prow-deploy/files/jobs/k8snetworkplumbingwg/ovs-cni/ovs-cni-postsubmits.yaml
```diff
@@ -18,8 +18,10 @@ postsubmits:
             command:
               - "/usr/local/bin/runner.sh"
               - "/bin/bash"
-              - "-c"
-              - "cat $QUAY_PASSWORD | docker login --username $(cat $QUAY_USER) --password-stdin=true quay.io && make docker-build docker-push"
+              - "-ce"
+              - |
```
ykulazhenkov, line 22: cmd format is aligned with "release-ovs-cni" job

phoracek:
@ykulazhenkov I don't think there is a reason to keep pushing to the old registry. It would be probably better to remove this automation, so we have less stuff to maintain.

ykulazhenkov:
> @ykulazhenkov I don't think there is a reason to keep pushing to the old registry. It would be probably better to remove this automation, so we have less stuff to maintain.

Okay, thanks. Then, I will close this PR. I don't know the exact list of jobs that are no longer needed. Can you submit a PR to remove them?

phoracek:
@ykulazhenkov sure, fair enough, posted: https://github.com/kubevirt/project-infra/pull/3357

PR #3476, opened 2024-06-19
vm-image-builder: update vm-image-builder for s390x

vamsikrishna-siddu:
Updated the  vm-image-builder to support the building of s390x MultiArch VM images.


brianmcarey (approved):
/lgtm

/approve

thanks @vamsikrishna-siddu

PR #3566, opened 2024-08-01
Prow jobs for publishing k8s provider multi arch images

chandramerla:
This PR adds multi-arch publish jobs for centos9 and k8s provider images. 

As we've added s390x support in k8s provider (started with 1.30 slim), centos9 and gocli images as part of https://github.com/kubevirt/kubevirtci/pull/1252, this PR does changes required in project-infra repo to enable the prow jobs for the same.

This PR makes sense to be merged along with changes said above.

**Checklist**

This checklist is not enforcing, but it's a reminder of items that could be relevant to every PR.
Approvers are expected to review this list.

- [ ] Design: A [design document](https://github.com/kubevirt/community/tree/main/design-proposals) was considered and is present (link) or not required
- [x] PR: The PR description is expressive enough and will help future contributors
- [x] Code: [Write code that humans can understand](https://en.wikiquote.org/wiki/Martin_Fowler#code-for-humans) and [Keep it simple](https://en.wikipedia.org/wiki/KISS_principle)
- [ ] Refactor: You have [left the code cleaner than you found it (Boy Scout Rule)](https://learning.oreilly.com/library/view/97-things-every/9780596809515/ch08.html)
- [ ] Upgrade: Impact of this change on upgrade flows was considered and addressed if required
- [x] Testing: New code requires [new unit tests](https://github.com/kubevirt/kubevirt/blob/main/docs/reviewer-guide.md#when-is-a-pr-good-enough). New features and bug fixes require at least on e2e test
- [ ] Documentation: A [user-guide update](https://github.com/kubevirt/user-guide/) was considered and is present (link) or not required. You want a user-guide update if it's a user facing feature / API change.
- [ ] Community: Announcement to [kubevirt-dev](https://groups.google.com/g/kubevirt-dev/) was considered

**Release note**:
```release-note
NONE
```

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/kubevirtci/kubevirtci-postsubmits.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/kubevirtci/kubevirtci-postsubmits.yaml
@@ -44,11 +44,83 @@ postsubmits:
           - "/usr/local/bin/runner.sh"
           - "/bin/bash"
           - "-c"
-          - >
+          - |
             cat $QUAY_PASSWORD | podman login --username $(<$QUAY_USER) --password-stdin quay.io &&
+            SHORT_SHA=$(git rev-parse --short HEAD) &&
             ./publish.sh &&
-            echo "$(git tag --points-at HEAD | head -1)" > latest &&
-            gsutil cp ./latest gs://kubevirt-prow/release/kubevirt/kubevirtci/latest
+            # Gets git tag created by above publish.sh run
+            KUBEVIRTCI_TAG=$(git tag --points-at HEAD | head -1) &&
+            echo "$KUBEVIRTCI_TAG" > latest &&
+            gsutil cp ./latest gs://kubevirt-prow/release/kubevirt/kubevirtci/latest &&
+            echo "$KUBEVIRTCI_TAG" > amd64-$SHORT_SHA &&
+            gsutil cp ./amd64-$SHORT_SHA gs://kubevirt-prow/release/kubevirt/kubevirtci/amd64-$SHORT_SHA
+          # docker-in-docker needs privileged mode
+          env:
+          - name: GIMME_GO_VERSION
+            value: "1.22.5"
+          securityContext:
+            privileged: true
+          volumeMounts:
+          - mountPath: /dev
+            name: devices
+          resources:
+            requests:
+              memory: "29Gi"
+    - name: publish-kubevirtci-s390x
+      branches:
+      - main
+      always_run: true
+      annotations:
+        testgrid-create-test-group: "false"
+      decorate: true
+      decoration_config:
+        timeout: 2h
+      max_concurrency: 1
+      extra_refs:
+      - org: kubevirt
+        repo: project-infra
+        base_ref: main
+      labels:
+        preset-podman-in-container-enabled: "true"
+        preset-gcs-credentials: "true"
+        preset-kubevirtci-quay-credential: "true"
+      cluster: prow-s390x-workloads
+      spec:
+        volumes:
+        - hostPath:
+            path: /dev
+            type: Directory
+          name: devices
+        containers:
+        - image: quay.io/kubevirtci/golang:v20241213-57bd934
+          command:
+          - "/usr/local/bin/runner.sh"
+          - "/bin/bash"
+          - "-c"
+          - |
+            cat $QUAY_PASSWORD | podman login --username $(<$QUAY_USER) --password-stdin quay.io &&
+            SHORT_SHA=$(git rev-parse --short HEAD) &&
+            GCS_FILE_PATH=release/kubevirt/kubevirtci/amd64-$SHORT_SHA &&
+            source /usr/local/bin/gcs_restapi.sh &&
+            CHECK_INTERVAL=30 &&
+            while true; do
+                if stat_gcs_file kubevirt-prow "$GCS_FILE_PATH"; then
+                    echo "File $GCS_FILE_PATH is now available."
+                    break
+                else
+                    echo "File $GCS_FILE_PATH not found. Checking again in $CHECK_INTERVAL seconds."
+                    sleep $CHECK_INTERVAL
+                fi
+            done &&
+            KUBEVIRTCI_TAG=$(cat_gcs_file kubevirt-prow "$GCS_FILE_PATH")
+            if [ $? -ne 0 ]; then
+                echo "Failed to fetch KUBEVIRTCI_TAG"
+                exit 1
+            fi
+            export KUBEVIRTCI_TAG &&
+            echo "Fetched KUBEVIRTCI_TAG: $KUBEVIRTCI_TAG" &&
+            ./publish.sh &&
+            rm_gcs_file kubevirt-prow "$GCS_FILE_PATH"
           # docker-in-docker needs privileged mode
           env:
           - name: GIMME_GO_VERSION


github/ci/prow-deploy/files/jobs/kubevirt/kubevirtci/kubevirtci-periodics.yaml
```diff
@@ -126,6 +182,21 @@ periodics:
       - "/bin/bash"
       - "-c"
       - >
+        # For getting centos image tag from s390x prow job and use same for x86 and manifest-list images
```
brianmcarey, line 185: I would prefer if the s390x job could relay on the tag from the x86 image rather than relying on the s390x job here.

github/ci/prow-deploy/files/jobs/kubevirt/kubevirtci/kubevirtci-postsubmits.yaml
```diff
@@ -46,6 +97,19 @@ postsubmits:
           - "-c"
           - >
             cat $QUAY_PASSWORD | podman login --username $(<$QUAY_USER) --password-stdin quay.io &&
+            SHORT_SHA=$(git rev-parse --short HEAD) &&
+            GCS_FILE_PATH=gs://kubevirt-prow/release/kubevirt/kubevirtci/s390x-$SHORT_SHA &&
```
brianmcarey, line 101: I would prefer if the s390x job could relay on the tag from the x86 image rather than relying on the s390x job here.

images/pr-creator/Dockerfile
```diff
@@ -1,6 +1,6 @@
 FROM quay.io/kubevirtci/golang:v20240229-74549d5
 
-ENV GIMME_GO_VERSION=1.21.6
+ENV GIMME_GO_VERSION=1.22.6
```
brianmcarey, line 3: What is the reason for updating this?

images/bootstrap/runner.sh
```diff
@@ -123,6 +123,59 @@ if [[ -n "${GOOGLE_APPLICATION_CREDENTIALS:-}" ]]; then
   gcloud auth activate-service-account --key-file="${GOOGLE_APPLICATION_CREDENTIALS}" || true
 fi
 
+#Function to get access token
+get_access_token() {
```
brianmcarey, line 127: Could we create a new script for these functions in project-infra and source it in the prow job? These won't be required by every prow job.

github/ci/prow-deploy/files/jobs/kubevirt/kubevirtci/kubevirtci-postsubmits.yaml
```diff
@@ -1,5 +1,56 @@
...
+          - >
+            cat $QUAY_PASSWORD | podman login --username $(<$QUAY_USER) --password-stdin quay.io &&
+            ./publish.sh &&
+            SHORT_SHA=$(git rev-parse --short HEAD) &&
+            echo "$(git tag --points-at HEAD | head -1)" > s390x-$SHORT_SHA &&
+            gsutil cp ./s390x-$SHORT_SHA gs://kubevirt-prow/release/kubevirt/kubevirtci/s390x-$SHORT_SHA
+          # docker-in-docker needs privileged mode
+          env:
+          - name: GIMME_GO_VERSION
+            value: "1.19.9"
```
brianmcarey, line 45: The go version will have to be updated to match the x86 job

github/ci/prow-deploy/files/jobs/kubevirt/kubevirtci/kubevirtci-periodics.yaml
```diff
@@ -96,6 +96,62 @@ periodics:
...
+        export PHASES=linux; export BYPASS_PMAN_CHANGE_CHECK=true; ./publish.sh &&
+        # For passing centos image tag to dependent prow job
+        SHORT_SHA=$(git rev-parse --short HEAD) &&
+        image_tag=$(cat cluster-provision/k8s/base-image | cut -d ':' -f 2) &&
+        echo "$image_tag" > s390x-centos9-$SHORT_SHA &&
+        upload_to_gcs ./s390x-centos9-$SHORT_SHA release/kubevirt/kubevirtci/s390x-centos9-$SHORT_SHA
+      # docker-in-docker needs privileged mode
+      env:
+      - name: GIMME_GO_VERSION
+        value: "1.19.9"
```
brianmcarey, line 146: The go version will have to be updated to match the x86 job

github/ci/prow-deploy/files/jobs/kubevirt/kubevirtci/kubevirtci-postsubmits.yaml
chandramerla: added reply to same comment in other place. Once you confirm, I will do change here also.

chandramerla (commented):
@brianmcarey Thanks for your review immediately after your vacation :)

I've added initial replies/questions to your comments and would start doing changes.

github/ci/prow-deploy/files/jobs/kubevirt/kubevirtci/kubevirtci-periodics.yaml
chandramerla: Updated with first having x86 built then s390x along with publishing multi-arch manifest

images/pr-creator/Dockerfile
chandramerla: Reverted this change

github/ci/prow-deploy/files/jobs/kubevirt/kubevirtci/kubevirtci-postsubmits.yaml
chandramerla: done

chandramerla (commented):
@brianmcarey Could you please review this PR again, I've tested all the changes locally and also updated with new pr-creator image now? Thanks for your inputs so far.

chandramerla:
The pre-requisite for this PR i.e., a new bootstrap image to be rebuilt with changes are moved to a separate PR (#3648 )

github/ci/prow-deploy/files/jobs/kubevirt/kubevirtci/kubevirtci-postsubmits.yaml
```diff
@@ -47,8 +47,76 @@ postsubmits:
...
+        preset-kubevirtci-quay-credential: "true"
+      cluster: s390x-prow-virtual
+      spec:
+        volumes:
+        - hostPath:
+            path: /dev
+            type: Directory
+          name: devices
+        containers:
+        - image: quay.io/kubevirtci/golang:v20240814-9b6c4af
```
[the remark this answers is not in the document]
chandramerla: As discussed during SIG CI call, moved the bootstrap related changes from this PR to separate one #3648 , so that it can be merged before this.

github/ci/prow-deploy/files/jobs/kubevirt/kubevirtci/kubevirtci-periodics.yaml
```diff
@@ -113,11 +157,12 @@ periodics:
...
     base_ref: main
   labels:
-    preset-docker-mirror: "true"
+    preset-docker-mirror-proxy: "true"
     preset-github-credentials: "true"
     preset-podman-in-container-enabled: "true"
     preset-kubevirtci-quay-credential: "true"
-  cluster: kubevirt-prow-workloads
+    preset-gcs-credentials: "true"
+  cluster: s390x-prow-virtual
```
brianmcarey, line 165: Should this be updated to the bare-metal cluster? 
```suggestion
  cluster: prow-s390x-workloads
```

github/ci/prow-deploy/files/jobs/kubevirt/kubevirtci/kubevirtci-postsubmits.yaml
```diff
@@ -47,8 +46,76 @@ postsubmits:
...
+                if stat_gcs_file kubevirt-prow "$GCS_FILE_PATH"; then
+                    echo "File $GCS_FILE_PATH is now available."
+                    break
+                else
+                    echo "File $GCS_FILE_PATH not found. Checking again in $CHECK_INTERVAL seconds."
+                    sleep $CHECK_INTERVAL
+                fi
+            done &&
+            export KUBEVIRTCI_TAG=$(cat_gcs_file kubevirt-prow "$GCS_FILE_PATH") || { echo "Failed to fetch KUBEVIRTCI_TAG"; exit 1; } &&
+            echo "Fetched KUBEVIRTCI_TAG: $KUBEVIRTCI_TAG" &&
```
brianmcarey, line 114: Could we create a small script for this under hack/ ? 

A lot of the same logic is used in the centos periodic too. 

To simplify the job config?

github/ci/prow-deploy/files/jobs/kubevirt/kubevirtci/kubevirtci-postsubmits.yaml
```diff
@@ -47,8 +46,76 @@ postsubmits:
...
+      - org: kubevirt
+        repo: project-infra
+        base_ref: main
+      labels:
+        preset-podman-in-container-enabled: "true"
+        preset-docker-mirror-proxy: "true"
+        preset-gcs-credentials: "true"
+        preset-github-credentials: "true"
+        preset-kubevirtci-quay-credential: "true"
+      cluster: s390x-prow-virtual
```
brianmcarey, line 85:
```suggestion
      cluster: prow-s390x-workloads
```

github/ci/prow-deploy/files/jobs/kubevirt/kubevirtci/kubevirtci-postsubmits.yaml
brianmcarey: Maybe this doesn't make sense as it relies on a file in the being present in the bootstrap image.

github/ci/prow-deploy/files/jobs/kubevirt/kubevirtci/kubevirtci-periodics.yaml
```diff
@@ -126,9 +171,26 @@ periodics:
...
+            else
+                echo "File $GCS_FILE_PATH not found. Checking again in $CHECK_INTERVAL seconds."
+                sleep $CHECK_INTERVAL
+            fi
+        done &&
+        export KUBEVIRTCI_TAG=$(cat_gcs_file kubevirt-prow "$GCS_FILE_PATH") || { echo "Failed to fetch KUBEVIRTCI_TAG"; exit 1; } &&
+        echo "Fetched KUBEVIRTCI_TAG: $KUBEVIRTCI_TAG" &&
         cat $QUAY_PASSWORD | podman login --username $(<$QUAY_USER) --password-stdin quay.io &&
         ./hack/bump-centos-version.sh &&
         GIT_ASKPASS=../project-infra/hack/git-askpass.sh ../project-infra/hack/git-pr.sh -c "PHASES=linux BYPASS_PMAN_CHANGE_CHECK=true ./publish.sh" -r kubevirtci -b bump-centos-stream -T main -p $(pwd) -s "Automatic bump of CentOS Stream to latest"
```
brianmcarey, line 192: I think I would prefer the creation of the PR to remain on the x86 job - the s390x will eventually become consistent.

github/ci/prow-deploy/files/jobs/kubevirt/kubevirtci/kubevirtci-postsubmits.yaml
```diff
@@ -47,8 +46,76 @@ postsubmits:
...
+                    echo "File $GCS_FILE_PATH not found. Checking again in $CHECK_INTERVAL seconds."
+                    sleep $CHECK_INTERVAL
+                fi
+            done &&
+            export KUBEVIRTCI_TAG=$(cat_gcs_file kubevirt-prow "$GCS_FILE_PATH") || { echo "Failed to fetch KUBEVIRTCI_TAG"; exit 1; } &&
+            echo "Fetched KUBEVIRTCI_TAG: $KUBEVIRTCI_TAG" &&
+            ./publish.sh &&
             echo "$(git tag --points-at HEAD | head -1)" > latest &&
-            gsutil cp ./latest gs://kubevirt-prow/release/kubevirt/kubevirtci/latest
+            upload_to_gcs ./latest kubevirt-prow release/kubevirt/kubevirtci/latest
```
brianmcarey, line 117: Same here - I would prefer if this step remains on the x86 postsubmit - s390x will eventually exist but I don't want to block the x86 publish.

chandramerla (commented):
@brianmcarey Done some changes and added question(s).

github/ci/prow-deploy/files/jobs/kubevirt/kubevirtci/kubevirtci-periodics.yaml
brianmcarey: > @brianmcarey I understand your concern when s390x job fails.
> 
>     * If we move PR creation to x86, then
> 
>     * [ ]  s390x jobs which consume this image will fail, as image mentioned in cluster-provision/k8s/base-image doesn't exist for s390x.

This is probably acceptable for the short period of time it will take s390x to be consistent. 
> 
>     * [x]  x86 jobs which consume the image will get the latest.
> 
>     * If we keep PR creation in s390x, then
> 
>     * [x]  s390x jobs which consume this image will pass, but using "existing" version.
> 
>     * [ ]  x86 jobs which consume the image will pass, but using "existing" version. If we can't wait till s390x jobs is fixed and we want to proceed with using latest published x86 image (which is pushed already independent of PR creation with same tag as multi-arch image would get), then it needs manual PR creation instead of by kubevirt-bot, updating centos version and image tag from prow job's output.
> 
> 
> So I think this is more of a decision we've to make depending on what is important to us, s390x builds also passing with older version, till the issue is fixed or importance/urgency of having latest centos in x86 builds. As you know better, I will do changes as per your decision. Please let me know, Thanks!

From my point of view - I want to take the least intrusive path to the x86 flow. We can make follow-up changes to improve the integration once we have verified the stability of the s390x publishing etc. but for now I would prefer for s390x to take the approach of being eventually consistent with x86. 

If we see that the s390x publish is very stable and never fails then we can improve the integration.

github/ci/prow-deploy/files/jobs/kubevirt/kubevirtci/kubevirtci-postsubmits.yaml
chandramerla: As discussed moved 'creating latest file' step to x86 postsubmit

github/ci/prow-deploy/files/jobs/kubevirt/kubevirtci/kubevirtci-periodics.yaml
chandramerla: > This is probably acceptable for the short period of time it will take s390x to be consistent.

Yes, in the case when "periodic-kubevirtci-bump-centos-base-s390x" job eventually passes. But if "periodic-kubevirtci-bump-centos-base-s390x" job fails, the "publish-kubevirtci-s390x" job will start failing till 'periodic-kubevirtci-bump-centos-base-s390x job is fixed and next build is promoted' or 'CentOS bump PR reverted to old build, where both arch centos base images (+multi-arch manifest) were published'.

Similarly when "publish-kubevirtci-s390x" fails, it affects jobs using 'make cluster up'.

But still as we discussed, probably this is fine as at least this wouldn't affect jobs of kubevirt/kubevirt (because it has its own reference of kubevirtci provider version) or other repos, but just affect dependent jobs on kubevirtci repo only. **Is my understanding correct?**

With this understanding and as per our last discussion, I've moved PR creation and commiting latest file in GCS to respective x86 jobs.

github/ci/prow-deploy/files/jobs/kubevirt/kubevirtci/kubevirtci-periodics.yaml
```diff
@@ -129,6 +130,67 @@ periodics:
...
+      env:
+      - name: GIMME_GO_VERSION
+        value: "1.22.5"
+      securityContext:
+        privileged: true
+      resources:
+        requests:
+          memory: "29Gi"
+- name: periodic-kubevirtci-bump-centos-base-s390x
+  cron: "0 4 * * 2"
```
brianmcarey, line 148: Does it make sense to trigger this job after the x86 job instead of triggering them at the same time?

chandramerla: @brianmcarey I actually kept it that way to first run x86 followed by s390x. But later thought that it is not necessary, as we are waiting on x86 to be finished (write amd64-centos9-$SHORT_SHA) before triggering publish.sh in s390x. As generally x86 job finishes within 30mins max, though this job waits for that much time, it can be continued immediately.
May be we can keep a gap of 15mins or so if we think that way we can reduce pressure on cluster resource usage, due to this continuous polling for the file existance. WDYT?

brianmcarey: I would just start this job an hour after it - then we should be fairly sure that the tag will be there.

github/ci/prow-deploy/files/jobs/kubevirt/kubevirtci/kubevirtci-periodics.yaml
```diff
@@ -129,6 +130,67 @@ periodics:
...
+  decoration_config:
+    timeout: 2h
+  max_concurrency: 1
+  extra_refs:
+  - org: kubevirt
+    repo: kubevirtci
+    base_ref: main
+    workdir: true
+  labels:
+    preset-docker-mirror-proxy: "true"
```
brianmcarey, line 161: You need to remove this preset as the s390x cluster does not have a  docker mirror proxy configured.

github/ci/prow-deploy/files/jobs/kubevirt/kubevirtci/kubevirtci-periodics.yaml
chandramerla: Sure, will update this

github/ci/prow-deploy/files/jobs/kubevirt/kubevirtci/kubevirtci-periodics.yaml
chandramerla: Sure, will do

chandramerla (commented):
@brianmcarey Addressed your latest comments. Thanks.

github/ci/prow-deploy/files/jobs/kubevirt/kubevirtci/kubevirtci-postsubmits.yaml
```diff
@@ -44,11 +44,78 @@ postsubmits:
...
+        containers:
+        - image: quay.io/kubevirtci/golang:v20241014-80f340c
+          command:
+          - "/usr/local/bin/runner.sh"
+          - "/bin/bash"
+          - "-c"
+          - |
+            cat $QUAY_PASSWORD | podman login --username $(<$QUAY_USER) --password-stdin quay.io &&
+            SHORT_SHA=$(git rev-parse --short HEAD) &&
+            GCS_FILE_PATH=release/kubevirt/kubevirtci/amd64-$SHORT_SHA &&
```
brianmcarey, line 103: Why don't we just curl for the expected public path? then we could avoid the need for gcs credentials etc. here?

github/ci/prow-deploy/files/jobs/kubevirt/kubevirtci/kubevirtci-periodics.yaml
```diff
@@ -125,10 +126,70 @@ periodics:
...
+    containers:
+    - image: quay.io/kubevirtci/golang:v20241014-80f340c
+      command:
+      - "/usr/local/bin/runner.sh"
+      - "/bin/bash"
+      - "-c"
+      - |
+        # For getting centos image tag from amd64 prow job and use same for s390x and manifest-list images
+        SHORT_SHA=$(git rev-parse --short HEAD) &&
+        GCS_FILE_PATH=release/kubevirt/kubevirtci/amd64-centos9-$SHORT_SHA &&
```
brianmcarey, line 175: Why don't we just curl for the expected public path? then we could avoid the need for gcs credentials etc. here?

Like how we do here in the publish - https://github.com/kubevirt/kubevirtci/blob/1048544d270ecef75b4972bce89d8dd97434e883/publish.sh#L9

chandramerla: You are right for this part. I've changed the code locally, but later realized that as we anyway need credentials as we are removing the file `        rm_gcs_file kubevirt-prow "$GCS_FILE_PATH"` below, probably we can leave this as it is.

github/ci/prow-deploy/files/jobs/kubevirt/kubevirtci/kubevirtci-postsubmits.yaml
chandramerla: You are right for this part. I've changed the code locally, but later realized that as we anyway need credentials as we are removing the file `        rm_gcs_file kubevirt-prow "$GCS_FILE_PATH"` below, probably we can leave this as it is.

brianmcarey (approved):
/approve 

thanks @chandramerla

PR #3743, opened 2024-11-07
CDI: Fix amd64 nightly build

nunnatsa:

(Followup #3739)

bazel failed to build with BUILD_ARCH=amd64, as it's not supported. 

See failure here: https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/logs/periodic-containerized-data-importer-push-nightly-AMD64/1853996433422684160

This PR changes the arch to `x86_64`, instead.

**Release note**:
```release-note
None
```


brianmcarey (commented):
/approve

PR #3755, opened 2024-11-11
HCO: Fix K8s versions

nunnatsa:
The HCO pre-submit test lanes are with the right names but with the wrong K8s versions, that are not match to their name and too old.

This PR bumps the K8s versions, and make them match the lane names.

**Release note**:

```release-note
HCO: Bump K8s version
```


orenc1 (approved):
/lgtm
/approve

PR #3934, opened 2025-02-03
ci: default containerdisk ci run for CentOS Stream

nestoracunablanco:
Due to the lack of support for the s390x architecture in Cirros, we propose selecting CentOS Stream as the default container disk. CentOS Stream supports a broader range of architectures, making it a more versatile choice.

**Release note**:

```release-note
NONE
```


0xFelix (commented):
We are using Fedora/CentOS throughout the KubeVirt project already. Would you mind changing this to Fedora instead?

nestoracunablanco:
> We are using Fedora/CentOS throughout the KubeVirt project already. Would you mind changing this to Fedora instead?

Done.

0xFelix (commented):
/approve

PR #3943, opened 2025-02-06
ci: s390x CentOS Stream 9 containerdisks always_run job

nestoracunablanco:
This job tests the CentOS Stream 9 in the container disks on the s390x cluster for every pull request.

The containerdisks tooling publishes containerdisks to quay.io, making early detection essential from the s390x perspective. For example, an error in the tooling may produce correct images for the amd64 architecture while generating incorrect ones for s390x. Without this check, along with others, incorrectly generated s390x images could also be published.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/containerdisks/containerdisks-presubmits.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/containerdisks/containerdisks-presubmits.yaml
@@ -87,6 +87,42 @@ presubmits:
           privileged: true
       nodeSelector:
         type: bare-metal-external
+  - always_run: true
+    optional: false
+    annotations:
+      testgrid-create-test-group: "false"
+    cluster: prow-s390x-workloads
+    decorate: true
+    decoration_config:
+      grace_period: 5m0s
+      timeout: 1h0m0s
+    labels:
+      preset-podman-in-container-enabled: "true"
+    max_concurrency: 3
+    name: pull-containerdisks-pipeline-centos-stream-9-s390x
+    spec:
+      containers:
+      - command:
+        - /usr/local/bin/runner.sh
+        - /bin/sh
+        - -c
+        - ./pipeline.sh
+        env:
+        - name: GIMME_GO_VERSION
+          value: "1.22.9"
+        - name: FOCUS
+          value: "centos-stream:9"
+        - name: KUBEVIRTCI_TAG
+          value: "2501140834-56eb34e6"
+        - name: KUBEVIRT_SLIM
+          value: "true"
+        image: quay.io/kubevirtci/golang:v20250124-4f7dce8
+        name: ""
+        resources:
+          requests:
+            memory: 12Gi
+        securityContext:
+          privileged: true
   - always_run: false
     run_if_changed: "artifacts/centosstream/.*"
     annotations:


github/ci/prow-deploy/files/jobs/kubevirt/containerdisks/containerdisks-presubmits.yaml
```diff
@@ -87,6 +87,43 @@ presubmits:
...
+    annotations:
+      testgrid-create-test-group: "false"
+    cluster: prow-s390x-workloads
+    decorate: true
+    decoration_config:
+      grace_period: 5m0s
+      timeout: 1h0m0s
+    labels:
+      preset-podman-in-container-enabled: "true"
+      preset-docker-mirror-proxy: "true"
```
brianmcarey, line 101: I think you will need to remove this preset - the docker-mirror-proxy is not available in the s390x env

brianmcarey (commented):
Looks good @nestoracunablanco - just one comment inline.

github/ci/prow-deploy/files/jobs/kubevirt/containerdisks/containerdisks-presubmits.yaml
nestoracunablanco: Thanks @brianmcarey . Preset removed.

nestoracunablanco:
@chandramerla @vamsikrishna-siddu, could you please assist me with this job? It doesn't seem to start properly. Did I overlook something in the syntax?

github/ci/prow-deploy/files/jobs/kubevirt/containerdisks/containerdisks-presubmits.yaml
```diff
@@ -87,6 +87,42 @@ presubmits:
...
+          value: "2501140834-56eb34e6"
+        image: quay.io/kubevirtci/golang:v20250124-4f7dce8
+        name: ""
+        resources:
+          requests:
+            memory: 12Gi
+        securityContext:
+          privileged: true
+      nodeSelector:
+        type: bare-metal-external
```
brianmcarey, line 125: @nestoracunablanco this node selector is only used for the x86 workloads cluster - can be removed for s390x. 

Sorry I missed this in my initial look.

nestoracunablanco: @brianmcarey nodeSelector removed. Could you please try the rehearse command again? Thanks :pray:

github/ci/prow-deploy/files/jobs/kubevirt/containerdisks/containerdisks-presubmits.yaml
```diff
@@ -87,6 +87,42 @@ presubmits:
...
+        - -c
+        - ./pipeline.sh
+        env:
+        - name: GIMME_GO_VERSION
+          value: "1.22.9"
+        - name: FOCUS
+          value: "centos-stream:9"
+        - name: KUBEVIRTCI_TAG
+          value: "2501140834-56eb34e6"
+        - name: SLIM
```
brianmcarey, line 117: I think it is : 
```suggestion
        - name: KUBEVIRT_SLIM
```

But I might be wrong.

nestoracunablanco: I commited this change, let's see if it works now. Thanks for your support here!

brianmcarey:
/rehearse 

:crossed_fingers:

nestoracunablanco:
@chandramerla @vamsikrishna-siddu The job indicates that nested virtualization is not enabled on s390x. I will reach out to you both to clarify this offline.

github/ci/prow-deploy/files/jobs/kubevirt/containerdisks/containerdisks-presubmits.yaml
```diff
@@ -87,6 +87,43 @@ presubmits:
...
+    annotations:
+      testgrid-create-test-group: "false"
+    cluster: prow-s390x-workloads
+    decorate: true
+    decoration_config:
+      grace_period: 5m0s
+      timeout: 1h0m0s
+    labels:
+      preset-podman-in-container-enabled: "true"
+      preset-kubevirtci-check-provision-env: "true"
```
brianmcarey, line 101: I don't think this is needed - it is just needed for the check-provision jobs from the kubevirtci repos.

nestoracunablanco: Alright, I will delete it and check again. A simpler setup is always preferable.

github/ci/prow-deploy/files/jobs/kubevirt/containerdisks/containerdisks-presubmits.yaml
```diff
@@ -87,6 +87,42 @@ presubmits:
...
+    annotations:
+      testgrid-create-test-group: "false"
+    cluster: prow-s390x-workloads
+    decorate: true
+    decoration_config:
+      grace_period: 5m0s
+      timeout: 1h0m0s
+    labels:
+      preset-podman-in-container-enabled: "true"
+    max_concurrency: 11
```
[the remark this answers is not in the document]
nestoracunablanco: @brianmcarey, done! Feel free to retrigger the rehearsal whenever you have time.

chandramerla:
Looks good to me.
In future, once we've good k8s provider builds on s390x from latest file in GCS, we can remove below lines:
```
        - name: KUBEVIRTCI_TAG
          value: "2501140834-56eb34e6"
```

0xFelix:
I suppose we should wait for https://github.com/kubevirt/containerdisks/pull/272#issuecomment-2650394980 first?

/hold

nestoracunablanco:
Hello @0xFelix, I believe we don't need to wait for the changes in container disks. The issue I'm addressing there only impacts the Fedora tests. Therefore, I'm working in parallel to enable the pipelines for other distributions in the project-infra repository.

0xFelix:
Not really sure if it is worth to have this job. `pipeline.sh` is about testing containerdisks / medius itself. That works on any arch, I don't think we need multiple for that. The built images need to be tested in a later stage, e.g. the periodic building and pushing the images.

nestoracunablanco:
@0xFelix, I understand your point, but I’m concerned that some changes might break the s390x architecture. For instance, modifications to the tools could result in successful builds for both amd64 and s390x, but produce an incorrect executable format for the latter. In this scenario, early detection is crucial for us to prevent any issues from being published.

0xFelix (approved):
I still don't fully understand why this tool needs to run on s390x. It's an upstream focused tool with the single purpose of publishing containerdisks to quay.io/containerdisks. I.e. there should be no need to run this in private somewhere else.

The pipeline itself looks fine to me, I'm just thinking about conserving resources.

/lgtm

/hold
@brianmcarey If you see this additional job not as an issue, please unhold.

nestoracunablanco:
@brianmcarey, what are your thoughts on @0xFelix's last comment? From my perspective, the cluster prow-s390x-workloads is maintained by IBM, which should mean minimal resource consumption on your end. Please let me know if my assumptions are incorrect.

brianmcarey:
> @brianmcarey, what are your thoughts on @0xFelix's last comment? From my perspective, the cluster prow-s390x-workloads is maintained by IBM, which should mean minimal resource consumption on your end. Please let me know if my assumptions are incorrect.

I don't think the additional job should be an issue. 

@nestoracunablanco can you just add some details around why you want this job in the PR description? Are we publishing s390x containerdisks?

nestoracunablanco:
> > @brianmcarey, what are your thoughts on @0xFelix's last comment? From my perspective, the cluster prow-s390x-workloads is maintained by IBM, which should mean minimal resource consumption on your end. Please let me know if my assumptions are incorrect.
> 
> I don't think the additional job should be an issue.
> 
> @nestoracunablanco can you just add some details around why you want this job in the PR description? Are we publishing s390x containerdisks?

@brianmcarey, I've added more details to the description. Please let me know if this is sufficient. As for your second question, yes, the container disks are being published as multiarch [here](https://quay.io/repository/containerdisks/centos-stream?tab=tags). We would also like to implement some checks for parity to ensure that we don't publish anything that doesn't work.

brianmcarey (commented):
thanks @nestoracunablanco 

/lgtm

/hold cancel

PR #4009, opened 2025-03-11
revert: set the job as optional until the s390x pipelines are stable.

nestoracunablanco:
With the s390x pipeline now stable, we can proceed to restore the job configurations.

The root cause of the issue was the disabling of nested virtualization.

This reverts commit a7efd1940af0e1487d9bd2c75e7b2a2cad0dd7df.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/containerdisks/containerdisks-presubmits.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/containerdisks/containerdisks-presubmits.yaml
@@ -88,7 +88,7 @@ presubmits:
       nodeSelector:
         type: bare-metal-external
   - always_run: true
-    optional: true
+    optional: false
     annotations:
       testgrid-create-test-group: "false"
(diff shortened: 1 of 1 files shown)


nestoracunablanco:
@codingben @ksimon1 what do you think about this change?

github/ci/prow-deploy/files/jobs/kubevirt/containerdisks/containerdisks-presubmits.yaml
```diff
@@ -88,7 +88,7 @@ presubmits:
       nodeSelector:
         type: bare-metal-external
   - always_run: true
-    optional: true
+    optional: false
```
brianmcarey, line 91: @nestoracunablanco just a nit comment can we be consistent here? for the rest of the jobs you have just removed `optional: true`- can you do the same here?

brianmcarey (commented):
/approve 

Just had one tiny nit comment but it should be ok.

PR #4111, opened 2025-04-23
Fix the periodic-kubevirtci-bump-kubevirt job

nunnatsa:
The current configuration is trying to call git-pr.sh with `-d`, pointing to not existing script. That causes failing of the job.

Fixing by removing the `-d` parameter and adding the `-R` to add release-notes of None in the PR description.

**Release note**:

```release-note
Fix the periodic-kubevirtci-bump-kubevirt job
```


brianmcarey (approved):
/approve

PR #4155, opened 2025-05-08
Fix the periodic-kubevirtci-bump-hco job

nunnatsa:
This job is trying to push PRs to the kubevirt repository, instead of to the hyperconverged-cluster-operator repository. see here for example: https://github.com/kubevirt/kubevirt/pull/14659

**Release note**:

```release-note
Fix the periodic-kubevirtci-bump-hco job
```


enp0s3 (approved):
/approve

PR #4186, opened 2025-05-29
Allow adding authentication patterns for mirrored urls

rmohr:

In case of private GCS buckets, make use of the popular auth pattern support in all currently mirrored bazel rules and add .netrc pattern rules for google cloud storage. Bazel will look for the credentials in ~/.netrc and use them if present.

The rules get added if `-requires-auth` is added.

**Release note**:

```release-note
NONE
```


brianmcarey (commented):
/lgtm 

thanks @rmohr !

rmohr:
> thanks @rmohr !

Thanks for the review brian!

PR #4251, opened 2025-07-02
Release-1.6: Kubernetes 1.33 voting

xpivarc:


xpivarc:
The rehearse is failing because the provider doesn't enable the fg, see:
```
16:01:14: time="2025-07-04T16:01:14Z" level=info msg="[node 2]: echo \"KUBELET_EXTRA_ARGS=--cgroup-driver=systemd --runtime-cgroups=/systemd/system.slice --kubelet-cgroups=/systemd/system.slice --fail-swap-on=false  --feature-gates=NodeSwap=true --cpu-manager-policy=static --kube-reserved=cpu=500m --system-reserved=cpu=500m\" | tee /etc/sysconfig/kubelet > /dev/null"
```

This happens because we take version from `--prefix` argument. That is `rehearsal-pull-kubevirt-e2e-k8s-1.33-sig-compute-1.6` in this case. The regex for version matches the 1.6 instead of intended 1.33. 

/cc @dhiller @brianmcarey 

The version needs to be taken from `KUBEVIRT_PROVIDER=k8s-1.33`

xpivarc:
https://github.com/kubevirt/kubevirtci/pull/1478

xpivarc:
PTAL @brianmcarey @dhiller

brianmcarey (approved):
/approve

thanks @xpivarc

PR #4252, opened 2025-07-02
Main: Kubernetes 1.33 voting

xpivarc:


brianmcarey (commented):
Thanks for this - can you also remove ` pull-kubevirt-e2e-k8s-1.32-sig-compute-serial` as part of this PR as the 1.33 serial lane is going required?

PR #4261, opened 2025-07-08
feat: add periodic job for containerdisks on s390x

nestoracunablanco:
This job utilizes the pipeline-periodic script while configuring environment variables to prevent image publication. The responsibility for publishing the images lies with the amd64 job.

**Release note**:

```release-note
NONE
```


nestoracunablanco:
@ashokpariya0 FYI

PR #4319, opened 2025-08-20
[kubevirt/kubevirt]: disable PSA in release-1.1 job configurations

enp0s3:

there is a problem with pod bring up latency in release-1.1 the root cause comes from the VM based providers and the trigger is usage of custom seccomp provider in the kubevirt e2e tests.

In order to keep the lanes stable we temporarily opt out from PSA and from using seccomp.

We are aware that this will reduce coverage in upstream e2e, but we also must have a way to merge critical fixes.
We will keep tracking this situation.

**Release note**:

```release-note
NONE
```


brianmcarey (approved):
/approve 

thanks @enp0s3

PR #4327, opened 2025-08-27
disable PSA in release-1.0 job configurations

fossedihelm:
there is a problem with pod bring up latency in release-1.0 the root cause comes from the VM based providers and the trigger is usage of custom seccomp provider in the kubevirt e2e tests.

In order to keep the lanes stable we temporarily opt out from PSA and from using seccomp.

We are aware that this will reduce coverage in upstream e2e, but we also must have a way to merge critical fixes.
We will keep tracking this situation.

ref: https://github.com/kubevirt/project-infra/pull/4319

**Release note**:

```release-note
NONE
```


enp0s3:
@dhiller Hi, thank you for looking into this!

The manual cherry-pick is here: https://github.com/kubevirt/kubevirt/pull/14570

So we are waiting for this PR to get merged so we can test the fix.

brianmcarey (approved):
/approve 

thanks @fossedihelm 

/hold cancel

PR #4328, opened 2025-08-27
[kubevirt/kubevirt]: make prom-rule-verify run conditional

enp0s3:

We can save some CI resources and time.

the job should run only when hack/prom-rule-ci/* has changed this make target is related to unit tests of prometheus alerts and recording rules. The tests themselves reside in prom-rule-ci directory so there is no point to run this target unless the tests of the image were changed.

**Release note**:

```release-note
NONE
```


brianmcarey (approved):
/approve 

thanks @enp0s3 - does it make sense to apply this to some of the recent release branches?

dhiller (commented):
Thank you @enp0s3 ! ❤️

dhiller (approved):

PR #4374, opened 2025-09-16
Request 16G RAM for nmstate tests

cathay4t:
The rust compiler for nmstate takes more than 4G RAM. Hence request 16G
RAM for its tests.


dhiller (approved):
/approve

Thanks @cathay4t ! 🎉

PR #4655, opened 2026-01-20
botreview: remove code base from the repository

dollierp:

The botreview custom external plugin didn't really work out and is not enabled anymore in KubeVirt CI Prow instance.

Hence, it is basically dead code and we won't spend more efforts on it.

Instead, we will remove the plugin code base entirely from this repository.

/cc @dhiller


dollierp:
Waiving build-pr-creator-image test because the failure is known and not related to this PR. It is addressed in another PR.

/override build-pr-creator-image

PR #4657, opened 2026-01-20
fix(prow-bump): improve prow changelog generation

dollierp:

When the script generating prow changelog is invoked, the changes from the autobumper are not committed yet, hence, we need to use `git diff` instead of `git show` to parse them.

If `git show` is used, the changelog is generated for the previous PR which modified the file.

/cc @dhiller


dollierp:
Needs to improve the `git diff` command to not ignore new files.

/hold

PR #4720, opened 2026-02-17
chore(jobs): reduce memory requests close to actual use

dollierp:

While the resources are not actually used, they remain reserved and the jobs spend a lot of time in the queue waiting for the resources to be released.

This PR reduce the memory requests for the following project-infra jobs:
- build-pr-creator-image
- build-vm-image-builder-image
- publish-pr-creator-image
- publish-vm-image-builder-image
- pull-project-infra-grafana-deploy-test
- pull-project-infra-prow-deploy-test

Each of them was rehearsed 10 times and no OOMKill was observed.

/cc @dhiller


dhiller (approved):
/approve

/hold to be removed when last rehearsal is successful

Issue #4785, opened 2026-03-06
Add Prow build cluster prow-s390x-openshift


cfilleke:
This is the PR described in https://github.com/kubevirt/project-infra/blob/main/docs/how-to-add-a-prow-cluster.md#cluster-administrator to create test automation named CDI-Mainframe for the CDI component of KubeVirt to be tested on an external s390x cluster. Attached is a kubeconfig file encrypted with @dhiller public key for accessing the cluster's  kubevirt-prow-jobs namespace as the prow-workloads-cluster-automation service account. 

[oz-prow-workloads-cluster-automation-kubeconfig.tar.gz](https://github.com/user-attachments/files/25782217/oz-prow-workloads-cluster-automation-kubeconfig.tar.gz)

dollierp:
Hi @cfilleke 

AFAIK, the three existing s390x clusters used by upstream KubeVirt CI are not managed by RH.
- prow-s390x-e2e-test
  - added: https://github.com/kubevirt/project-infra/issues/4284
  - owner: @vamsikrishna-siddu
  - jobs: [https://prow.ci.kubevirt.io/?job=*-e2e-test-S390X\*](https://prow.ci.kubevirt.io/?job=*-e2e-test-S390X*)
- prow-s390x-secure-execution
  - added: https://github.com/kubevirt/project-infra/pull/4452
  - owner: @jschintag
  - jobs: [https://prow.ci.kubevirt.io/?job=*secure-execution\*](https://prow.ci.kubevirt.io/?job=*secure-execution*)
- prow-s390x-workloads
  - added: https://github.com/kubevirt/project-infra/issues/3140
  - owner: @chandramerla?
  - jobs: [https://prow.ci.kubevirt.io/?cluster=prow-s390x-workloads](https://prow.ci.kubevirt.io/?cluster=prow-s390x-workloads)

PR #4832, opened 2026-03-16
Add s390 prow e2e job

Davo911:
Adds `pull-containerized-data-importer-e2e-s390x`, a presubmit job that runs CDI functional e2e tests against an external s390x OpenShift cluster, similar to how kubevirt runs [periodic-kubevirt-e2e-test-S390X.](https://github.com/kubevirt/project-infra/blob/a81444c44019908fca1a1c85f59f197c93a5f759/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-periodics.yaml#L766-L831)

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Related**:
https://github.com/kubevirt/project-infra/issues/4785

**Release note**:

```release-note
NONE
```

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/containerized-data-importer/containerized-data-importer-presubmits.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/containerized-data-importer/containerized-data-importer-presubmits.yaml
@@ -470,6 +470,79 @@ presubmits:
           resources:
             requests:
               memory: "52Gi"
+  - name: pull-containerized-data-importer-e2e-s390x
+    skip_branches:
+      - release-v\d+\.\d+
+    annotations:
+      fork-per-release: "true"
+      testgrid-dashboards: kubevirt-containerized-data-importer-presubmits
+    always_run: true
+    optional: true
+    decorate: true
+    decoration_config:
+      timeout: 6h
+      grace_period: 10m
+    max_concurrency: 1
+    cluster: kubevirt-prow-control-plane
+    extra_refs:
+    - org: kubevirt
+      repo: project-infra
+      base_ref: main
+    labels:
+      preset-podman-in-container-enabled: "true"
+      preset-docker-mirror-proxy: "true"
+      preset-kubevirtci-quay-credential: "true"
+      preset-github-credentials: "true"
+      preset-pgp-bot-key: "true"
+    spec:
+      containers:
+        - image: quay.io/kubevirtci/bootstrap:v20260306-351d3b1
+          env:
+          - name: KUBECONFIG
+            value: /kubeconfig
+          - name: TARGET
+            value: external
+          - name: BUILD_ARCH
+            value: crossbuild-s390x
+          - name: ARCHITECTURE
+            value: s390x
+          - name: PULL_POLICY
+            value: Always
+          - name: CDI_NAMESPACE
+            value: cdi
+          - name: CDI_E2E_SKIP
+            value: "Destructive|Cert rotation|cdi-apiserver tests|Importer Test Suite-prometheus"
+          - name: SNAPSHOT_SC
+            value: ocs-storagecluster-ceph-rbd
+          - name: BLOCK_SC
+            value: ocs-storagecluster-ceph-rbd
+          - name: CSICLONE_SC
+            value: ocs-storagecluster-ceph-rbd
+          - name: CDI_LABEL_FILTER
+            value: "!VDDK && !ImageIO"
+          - name: UPLOAD_PROXY_URL_OVERRIDE
+            value: "https://cdi-uploadproxy-cdi.apps.odf.cdi.ci:6005"
+          command:
+            - "/usr/local/bin/runner.sh"
+            - "/bin/sh"
+            - "-c"
+            - |
+              source ../project-infra/hack/manage-secrets.sh
+              decrypt_secrets
+              install -m400 "${secrets_repo_dir}"/secrets/kubeconfigs/prow-s390x-openshift /kubeconfig
+              cleanup_secrets
+
+              cat "$QUAY_PASSWORD" | docker login --username $(cat "$QUAY_USER") --password-stdin=true quay.io
+              echo "195.212.74.206 api.odf.cdi.ci oauth-openshift.apps.odf.cdi.ci cdi-uploadproxy-cdi.apps.odf.cdi.ci" >> /etc/hosts
+
+              kubectl delete cdi cdi --ignore-not-found --timeout=120s
+              kubectl delete ns cdi --ignore-not-found --timeout=120s
+              automation/test.sh
+          securityContext:
+            privileged: true
+          resources:
+            requests:
+              memory: "8Gi"
   - name: pull-containerized-data-importer-e2e-ceph-wffc
     skip_branches:
       - release-v\d+\.\d+


dhiller (changes requested):

Davo911:
did bumping `rules_oci` fix this for kubevirt?  https://github.com/kubevirt/kubevirt/pull/17138 @brianmcarey

Davo911:
@dollierp we again seem to hit that `extra_refs` issue from earlier

dollierp:
> @dollierp we again seem to hit that `extra_refs` issue from earlier

My bad, I totally forgot that, I retriggered a rehearsal job manually based on `mkpj`.

dollierp:
Looks like it failed to pull some images not available for s390x:
- quay.io/kubevirt/imageio-init:latest
- quay.io/kubevirt/vcenter-simulator:latest

cfilleke:
> Looks like it failed to pull some images not available for s390x:
> 
>     * quay.io/kubevirt/imageio-init:latest
> 
>     * quay.io/kubevirt/vcenter-simulator:latest

Hi, Denis - imageio & vcenter are not supported in s390x so it should not be trying to pull any images associated with those features in the first place.

Davo911:
That's why it should be filtered out in the cluster-sync step
https://github.com/kubevirt/containerized-data-importer/blob/671679e0bba2875687130c1cfc196631d149fa04/cluster-sync/sync.sh#L228-L236

but the logs tell us
```shell
ARCHITECTURE=x86_64
``` 

is ARCHITECTURE -> cluster arch ?
and HOST_ARCHITECTURE -> prow machine ?

Then just adding `ARCHITECTURE=s390x ` to the job should fix this

Davo911:
Could you trigger the rehearsal again please @dollierp

dollierp:
> Could you trigger the rehearsal again please @dollierp

Done. IIUC, new rehearsals are now failing to deploy CDI.

Davo911:
~It tries to deploy `latest`, but that's is not multiarch...~
The `ARCHITECTURE=s390x` also didn't really propagate through, so it  would still deploy vcenter/imageio i suspect
It's in the commit, but not in the [Prow Job YAML](https://prow.ci.kubevirt.io/prowjob?prowjob=rehearsal-pull-containerized-data-importer-e2e-s390x-c4725)
Any idea why this could be?

Davo911:
Maybe because I amended and force pushed?

dollierp:
> Maybe because I amended and force pushed?

My bad, I didn't properly merge the changes between the previous ProwJob and the one re-generated after your changes.

Retriggered with a proper manifest this time (hopefully).

Davo911:
@dollierp mhh i checked the cluster and there were multiple stale CDI namespaces around and a stuck cdi CR.
Just cleaned them up, could you give it another try.

~I also suspect we should add a fixed namespace, because these randomzied ones never cleaned up and we're only running with no concurrency, what do you think?~
```yaml
- name: CDI_NAMESPACE
  value: cdi
```
Edit: https://github.com/kubevirt/containerized-data-importer/blob/a0f9566d61db5a8d578c19bc7cf5173f5f4429b3/automation/test.sh#L27 always overrides the ns..

dollierp:
> mhh i checked the cluster and there were multiple stale CDI namespaces around and a stuck cdi CR. Just cleaned them up, could you give it another try.

Re-triggered, the tests are now executed but failing with
```
lookup cdi-uploadproxy-cdi-ll82p.apps.odf.cdi.ci: no such host
```

Davo911:
This confirms, that we have a problem with this namespace randomisation ..
I cannot determine yet if this DNS issue is a prow problem or a cluster problem
Maybe @awels got an idea?

Another Issue seems to be the cleanup process, as as all ressources are left on the cluster after finishing the job

awels:
We purposely randomize the CDI namespace, to catch anyone hard coding `cdi` as their namespace in tests. You can override the randomization if you set the `CDI_NAMESPACE` environment variable to `cdi`.

Davo911:
@awels Problem only is, that the [test.sh overwrites that](https://github.com/kubevirt/containerized-data-importer/blob/a0f9566d61db5a8d578c19bc7cf5173f5f4429b3/automation/test.sh#L27), already opened a [PR](https://github.com/kubevirt/containerized-data-importer/pull/4100) for that

awels:
I see, thanks for the PR, I though line 77 would fix it, but we overrode the value passed in.

Davo911:
https://github.com/kubevirt/containerized-data-importer/pull/4100 got merged
Could we give it another try @dollierp

awels:
it looks like it is not using the random cdi namespace anymore, but something else is still wrong.

dollierp:
> it looks like it is not using the random cdi namespace anymore, but something else is still wrong.

It sees the name resolution issue for the cluster fqdn is still present.

Also the service account `prow-workloads-cluster-automation` needs to be granted more permissions because the jobs is using the kubevirtci external provider.

Davo911:
Added the permissions missing, thanks @dollierp , but I'm unsure if the connection failure to the `cdi-uploadproxy-cdi.apps.odf.cdi.ci` is a cluster issue or a prow connection issue...

In my local enviroment a simple hosts entry fixed this error

dollierp:
> I'm unsure if the connection failure to the `cdi-uploadproxy-cdi.apps.odf.cdi.ci` is a cluster issue or a prow connection issue...

IIUC, tests are using this URL because it is not overridden for external clusters (https://github.com/kubevirt/containerized-data-importer/blob/main/cluster-sync/sync.sh#L322-L328).

Usually OpenShift clusters have their `*.apps."${cluster_name}"."${cluster_domain}"` publicly resolvable.

If that's not the case for your cluster I guess you can simply modify the Prow job to add this record to the `/etc/hosts` file inside the Prow pod before running the `automation/test.sh` script.

Davo911:
Well I added that, let's try another run please @dollierp

Davo911:
There seemed to be an issue setting up cdi, the cluster looks good though, there was one stuck ressource i cleaned up

Davo911:
I successfully kicked off a few runs manually, I'd give it another go, before considering to reinstall the whole cluster..

Davo911:
Ah found the problem and was able to replicate it.

My manual tests deployed `latest`, while the Prow job's upgrade test tries to install `v1.65.0` first — which is a downgrade the operator seems to rejects (`Reconciling to error state, illegal phase`). 
Properly cleaning up fixed it for me locally, so I added that step to the job..

Davo911:
`ERROR: Post "https://cdi-uploadproxy-cdi.apps.odf.cdi.ci/v1beta1/upload": dial tcp 172.16.41.35:443: connect: connection timed out`

at least it finds the host now, but times out connecting..

dollierp:
> `ERROR: Post "https://cdi-uploadproxy-cdi.apps.odf.cdi.ci/v1beta1/upload": dial tcp 172.16.41.35:443: connect: connection timed out`
> 
> at least it finds the host now, but times out connecting..

The ginkgo test program is executed in a Pod on the KubeVirt Prow Control Plane cluster.

The IP address `172.16.41.35` is private and can't be reached from this cluster.

Davo911:
Yes, I just talked t an admin about this. 
For one thats a internal IP, I'll adjust it to an external, but also there is a port mapping happening, so trying to reach `https://cdi-uploadproxy-cdi.apps.<cluster>:443` won't work, we have to speak to a different port. 

I'm adjusting this with iptables, but in future it would be more elegant to be able to set the uploadproxURL also for external clusters IMO

awels:
> I'm adjusting this with iptables, but in future it would be more elegant to be able to set the uploadproxURL also for external clusters IMO

You can set the uploadProxyURL In the CDI CR.

Davo911:
@awels I guess through 
`kubectl patch cdi ${CR_NAME} --type=merge -p '{"spec": {"config": {"uploadProxyURLOverride": <url>}}}'`
But how would this be done when using the `/automation/test.sh` flow?

awels:
You already found the code that sets it for internal providers in sync.sh. We should allow setting an ENV variable to override that. Then in sync.sh see if that variable exists, and set it.

Davo911:
Fix is merged and parameter in place, could we give it another go? @dollierp

Davo911:
Still seeing 
```shell
  ERROR: Post "https://cdi-uploadproxy-cdi.apps.odf.cdi.ci/v1beta1/upload-async": dial tcp 195.212.74.206:443: connect: connection timed out
``` 

even though we succesfully patched
```shell
+ kubectl patch cdi cdi --type=merge -p '{"spec": {"config": {"uploadProxyURLOverride": "https://cdi-uploadproxy-cdi.apps.odf.cdi.ci:6005"}}}'
cdi.cdi.kubevirt.io/cdi patched
```

Davo911:
Ok as far as I understand [automation/test.sh](https://github.com/kubevirt/containerized-data-importer/blob/6957003039f0565f16066c1cd085cebf146e9057/automation/test.sh) auto-sets `UPGRADE_FROM=v1.65.0`, so the job does a v1.65.0 -> latest upgrade before patching. 
The config controller doesn't seem to pick up the override after that upgrade reconciliation. 
This is just a guess. 

I'm adding a 
```yaml
- name: RANDOM_CR
  value: "true"
```
just like the amd64 ceph jobs

Could we run this again  @dollierp 
After this iptables approach would be my only other option

Davo911:
I tried to pinpoint where the override was and posted a PR for this, I'd be happy if I'd get a review.
Until then I'd like to try the port mapping approach to see a successful rehearsal job and see if there are any other issues. @dollierp

dhiller:
ran the job here: https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_containerized-data-importer/4127/pull-containerized-data-importer-e2e-s390x/2054559671112437760

Davo911:
thanks @dhiller could we hearse again?

Davo911:
https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_containerized-data-importer/4162/pull-containerized-data-importer-e2e-s390x/2059922470143528960

Very promising, the PR did what it supposed to do
the 6 failures left are 
`dial tcp 127.0.0.1:18443: connect: connection refused` --> port-forward problem, maybe realted to external provider?
and
`Timeout waiting for DataImportCron conditions` --> flake?

I'd suggest to just skip them? 

`CDI_E2E_SKIP=Destructive|Cert rotation|cdi-apiserver tests|Importer Test Suite-prometheus`

awels:
> https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_containerized-data-importer/4162/pull-containerized-data-importer-e2e-s390x/2059922470143528960
> 
> Very promising, the PR did what it supposed to do the 6 failures left are `dial tcp 127.0.0.1:18443: connect: connection refused` --> port-forward problem, maybe realted to external provider? and `Timeout waiting for DataImportCron conditions` --> flake?
> 
> I'd suggest to just skip them?
> 
> `CDI_E2E_SKIP=Destructive|Cert rotation|cdi-apiserver tests|Importer Test Suite-prometheus`

For the upload failures, you are probably running a real cluster with a real Ingress/Route. For kubevirtci we use a nodePort that is where the `127.0.0.1` comes from. But you should be able to determine the actual address (or see if CDI can figure it out for you). But if you look in the bash scripts there is some code that figures out what the nodePort is and sets the `uploadProxyURLOverride` to the nodePort. I would see if CDI figures it out if you don't override it. If not you should be able to determine the Ingress/Route and manually set it.

Davo911:
Added the skip

PR #4929, opened 2026-04-06
chore(prow-deploy): optimize image build

dollierp:

Reduces the image build time and size by ~10%.

/cc @dhiller


.gitignore
```diff
@@ -1,6 +1,4 @@
 github/ci/**/group_vars/all/main.yml
-github/ci/prow-deploy/vars/ibmcloud-production/secrets.yml
```
dhiller, line 2: nit: since the pathes don't exist anymore this makes sense - however this was meant as a security guardrail to avoid accidentally checking in secrets files. Maybe we rather update them to the new ones instead of removing them?

images/prow-deploy/Dockerfile
```diff
@@ -2,30 +2,27 @@ FROM quay.io/kubevirtci/golang:v20260319-c8f1db8 as builder
...
 
 FROM quay.io/kubevirtci/bootstrap:v20260319-c8f1db8
 
-COPY --from=builder /usr/local/bin/config-bootstrapper /usr/local/bin/phony /usr/local/bin/hmac /usr/local/bin/
+COPY --from=builder /usr/local/bin/phony /usr/local/bin/
+COPY --from=cfg-bootstrap /ko-app/config-bootstrapper /usr/local/bin/
+COPY --from=hmac /ko-app/hmac /usr/local/bin/
 
-RUN curl -Lo ./kustomize.tar.gz https://github.com/kubernetes-sigs/kustomize/releases/download/kustomize/v3.8.7/kustomize_v3.8.7_linux_amd64.tar.gz && \
+RUN curl -fLo ./kustomize.tar.gz https://github.com/kubernetes-sigs/kustomize/releases/download/kustomize/v3.8.7/kustomize_v3.8.7_linux_amd64.tar.gz && \
```
dhiller, line 17: nit: while we're at that maybe update to more recent?

images/prow-deploy/requirements.txt
```diff
@@ -1,7 +0,0 @@
-ansible == 12.2.0
```
dhiller, line 1: Why is the requirements file removed? It is still referenced in the Containerfile? https://github.com/kubevirt/project-infra/blob/338d68c4987b7b5508832d2ac4378f49c65a29eb/images/prow-deploy/Dockerfile#L29

dollierp: No, it has been replaced by `dnf install ansible-core` in this PR.

images/prow-deploy/Dockerfile
dollierp: I tried but it seems our kustomizations are not compatible with newer kustomize versions.

I will try to fix that after all the manifests changes for the transition to a GitHub app are completed.

.gitignore
dollierp: Right, I replaced them instead of removing them.

dhiller (approved):
/approve

PR #4958, opened 2026-04-15
chore(webhooks): rotate secrets after GitHub notice

dollierp:

GitHub sent a notice stating that the webhook secrets were leaked in HTTP Headers by GitHub between September 2025 and January 2026.

The risk of having leaked the secret externally is very low because it seems Hook and external plugins don’t log the headers (and the logs are already rotated anyway).

Regardless, to be 100% safe, all the managed_webhooks should be rotated.

/cc @dhiller


dhiller (approved):
/approve

PR #4960, opened 2026-04-15
Fix(coverage): Hardcode toolchain and coverage paths

Whitedyl:
After merging the previous fix (PR #4933), the coverage plugin successfully created pods but jobs failed due to triggering e2e tests.

This PR fixes two issues: 
e2e tests failing: Running go test ./... included e2e/integration tests. The test targets now match the Makefile's COVERAGE_TARGETS (./external-plugins/... ./releng/... ./robots/... ./cmd/... ./pkg/...).

Go toolchain mismatch: The go.mod specifies go 1.25.0 with toolchain go 1.25.1. The entrypoint installs Go 1.25.0 via gimme, but Go then downloads 1.25.1 at runtime, causing a compile tool version mismatch. Setting GOTOOLCHAIN=local tells Go to use the installed version and ignore the toolchain directive.

@dhiller @dollierp 

The test targets and GOTOOLCHAIN value are hardcoded for now, these will be moved to a ConfigMap in a follow-up PR.

### DIFF
--- a/external-plugins/coverage/plugin/handler/handler.go
+++ b/external-plugins/coverage/plugin/handler/handler.go
@@ -125,13 +125,17 @@ func (h *GitHubEventsHandler) generateCoverageJob(
 							"-ce",
 						},
 						Args: []string{
-							"go test ./... -coverprofile=${ARTIFACTS}/filtered.cov && " +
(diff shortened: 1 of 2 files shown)


dhiller (approved):
/approve

PR #4962, opened 2026-04-15
feat(prow): bootstrap transition to GitHub App credentials

dollierp:

* create relevant secrets
* remove unused manifests from k8s test_infra
* add manifests from upstream prow but do not use them yet

The only difference in manifests generated by kustomize is a new secret to be used in follow up PRs:

```
kind: Secret
apiVersion: v1
metadata:
  name: github-token
  namespace: kubevirt-prow
type: Opaque
data:
  appid: <Prow GitHub Application ID>
  cert: <Prow GitHub Application private key>
  # For backwards compatibility
  oauth: <Same as oauth-token secret>
  token: <Same as oauth>
```

/cc @dhiller


dhiller:
looks like this pr still has issues right?

dollierp:
> looks like this pr still has issues right?

I hope not :)

The failure of pull-project-infra-prow-deploy-test is expected due to changes in this PR but is also fixed by updating the job definition in this PR, that's why I had to rehearse it.

dhiller:
> > looks like this pr still has issues right?
> 
> I hope not :)
> 
> The failure of pull-project-infra-prow-deploy-test is expected due to changes in this PR but is also fixed by updating the job definition in this PR, that's why I had to rehearse it.

Thanks for clarification!

github/ci/prow-deploy/hack/deploy.sh
```diff
@@ -1,17 +1,47 @@
...
+    export GIT_ASKPASS="${project_infra_root}"/hack/git-askpass.sh
+    cat << EOF > inventory
+[local]
+localhost ansible_connection=local
+EOF
+    ANSIBLE_ROLES_PATH="${base_dir%/*}" ansible-playbook -i inventory prow-deploy.yaml
+}
+
+# Generate a consolidated kubeconfig to use in Prow secrets
+generate_kubeconfig(){
```
dhiller: nit: rename the function, this way you could remove the comment.

```suggestion
generate_consolidated_kubeconfig(){
```

github/ci/prow-deploy/kustom/base/manifests/local/mirror-service.yaml
```diff
@@ -1,9 +0,0 @@
-apiVersion: v1
```
dhiller: The removal of `mirror-service.yaml` and the `docker-mirror_deployment.yaml` strategic merge patch is not mentioned in the PR description. Was this intentional cleanup? If so, worth a brief note in the PR body for reviewers. If not, it may warrant a separate commit.

github/ci/prow-deploy/hack/sync-manifests.sh
```diff
@@ -0,0 +1,75 @@
...
+            | select(test("^(v?\d{8}-(?:v\d(?:[.-]\d+)*-g)?[0-9a-f]{6,10})$"))
+          '
+)
+
+sed -re "s|v[0-9]{8}-[a-f0-9]{6,10}|${latest_prow_version}|" -i "${manifests_base}"
+
+# Split the monolitic manifest into subfiles using the naming patterns:
+# - name_kind.yml for resources in the main prow namespace
+# - name_kind_namespace.yml for resources in the test pods namespace
+rm -f *.yml
```
dhiller: nit: this rm -f *.yml relies on an implicit convention that downloaded files use .yaml (preserved) while split outputs use .yml (cleared) - now if someone adds a manually-maintained .yml file to this directory it would be silently deleted.

It would be worthwile adding a README.md in the upstream directory with a comment explaining the .yaml vs .yml extension convention so future maintainers understand they have to avoid adding .yml files.

dhiller (approved):
/approve

/hold to be removed when ready

github/ci/prow-deploy/hack/deploy.sh
dollierp: Function renamed and comment removed.

github/ci/prow-deploy/hack/sync-manifests.sh
dollierp: Actually, it was not my intention to have a `.yaml` vs `.yml` convention. It is just a side effect of `yq` using `.yml` extension with the `--split-exp` option.

The `rm` command is now executed from the begining and is now cleaning up previous both .yaml and .yml manifests before downloading and generating the new ones.

dhiller (approved):
/unhold

dollierp:
Rehearsal passed.

PR #4995, opened 2026-04-23
fix(git-pr.sh): correct script invocation in some periodics

dollierp:

When switching to long options, yaml multi-line strings where not properly used in some periodic jobs.

This PR aims to fix those jobs.

/cc @dhiller


dhiller (approved):
/approve

Thank you @dollierp ! 💯

PR #5012, opened 2026-04-27
SEV: disable lanes reporting

dollierp:

While the SEV lanes are declared with `always_run: false`, they are still triggered by the phased plugin and blocking the PRs from getting merged.

As a workaround, disable the reporting of those lanes so that they are actually ignored.

/cc @dhiller


dhiller (approved):
/approve

PR #5021, opened 2026-04-28
chore(kubectl-virt-plugin): use common bootstrap image

dollierp:

Most of the workflow is now handled by a GitHub Action, the remaining code is mainly `curl` commands which do not require a dedicated image and specific privileges.

/cc @dhiller


dhiller (approved):
/approve

PR #5072, opened 2026-05-19
Add prow-job-taskforce github team to override plugin

Whitedyl:
Previously only top level approvers could override jobs but this PR adds prow-job-taskforce GitHub team into the override plugin allowing members to override jobs inside PRs across the kubevirt org.

/cc @dhiller


dhiller (approved):
/approve

PR #5078, opened 2026-05-20
Set bazel remote cache to false

Whitedyl:
This PR is turning off bazel caching, as the ci is experiencing multiple build failures due to bazel caching. We have tried to clear the cache and retest jobs but this has not worked, leading to this change.

Example of bazel caching failures:

- https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/17661/pull-kubevirt-e2e-k8s-1.36-sig-compute/2056739171833221120

/cc @dhiller


Whitedyl:
My only issue is that jobs that have already been triggered have the old config and will continue to send bazel cache to greenhouse. It may only work once the triggered jobs finish.

Whitedyl:
/close 

Bazel caching has stabilized, is not causing builds to fail anymore.

PR #5083, opened 2026-05-21
Update help provider for test-subset

Whitedyl:
What this PR does / why we need it:
Updates the test-subset plugin's HelpProvider to bring the command help page at prow.ci.kubevirt.io/command-help in line with the README. Adds documentation for the --filter, --focus, and --verbosity parameters including the environment variables they map to, clarifies that jobs run as a separate test-subset- status context, and corrects the permission requirement to members of the Kubevirt org.

Special notes for your reviewer:
/cc @dhiller


dhiller (approved):
/approve

PR #5085, opened 2026-05-25
Enable reporting for coverage plugin

Whitedyl:
This PR sets skip report to false which enables GitHub status reporting in PR's for coverage plugin. 

Previously, SkipReport was set to true, meaning the coverage job would run but its results were not visible as a status check on PRs. With this change, coverage results will now appear directly in the PR status checks.

/cc @dhiller 

**Checklist**
This change has been tested by triggering the coverage plugin simultaneously against both project-infra and ci-health repositories, and it ran successfully in both.

### DIFF
--- a/external-plugins/coverage/plugin/handler/handler.go
+++ b/external-plugins/coverage/plugin/handler/handler.go
@@ -312,7 +312,7 @@ func (h *GitHubEventsHandler) generateCoverageJob(
 		},
 		Reporter: config.Reporter{
 			Context:    "coverage-auto",
-			SkipReport: true,
+			SkipReport: false,
 		},
 	}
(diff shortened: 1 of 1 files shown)


external-plugins/coverage/plugin/handler/handler.go
```diff
@@ -312,7 +312,7 @@ func (h *GitHubEventsHandler) generateCoverageJob(
 		},
 		Reporter: config.Reporter{
 			Context:    "coverage-auto",
-			SkipReport: true,
```
dhiller, line 315: I would really like to have this as a part of the ConfigMap, but yes, let's do it now :)

dhiller (approved):
/approve

Issue #5073, opened 2026-05-19
Add signal trap handlers to prevent unkillable container processes in Prow jobs


dhiller:
## Context

Follow-up from #5070: 78 Prow job pods were stuck in `Terminating` state (up to 50 days) because their container processes entered kernel D-state (uninterruptible sleep) and could not be killed by SIGTERM or SIGKILL. The root cause is that certain jobs use resources (NBD devices, Bazel sandboxes) that require explicit cleanup before process exit.

## Problem

When the Prow entrypoint timeout fires, it sends SIGTERM then SIGKILL. But if the process is blocked on NBD I/O or inside a Bazel sandbox mount namespace, those signals are ignored by the kernel. The container stays "running" forever, the kubelet can't confirm termination, and the pod is stuck in `Terminating` indefinitely. Sinker cannot clean it up.

### 1. Alpine/containerdisk jobs (`check-provision-alpine-with-test-tooling`, `publish-alpine-test-tooling`)

The `create-containerdisk.sh` script uses `qemu-nbd` to attach NBD devices (`/dev/nbdN`) for disk image creation. When killed mid-operation, the NBD I/O blocks in D-state.

**Fix**: Add a `trap` handler in `create-containerdisk.sh` that disconnects all NBD devices on SIGTERM/SIGINT before exiting:

```bash
cleanup() {
    for nbd in /dev/nbd*; do
        qemu-nbd -d "$nbd" 2>/dev/null || true
    done
}
trap cleanup EXIT SIGTERM SIGINT
```

Also consider increasing the entrypoint grace period from 15s to give the cleanup trap time to run (e.g., 30-60s).

### 2. Bazel generate jobs (`pull-kubevirt-generate-*`)

Bazel's `processwrapper-sandbox` creates nested mount namespaces. Processes inside these sandboxes that are mid-syscall resist SIGKILL.

**Fix**: Add a `trap` handler in the job script that shuts down Bazel before exiting:

```bash
cleanup() {
    bazel shutdown 2>/dev/null || true
}
trap cleanup EXIT SIGTERM SIGINT
```

### 3. General: consider a periodic cleanup CronJob

As a safety net, consider a CronJob that force-deletes pods stuck in `Terminating` for more than N hours:

```bash
kubectl get pods -n kubevirt-prow-jobs --field-selector=status.phase=Running \
  -o json | jq -r '.items[] | select(.metadata.deletionTimestamp) | .metadata.name' | \
[...]

Issue #5070, opened 2026-05-19
Prow job pods stuck in Terminating state due to unkillable processes (NBD/Bazel D-state)


dhiller:
## Summary

There are currently **78 pods** stuck in `Terminating` state in the `kubevirt-prow-jobs` namespace, spanning 2 to 50 days. Prow's sinker cannot remove them because the kubelet is unable to confirm container termination — the underlying processes are stuck in kernel D-state (uninterruptible sleep) and cannot be killed by any signal, including SIGKILL.

## Affected Jobs

| Job | Count | % | Image |
|---|---|---|---|
| `check-provision-alpine-with-test-tooling` | 48 | 62% | `quay.io/kubevirtci/golang` |
| `publish-alpine-test-tooling` | 14 | 18% | `quay.io/kubevirtci/golang` |
| `pull-kubevirt-generate-1.8` | 12 | 15% | `quay.io/kubevirtci/bootstrap` |
| `pull-kubevirt-generate-1.7` | 3 | 4% | `quay.io/kubevirtci/bootstrap` |
| `pull-kubevirt-generate-1.6` | 1 | 1% | `quay.io/kubevirtci/bootstrap` |

## Common Properties

All 78 pods share these traits:

- **SCC**: `insights-runtime-extractor-scc` (privileged)
- **Container pattern**: `sidecar` exited cleanly (exit 0), `test` container still shows `state: running`
- **Last log line**: `"Process did not exit before N grace period"` followed by `"process timed out"`
- **Node spread**: distributed across 10 worker nodes (bm02–bm13), not isolated to a single node
- **Age**: oldest stuck since 2026-03-29, newest since 2026-05-18

### Alpine/containerdisk jobs (80% of stuck pods)

The `check-provision-alpine-with-test-tooling` and `publish-alpine-test-tooling` jobs run:

```bash
dnf install -y genisoimage libvirt && \
  cd cluster-provision/images/vm-image-builder && \
  ./create-containerdisk.sh alpine-cloud-init
```

This script uses **`qemu-nbd`** to attach Network Block Devices (`/dev/nbdN`), then performs partitioning and filesystem operations on them. When the Prow entrypoint timeout fires (20m/60m) and sends SIGTERM → SIGKILL, any process blocked on NBD I/O is in D-state and **cannot be killed by any signal**. The kernel will not reap the process until the I/O completes, which never happens if the NBD session is broken.

[...]

PR #5091, opened 2026-05-26
Add weekly CronJob to clean up pods stuck in Terminating state

Whitedyl:
Adds a weekly CronJob (`terminating-pod-cleanup`) that force-deletes pods stuck in Terminating state across all federated build clusters.

Follow-up from #5070: 78 Prow job pods were stuck in Terminating (up to 50 days) because their container processes entered kernel D-state from NBD device I/O or Bazel sandbox mount namespaces. In D-state, processes ignore SIGTERM and SIGKILL, so the kubelet can never confirm termination and sinker cannot clean them up.

The CronJob runs every Monday at 06:30 UTC, iterates over all cluster contexts from the consolidated kubeconfig, and force-deletes any Terminating pods in the `kubevirt-prow-jobs` namespace.

**Which issue(s) this PR fixes**
Partly addresses #5073 

/cc @dhiller 

## Summary by CodeRabbit

* **New Features**
  * Added an automated cleanup job for long-terminating pods in the cluster.
  * Included the new cleanup job in the deployed Kubernetes manifests, so it runs as part of the environment setup.

### DIFF
--- a/github/ci/prow-deploy/kustom/base/kustomization.yaml
+++ b/github/ci/prow-deploy/kustom/base/kustomization.yaml
@@ -68,6 +68,7 @@ resources:
   - manifests/local/prow-coverage-service.yaml
   - manifests/local/release-blocker_deployment.yaml
   - manifests/local/release-blocker_service.yaml
+  - manifests/local/terminating-pod-cleanup.yaml
 
   - manifests/test_infra/current/cherrypicker_deployment.yaml
   - manifests/test_infra/current/cherrypicker_service.yaml
--- a/github/ci/prow-deploy/kustom/base/manifests/local/terminating-pod-cleanup.yaml
+++ b/github/ci/prow-deploy/kustom/base/manifests/local/terminating-pod-cleanup.yaml
@@ -18,6 +18,7 @@ spec:
             - name: cleanup
               image: quay.io/kubevirtci/bootstrap:v20260417-7427ea2
               command:
+                - /usr/local/bin/runner.sh
                 - /bin/sh
                 - -c
                 - |
@@ -25,9 +26,9 @@ spec:
                   KC="--context=prow-workloads --kubeconfig=/etc/kubeconfig/config --request-timeout=30s"
                   CUTOFF=$(date -u -d '2 days ago' +%Y-%m-%dT%H:%M:%SZ)
                   kubectl get pods -n kubevirt-prow-jobs ${KC} \
-                    -o jsonpath='{range .items[?(@.metadata.deletionTimestamp)]}{.metadata.name}{"\t"}{.metadata.deletionTimestamp}{"\n"}{end}' | \
-                    awk -v cutoff="${CUTOFF}" '$2 <= cutoff {print $1}' | \
-                    xargs -r kubectl delete pod -n kubevirt-prow-jobs ${KC} --force --grace-period=0
+                    -o jsonpath='{range .items[?(@.metadata.deletionTimestamp)]}{.metadata.name}{"\t"}{.metadata.deletionTimestamp}{"\n"}{end}' \
+                    | awk -v cutoff="${CUTOFF}" '$2 <= cutoff {print $1}' \
+                    | xargs -r kubectl delete pod -n kubevirt-prow-jobs ${KC} --force --grace-period=0
                   echo "Cleanup complete."
               resources:
                 requests:
--- a/github/ci/prow-deploy/kustom/base/manifests/local/terminating-pod-cleanup.yaml
+++ b/github/ci/prow-deploy/kustom/base/manifests/local/terminating-pod-cleanup.yaml
@@ -21,13 +21,13 @@ spec:
                 - /bin/sh
                 - -c
                 - |
-                  for CONTEXT in $(kubectl config get-contexts -o name --kubeconfig=/etc/kubeconfig/config); do
-                    echo "--- Checking cluster ${CONTEXT} ---"
-                    KC="--context=${CONTEXT} --kubeconfig=/etc/kubeconfig/config"
-                    kubectl get pods -n kubevirt-prow-jobs ${KC} | \
-                      awk '/Terminating/{print $1}' | \
-                      xargs -r kubectl delete pod -n kubevirt-prow-jobs ${KC} --force --grace-period=0
-                  done
+                  set -euo pipefail
+                  KC="--context=prow-workloads --kubeconfig=/etc/kubeconfig/config --request-timeout=30s"
+                  CUTOFF=$(date -u -d '2 days ago' +%Y-%m-%dT%H:%M:%SZ)
+                  kubectl get pods -n kubevirt-prow-jobs ${KC} \
+                    -o jsonpath='{range .items[?(@.metadata.deletionTimestamp)]}{.metadata.name}{"\t"}{.metadata.deletionTimestamp}{"\n"}{end}' | \
+                    awk -v cutoff="${CUTOFF}" '$2 <= cutoff {print $1}' | \
+                    xargs -r kubectl delete pod -n kubevirt-prow-jobs ${KC} --force --grace-period=0
                   echo "Cleanup complete."
               resources:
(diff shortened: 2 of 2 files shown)


github/ci/prow-deploy/kustom/base/manifests/local/terminating-pod-cleanup.yaml
```diff
@@ -0,0 +1,48 @@
...
+      template:
+        spec:
+          containers:
+            - name: cleanup
+              image: quay.io/kubevirtci/bootstrap:v20260417-7427ea2
+              command:
+                - /bin/sh
+                - -c
+                - |
+                  for CONTEXT in $(kubectl config get-contexts -o name --kubeconfig=/etc/kubeconfig/config); do
```
dhiller: not sure whether we should limit this to the prow-workloads cluster at first?

github/ci/prow-deploy/kustom/base/manifests/local/terminating-pod-cleanup.yaml
```diff
@@ -0,0 +1,48 @@
...
+                    echo "--- Checking cluster ${CONTEXT} ---"
+                    KC="--context=${CONTEXT} --kubeconfig=/etc/kubeconfig/config"
+                    kubectl get pods -n kubevirt-prow-jobs ${KC} | \
```
dhiller: I would add a condition here that only kills pods older than two days (or maybe a week), since sinker does only look at pods older than two days IIRC.

dhiller (commented):
Hey @Whitedyl - thank you for the PR!

Some suggestions inline.

github/ci/prow-deploy/kustom/base/manifests/local/terminating-pod-cleanup.yaml
Whitedyl: Originally I used `kubectl config get-contexts` to iterate over all contexts in the consolidated kubeconfig — the same approach sinker uses (`--kubeconfig=/etc/kubeconfig/config`) to operate across all federated clusters. But makes sense to only target workloads cluster.

github/ci/prow-deploy/kustom/base/manifests/local/terminating-pod-cleanup.yaml
```diff
@@ -0,0 +1,48 @@
...
+                      awk '/Terminating/{print $1}' | \
+                      xargs -r kubectl delete pod -n kubevirt-prow-jobs ${KC} --force --grace-period=0
```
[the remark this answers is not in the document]
Whitedyl: Addressing this

dhiller:
@Whitedyl updated the PR description from "Partly fixes #" to "Partly addresses #" as otherwise the issue would get closed when this PR merges.

github/ci/prow-deploy/kustom/base/manifests/local/terminating-pod-cleanup.yaml
```diff
@@ -0,0 +1,48 @@
+---
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: terminating-pod-cleanup
+spec:
+  schedule: "30 6 * * 1"
```
dhiller, line 7: dumb q probably, but why Monday? My thinking is that when it runs on Saturday the impact is lower in case the filtering has errors.

dhiller (commented):
/approve

@dollierp WDYT?

github/ci/prow-deploy/kustom/base/manifests/local/terminating-pod-cleanup.yaml
```diff
@@ -0,0 +1,48 @@
...
+        app: terminating-pod-cleanup
+    spec:
+      activeDeadlineSeconds: 300
+      template:
+        spec:
+          containers:
+            - name: cleanup
+              image: quay.io/kubevirtci/bootstrap:v20260417-7427ea2
+              command:
+                - /bin/sh
```
dollierp, line 22: I think you need to use `/usr/local/bin/runner.sh` to get `kubectl` added to the PATH.

github/ci/prow-deploy/kustom/base/manifests/local/terminating-pod-cleanup.yaml
```diff
@@ -0,0 +1,48 @@
...
+                - -c
+                - |
+                  set -euo pipefail
+                  KC="--context=prow-workloads --kubeconfig=/etc/kubeconfig/config --request-timeout=30s"
+                  CUTOFF=$(date -u -d '2 days ago' +%Y-%m-%dT%H:%M:%SZ)
+                  kubectl get pods -n kubevirt-prow-jobs ${KC} \
+                    -o jsonpath='{range .items[?(@.metadata.deletionTimestamp)]}{.metadata.name}{"\t"}{.metadata.deletionTimestamp}{"\n"}{end}' | \
```
dollierp: Nit: our convention is to put the `|` at the beginning of the next line, not at the end of the previous one.

dollierp:
> @dollierp WDYT?

I have mixed feeling about force removing the Pods if its processes are unkillable and still present on the node.

github/ci/prow-deploy/kustom/base/kustomization.yaml
```diff
@@ -42,6 +42,11 @@ resources:
   - manifests/local/prow-coverage-service.yaml
   - manifests/local/release-blocker_deployment.yaml
   - manifests/local/release-blocker_service.yaml
+<<<<<<< HEAD
```
dhiller: that looks like a rebase gone wrong?

Whitedyl: Yeah it was, I fixed it now. Thanks for highlighting!

github/ci/prow-deploy/kustom/base/manifests/local/terminating-pod-cleanup.yaml
Whitedyl: With the jsonpath change errors should not happen if its using the K8s API, so the day shouldn't matter. If we want it running at an off time I can make that change, WDYT?

dhiller:
> > @dollierp WDYT?
> 
> I have mixed feeling about force removing the Pods if its processes are unkillable and still present on the node.

So IMHO we have two options here:
* create a regular habit of manually cleaning up the nodes, i.e. deleting stale pods, then checking each node for stale processes and reboot them if so (see  https://github.com/kubevirt/project-infra/issues/5070)
* merging this one and add a regular automated restart of each node on the weekend (i.e. Saturday)

WDYT?

Whitedyl:
> > > @dollierp WDYT?
> > 
> > 
> > I have mixed feeling about force removing the Pods if its processes are unkillable and still present on the node.
> 
> So IMHO we have two options here:
> 
> * create a regular habit of manually cleaning up the nodes, i.e. deleting stale pods, then checking each node for stale processes and reboot them if so (see  [Prow job pods stuck in Terminating state due to unkillable processes (NBD/Bazel D-state) #5070](https://github.com/kubevirt/project-infra/issues/5070))
> * merging this one and add a regular automated restart of each node on the weekend (i.e. Saturday)
> 
> WDYT?

Hey @dollierp, any opinions on the above? Thanks

dollierp:
> Hey @dollierp, any opinions on the above? Thanks

I can't make up my mind, so I'll go along with whatever you decide.

dhiller (approved):
Looks good to me.

Whitedyl:
> > Hey @dollierp, any opinions on the above? Thanks
> 
> I can't make up my mind, so I'll go along with whatever you decide.
Thanks @dollierp 

I think we should do it for now but also we all agree that this isn't the full fix, we also need to add the job to restart nodes each week. Together hopefully can help the CI's performance. 

Ultimately these are patches until we investigate and fix the underlying issue further which I think we also all agree on.

dollierp (approved):
/lgtm

PR #5095, opened 2026-05-27
Remove CentOS 9 bumping jobs

Whitedyl:
The new CentOS 9 base image (20260513.0) is causing jobs to fail due to SSH timeouts during cluster-up. The base image was reverted in kubevirt/kubevirtci#1716.

This PR removes the periodic-kubevirtci-bump-centos-base and periodic-kubevirtci-bump-centos-base-s390x jobs to prevent CentOS 9 from being automatically bumped again, keeping CI stable on the known-good image.

These jobs will be re-added once the root cause of the SSH failures in newer CentOS 9 Stream images is resolved.

/cc @dhiller


dhiller (approved):
/approve

@Whitedyl please create a GitHub tracker issue for the efforts of finding the root cause

dhiller:
see tracker https://github.com/kubevirt/project-infra/issues/5098

PR #5109, opened 2026-06-03
fix(grafana-token): fixup early exit condition

dollierp:

Fixup the shell scriptlet to actually execute the early `exit 1` only if the token was not generated properly.

/cc @Whitedyl


Whitedyl (approved):
Thanks @dollierp ! 

/lgtm

PR #5126, opened 2026-06-09
Replace the in cluster greenhouse HTTP cache with a dedicated GCS bucket

Whitedyl:
Migrates the Bazel remote cache from the in-cluster greenhouse HTTP service to a dedicated GCS bucket (kubevirt-bazel-cache). Moving to GCS gives us durable, scaleable object storage as well as reducing poisoned cache failures.

- New bazel-cache-gcs Kubernetes Secret added to all three cluster overlays, sourced from a dedicated service account scoped only to the new bucket.
- New preset-bazel-cache-gcs-credentials preset that mounts the secret and exposes the credential path via BAZEL_CACHE_GOOGLE_CREDENTIALS (intentionally separate from GOOGLE_APPLICATION_CREDENTIALS to avoid conflicts with preset-gcs-credentials).
- preset-bazel-cache updated to point at the GCS bucket URL instead of the greenhouse cluster-local host
- create_bazel_cache_rcs.sh updated to use the GCS URL when BAZEL_REMOTE_CACHE is set, preserving per-workspace/toolchain path isolation within the bucket and authenticating via --google_credentials
- preset-bazel-cache-gcs-credentials added to all jobs that use Bazel caching

/cc @dhiller @dollierp 
Please add a /hold after approval as prow-deploy periodic will need to be manually ran after merging to deploy the SA secret to the cluster. 
Two follow up PR's will be needed:
- Removing greenhouse 
- Updating hack/clear-bazel-cache.sh 

**Checklist**
- [x] Create GSC bucket(kubevirt-bazel-cache) and dedicated service account(key stored in k/secrets).

### DIFF
--- a/github/ci/prow-deploy/kustom/base/configs/current/config/config.yaml
+++ b/github/ci/prow-deploy/kustom/base/configs/current/config/config.yaml
@@ -675,12 +675,24 @@ presets:
   env:
     - name: BAZEL_VERSION
       value: "6.5.0"
-    - name: CACHE_HOST
-      value: bazel-cache.kubevirt-prow
-    - name: CACHE_PORT
-      value: "8080"
+    - name: BAZEL_REMOTE_CACHE
+      value: https://storage.googleapis.com/kubevirt-bazel-cache
     - name: BAZEL_REMOTE_CACHE_ENABLED
       value: "true"
+- labels:
+    preset-bazel-cache-gcs-credentials: "true"
+  env:
+    - name: BAZEL_CACHE_GOOGLE_CREDENTIALS
+      value: /etc/bazel-cache-gcs/bazel-cache-sa.json
+  volumes:
+    - name: bazel-cache-gcs
+      secret:
+        secretName: bazel-cache-gcs
+        defaultMode: 0400
+  volumeMounts:
+    - name: bazel-cache-gcs
+      mountPath: /etc/bazel-cache-gcs
+      readOnly: true
 - labels:
     preset-bazel-unnested: "true"
   env:
--- a/github/ci/prow-deploy/kustom/base/configs/current/config/config.yaml
+++ b/github/ci/prow-deploy/kustom/base/configs/current/config/config.yaml
@@ -679,9 +679,6 @@ presets:
       value: https://storage.googleapis.com/kubevirt-bazel-cache
     - name: BAZEL_REMOTE_CACHE_ENABLED
       value: "true"
-- labels:
-    preset-bazel-cache-gcs-credentials: "true"
-  env:
     - name: BAZEL_CACHE_GOOGLE_CREDENTIALS
       value: /etc/bazel-cache-gcs/bazel-cache-sa.json
   volumes:
--- a/github/ci/prow-deploy/kustom/overlays/kubevirt-prow-control-plane/kustomization.yaml
+++ b/github/ci/prow-deploy/kustom/overlays/kubevirt-prow-control-plane/kustomization.yaml
@@ -280,6 +280,11 @@ secretGenerator:
   - <<: *gcs_credentials
     name: gcs
     namespace: kubevirt-prow-jobs
+  - name: bazel-cache-gcs
+    namespace: kubevirt-prow-jobs
+    files:
+      - secrets/prow/google/service-accounts/bazel-cache/bazel-cache-sa.json
+    type: Opaque
   - name: virtci-gcs
     namespace: kubevirt-prow-jobs
     files:
--- a/images/bootstrap/create_bazel_cache_rcs.sh
+++ b/images/bootstrap/create_bazel_cache_rcs.sh
@@ -75,14 +75,20 @@ make_bazel_rc () {
     echo "startup --host_jvm_args=-Dbazel.DigestFunction=sha256"
     # don't fail if the cache is unavailable
     echo "build --remote_local_fallback"
-    # point bazel at our http cache ...
-    # NOTE our caches are versioned by all path segments up until the last two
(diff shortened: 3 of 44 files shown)


github/ci/prow-deploy/kustom/base/configs/current/config/config.yaml
```diff
@@ -673,12 +673,24 @@ presets:
...
       value: "6.5.0"
-    - name: CACHE_HOST
-      value: bazel-cache.kubevirt-prow
-    - name: CACHE_PORT
-      value: "8080"
+    - name: BAZEL_REMOTE_CACHE
+      value: https://storage.googleapis.com/kubevirt-bazel-cache
     - name: BAZEL_REMOTE_CACHE_ENABLED
       value: "true"
+- labels:
```
dollierp: I would have added this to the existing `preset-bazel-cache` preset.

github/ci/prow-deploy/kustom/base/configs/current/config/config.yaml
```diff
@@ -673,12 +673,24 @@ presets:
   env:
     - name: BAZEL_VERSION
       value: "6.5.0"
-    - name: CACHE_HOST
-      value: bazel-cache.kubevirt-prow
-    - name: CACHE_PORT
-      value: "8080"
+    - name: BAZEL_REMOTE_CACHE
+      value: https://storage.googleapis.com/kubevirt-bazel-cache
```
dollierp, line 679: Did you add a retention policy to this new bucket so that it won't grow forever and exhaust our budget?

github/ci/prow-deploy/kustom/overlays/kubevirt-prow-control-plane/kustomization.yaml
```diff
@@ -286,6 +286,11 @@ secretGenerator:
   - <<: *gcs_credentials
     name: gcs
     namespace: kubevirt-prow-jobs
+  - name: bazel-cache-gcs
+    namespace: kubevirt-prow-jobs
+    files:
+      - secrets/prow/google/service-accounts/bazel-cache/bazel-cache-sa.json
```
dollierp, line 286: The checkout of the secrets repository is not directly available to kustomize like this.

See the `populate_secrets()` function in https://github.com/kubevirt/project-infra/blob/main/github/ci/prow-deploy/hack/deploy.sh and https://github.com/kubevirt/project-infra/blob/main/github/ci/prow-deploy/hack/test.sh.

github/ci/prow-deploy/kustom/base/configs/current/config/config.yaml
dollierp: Replying to myself, there is a lifecycle policy to remove objects older than 30days.

github/ci/prow-deploy/kustom/base/configs/current/config/config.yaml
Whitedyl: Yeah that makes sense, I was worried about a collision with `preset-gcs-credentials` so I made a new label but looking at it again it's better to put it there and I can remove the extra label from each job.

github/ci/prow-deploy/kustom/overlays/kubevirt-prow-control-plane/kustomization.yaml
Whitedyl: Thanks for catching this, I have updated it now

Whitedyl:
@dollierp I've made the requested changes can you take a look? 
Thanks!

images/bootstrap/create_bazel_cache_rcs.sh
```diff
@@ -75,14 +75,22 @@ make_bazel_rc () {
...
     echo "build --remote_local_fallback"
-    # point bazel at our http cache ...
-    # NOTE our caches are versioned by all path segments up until the last two
-    # IE PUT /foo/bar/baz/cas/asdf -> is in cache "/foo/bar/baz"
-    local cache_id
-    cache_id="$(get_workspace),$(hash_toolchains)"
-    local cache_url
-    cache_url="http://${CACHE_HOST}:${CACHE_PORT}/${cache_id}"
-    echo "build --remote_cache=${cache_url}"
+    if [[ -n "${BAZEL_REMOTE_CACHE:-}" ]]; then
```
dollierp: I have the feeling this can be improved

```shell
# NOTE our caches are versioned by all path segments up until the last two
# IE PUT /foo/bar/baz/cas/asdf -> is in cache "/foo/bar/baz"
local cache_id
cache_id="$(get_workspace),$(hash_toolchains)"
local cache_url
if [[ -n "${BAZEL_REMOTE_CACHE-}" ]]; then
    cache_url=${BAZEL_REMOTE_CACHE}/${cache_id}
else
    cache_url="http://${CACHE_HOST}:${CACHE_PORT}/${cache_id}"
fi
echo "build --remote_cache=${cache_url}"

if [[ -n "${BAZEL_CACHE_GOOGLE_CREDENTIALS-}" ]]; then
    echo "build --google_credentials=${BAZEL_CACHE_GOOGLE_CREDENTIALS}"
fi
```

Whitedyl: Updated this now!

dollierp (approved):
/approve

dhiller:
@Whitedyl it seems you need to resolve merge conflicts here first

Whitedyl:
> @Whitedyl it seems you need to resolve merge conflicts here first

Yeah there were additions to deploy.sh and test.sh, I have resolved them now.

dhiller (approved):

PR #5144, opened 2026-06-13
fix(bazel): switch back to greenhouse cache

dollierp:

Jobs are currently failing to use the GCS cache:

```
ERROR: Could not open auth credentials file '/etc/bazel-cache-gcs/bazel-cache-sa.json': /etc/bazel-cache-gcs/bazel-cache-sa.json (No such file or directory)
ERROR: Error initializing RemoteModule
```

It seems the automation needs to be improved to forward the file `/etc/bazel-cache-gcs/bazel-cache-sa.json` to the bazel container.

Switch back to greenhouse cache in the meantime.

/cc @dhiller @Whitedyl


dollierp (commented):
/approve
/lgtm

PR #5146, opened 2026-06-14
fix(ci-search): set Jira endpoint which is now required

dollierp:

Add `--jira-endpoint` command-line parameter to ci-search container.

With the latest ci-search image, the StatefulSet won't start if the argument is not set:

```
Unable to configure jira: unable to configure jira: empty --jira-endpoint, can not create a client
```

/cc @dhiller


dhiller (approved):
/approve

PR #5097, opened 2026-05-28
Make check-provision-k8s-1.34 and check-provision-k8s-1.34-s390x jobs optional

Whitedyl:
**What this PR does / why we need it**:
This PR is making both check-provision-k8s-1.34 and check-provision-k8s-1.34-s390x jobs optional as they are failing at a very high rate, and only passing with false positives as the tests are skipped.

Link to failing tests:
https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirtci/1714/check-provision-k8s-1.34/2059970746565791744

Link to false positive:
https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirtci/1706/check-provision-k8s-1.34/2059959188347424768

**Special notes for your reviewer**:
/cc @dhiller

PR #5174, opened 2026-06-22
feat(robots): add generic --extra-archs flag for multi-arch presubmit automation

carterpewpew:

Reworks the kubevirtci-presubmit-creator and kubevirtci-presubmit-remover robots to support optional extra architectures via a generic `--extra-archs` flag, keeping the default amd64 behavior completely unchanged.

When `--extra-archs` is empty (default), behavior is identical to upstream. When set (e.g. `--extra-archs=s390x`), the creator generates arch-specific presubmit jobs alongside the amd64 job, and the remover deletes both amd64 and arch jobs for retired k8s versions.

s390x check-provision jobs were previously added manually and never cleaned up, leaving stale entries (1.32-s390x, 1.33-s390x) while their amd64 counterparts were already removed by the robot. This PR automates the lifecycle so no manual YAML edits are needed going forward.

**Changes:**

- **Creator robot**: Added `ArchConfig` struct and `knownArchConfigs` map for arch-specific job parameters. New `CreatePresubmitJobForReleaseArch()` function builds arch jobs. `AddNewPresubmitIfNotExists()` now accepts `extraArchs` and creates arch jobs alongside amd64.
- **Remover robot**: Added `--extra-archs` flag. `deletePresubmitJobForRelease()` now includes arch job names in the deletion set.
- **YAML cleanup**: Removed stale 1.32-s390x and 1.33-s390x. 1.34-s390x changed from `always_run: true` to `false` (aligning with ArchConfig and reducing wasted CI on a lane already marked optional per #5097). Added 1.35-s390x to backfill the gap. 1.36-s390x will be created automatically by the robot on its next periodic run.
- **Periodic config**: Both creator and remover periodic jobs now pass `--extra-archs=s390x`.
- **Tests**: Added unit tests for arch-specific job creation, existence checks for mixed amd64+arch scenarios, and combined deletion.

**Adding a new architecture in the future** requires only two changes:
1. Add an entry to `knownArchConfigs` in the creator (cluster, timeout, labels, env vars)
2. Update `--extra-archs` in the periodic job config (e.g. `--extra-archs=s390x,ppc64le`)

**Special notes for reviewer**:

The s390x ArchConfig intentionally differs from amd64 (matching existing manually-created s390x jobs):
- `cluster: prow-s390x-workloads` (not `prow-workloads`)
- `timeout: 4h` (not 3h)
- `always_run: false`, `optional: true` (manual trigger only, non-blocking)
- No `preset-docker-mirror-proxy` label
- No `NodeSelector` (s390x cluster doesn't use bare-metal-external)
- Extra env vars: `SLIM=true`, `RUN_KUBEVIRT_CONFORMANCE=false`

**jira-ticket**:

**Checklist**

- [ ] Design: A [design document](https://github.com/kubevirt/community/tree/main/design-proposals) was considered and is present (link) or not required
- [X] PR: The PR description is expressive enough and will help future contributors
- [X] Code: [Write code that humans can understand](https://en.wikiquote.org/wiki/Martin_Fowler#code-for-humans) and [Keep it simple](https://en.wikipedia.org/wiki/KISS_principle)
- [X] Refactor: You have [left the code cleaner than you found it (Boy Scout Rule)](https://learning.oreilly.com/library/view/97-things-every/9780596809515/ch08.html)
- [X] Upgrade: Impact of this change on upgrade flows was considered and addressed if required
- [X] Testing: New code requires [new unit tests](https://github.com/kubevirt/kubevirt/blob/main/docs/reviewer-guide.md#when-is-a-pr-good-enough).
- [ ] Documentation: A [user-guide update](https://github.com/kubevirt/user-guide/) was considered and is present (link) or not required.
- [ ] Community: Announcement to [kubevirt-dev](https://groups.google.com/g/kubevirt-dev/) was considered

**Release note**:
```release-note
NONE
```

## Summary by CodeRabbit

- **New Features**
  - Added support for creating and removing architecture-specific presubmit jobs via an `--extra-archs` option (including s390x variants).
  - Updated periodic presubmit maintenance to include s390x job generation/removal.

- **Updates**
  - Advanced s390x Kubernetes provisioning checks to Kubernetes 1.34 and 1.35.
  - Updated generated presubmit job execution command and environment setup to the standardized entrypoint flow.

- **Tests**
  - Expanded coverage for architecture-specific presubmit creation/detection and deletion behavior.

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/kubevirtci/kubevirtci-presubmits.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/kubevirtci/kubevirtci-presubmits.yaml
@@ -494,67 +494,6 @@ presubmits:
       preset-kubevirtci-check-provision-env: "true"
       preset-podman-in-container-enabled: "true"
     max_concurrency: 3
-    name: check-provision-k8s-1.32-s390x
-    optional: true
-    spec:
-      containers:
-      - command:
-        - /usr/local/bin/entrypoint.sh
-        - /bin/sh
-        - -c
-        - cd cluster-provision/k8s/1.32 && ../provision.sh
-        env:
-        - name: GO_MOD_PATH
-          value: cluster-provision/gocli/go.mod
-        - name: SLIM
-          value: "true"
-        - name: RUN_KUBEVIRT_CONFORMANCE
-          value: "false"
-        image: quay.io/kubevirtci/golang:v20260715-af3e234
-        resources:
-          requests:
-            memory: 16Gi
-        securityContext:
-          privileged: true
-  - always_run: false
-    cluster: prow-s390x-workloads
-    decoration_config:
-      timeout: 4h0m0s
-    labels:
-      preset-kubevirtci-check-provision-env: "true"
-      preset-podman-in-container-enabled: "true"
-    max_concurrency: 3
-    name: check-provision-k8s-1.33-s390x
-    optional: true
-    spec:
-      containers:
-      - command:
-        - /usr/local/bin/entrypoint.sh
-        - /bin/sh
-        - -c
-        - cd cluster-provision/k8s/1.33 && ../provision.sh
-        env:
-        - name: GO_MOD_PATH
-          value: cluster-provision/gocli/go.mod
-        - name: SLIM
-          value: "true"
-        - name: RUN_KUBEVIRT_CONFORMANCE
-          value: "false"
-        image: quay.io/kubevirtci/golang:v20260715-af3e234
-        resources:
-          requests:
-            memory: 16Gi
-        securityContext:
-          privileged: true
-  # TODO: revert once the prow-s390x-workloads cluster is healthy
-  - always_run: false
-    cluster: prow-s390x-workloads
-    decoration_config:
-      timeout: 4h0m0s
-    labels:
-      preset-kubevirtci-check-provision-env: "true"
-      preset-podman-in-container-enabled: "true"
-    max_concurrency: 3
     name: check-provision-k8s-1.34-s390x
     optional: true
     spec:
@@ -564,6 +503,36 @@ presubmits:
         - /bin/sh
         - -c
         - cd cluster-provision/k8s/1.34 && ../provision.sh
+        env:
+        - name: GO_MOD_PATH
+          value: cluster-provision/gocli/go.mod
+        - name: SLIM
+          value: "true"
+        - name: RUN_KUBEVIRT_CONFORMANCE
+          value: "false"
+        image: quay.io/kubevirtci/golang:v20260715-af3e234
+        resources:
+          requests:
+            memory: 16Gi
+        securityContext:
+          privileged: true
+  - always_run: false
+    cluster: prow-s390x-workloads
+    decoration_config:
+      timeout: 4h0m0s
+    labels:
+      preset-kubevirtci-check-provision-env: "true"
+      preset-podman-in-container-enabled: "true"
+    max_concurrency: 3
+    name: check-provision-k8s-1.35-s390x
+    optional: true
+    spec:
+      containers:
+      - command:
+        - /usr/local/bin/entrypoint.sh
+        - /bin/sh
+        - -c
+        - cd cluster-provision/k8s/1.35 && ../provision.sh
         env:
         - name: GO_MOD_PATH
           value: cluster-provision/gocli/go.mod
--- a/github/ci/prow-deploy/files/jobs/kubevirt/kubevirtci/kubevirtci-presubmits.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/kubevirtci/kubevirtci-presubmits.yaml
@@ -443,7 +443,7 @@ presubmits:
             memory: 1Gi
         securityContext:
           privileged: true
-  - always_run: true
+  - always_run: false
     cluster: prow-s390x-workloads
     decoration_config:
       timeout: 4h0m0s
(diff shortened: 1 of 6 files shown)


carterpewpew:
cc @chandramerla

dhiller (commented):
👋 Hey @carterpewpew - first of all thank you for your contribution! 

As you might have seen in our [CI overview](https://github.com/kubevirt/project-infra/blob/main/docs/infrastructure-components.md) we do also have other architectures besides s390x.

However the automation here is trying to solve a different usecase - namely updating the specific set of **default providers for Kubernetes**.

So I would not want to update the current status quo, but I'd be open to extend it with optional arguments to cover the case you are looking at and also creating additional jobs for them.

carterpewpew:
> 👋 Hey @carterpewpew - first of all thank you for your contribution!
> 
> As you might have seen in our [CI overview](https://github.com/kubevirt/project-infra/blob/main/docs/infrastructure-components.md) we do also have other architectures besides s390x.
> 
> However the automation here is trying to solve a different usecase - namely updating the specific set of **default providers for Kubernetes**.
> 
> So I would not want to update the current status quo, but I'd be open to extend it with optional arguments to cover the case you are looking at and also creating additional jobs for them.

Thanks for the review and the feedback @dhiller! That makes sense, I agree we should keep this generic rather than hardcoding s390x. I'll rework the approach to use optional arguments (e.g. --extra-archs) so the robots can handle any additional architecture. Will update the PR once that's ready.

github/ci/prow-deploy/files/jobs/kubevirt/kubevirtci/kubevirtci-presubmits.yaml
```diff
@@ -442,15 +442,15 @@ presubmits:
...
+    name: check-provision-k8s-1.34-s390x
     optional: true
     spec:
       containers:
       - command:
         - /usr/local/bin/runner.sh
         - /bin/sh
         - -c
-        - cd cluster-provision/k8s/1.32 && ../provision.sh
+        - cd cluster-provision/k8s/1.34 && ../provision.sh
```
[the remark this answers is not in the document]
carterpewpew: This is intentional. The ArchConfig uses AlwaysRun: false, not true. The previous review was against the old hardcoded approach which had AlwaysRun=true. The reworked version deliberately uses false because these lanes were already marked optional in #5097 due to high failure rates. always_run: false + optional: true means the lane exists but must be triggered manually via /test. This is consistent with what the robot will produce for future versions (1.36-s390x etc).

robots/kubevirtci-presubmit-creator/main.go
```diff
@@ -156,21 +220,33 @@ func main() {
 	}
 }
 
-func AddNewPresubmitIfNotExists(jobConfig config.JobConfig, latestReleaseSemver *querier.SemVer) (newJobConfig config.JobConfig, jobExists bool) {
+func AddNewPresubmitIfNotExists(jobConfig config.JobConfig, latestReleaseSemver *querier.SemVer, extraArchs []string) (newJobConfig config.JobConfig, jobExists bool) {
```
dhiller: nit: rename named return variable
```suggestion
func AddNewPresubmitIfNotExists(jobConfig config.JobConfig, latestReleaseSemver *querier.SemVer, extraArchs []string) (newJobConfig config.JobConfig, allExist bool) {
```

robots/kubevirtci-presubmit-creator/main.go
```diff
@@ -131,7 +193,9 @@ func main() {
 		log.Panicln(err)
 	}
 
-	newJobConfig, exists := AddNewPresubmitIfNotExists(jobConfig, latestReleaseSemver)
+	extraArchs := parseExtraArchs(o.extraArchs)
+
+	newJobConfig, exists := AddNewPresubmitIfNotExists(jobConfig, latestReleaseSemver, extraArchs)
```
dhiller: nit: rename
```suggestion
	newJobConfig, allExist := AddNewPresubmitIfNotExists(jobConfig, latestReleaseSemver, extraArchs)
```

dhiller (commented):
/approve

@dollierp mind having a quick look here? Thanks!

robots/kubevirtci-presubmit-creator/main.go
```diff
@@ -230,6 +306,61 @@ func CreatePresubmitJobForRelease(semver *querier.SemVer) config.Presubmit {
 	return res
 }
 
+func CreatePresubmitJobForReleaseArch(semver *querier.SemVer, arch string, archCfg ArchConfig) config.Presubmit {
+	yes := true
+	golangImage := "quay.io/kubevirtci/golang:v20260319-c8f1db8"
+
+	envVars := []v1.EnvVar{
+		{Name: "GIMME_GO_VERSION", Value: "1.24.7"},
```
dollierp: It might be a good opportunity to switch to `{Name: "GO_MOD_PATH", Value: "go.mod"}`.

dhiller:
The golang image tag is hardcoded here in `CreatePresubmitJobForReleaseArch`:

https://github.com/kubevirt/project-infra/pull/5174/files#diff-6e6b4c7e02ee9f0c4e8c2b91d54c8e8fc4f01a5079ebf6407c04b7c2f24fc982R308

The daily `periodic-project-infra-image-bump` job only bumps image tags in YAML files — it won't update this Go source string. Newly created arch presubmit jobs will be stamped with a potentially stale image tag until the bumper catches the generated YAML.

The existing `CreatePresubmitJobForRelease` has the same issue. Tracked in #5215.

🤖 Assisted by Claude

robots/kubevirtci-presubmit-creator/main.go
carterpewpew: Thanks for the suggestion @dollierp!

robots/kubevirtci-presubmit-creator/main.go
```diff
@@ -208,12 +284,63 @@ func CreatePresubmitJobForRelease(semver *querier.SemVer) config.Presubmit {
...
+			MaxConcurrency: 3,
+			Labels:         archCfg.Labels,
+			Cluster:        archCfg.Cluster,
+			Spec: &v1.PodSpec{
+				NodeSelector: archCfg.NodeSelector,
+				Containers: []v1.Container{
+					{
+						Image: golangImage,
+						Command: []string{
+							"/usr/local/bin/runner.sh",
```
dollierp: [`/usr/local/bin/entrypoint.sh`](https://github.com/kubevirt/project-infra/blob/main/images/golang/entrypoint.sh) must be used to support `GO_MOD_PATH`

carterpewpew: Good catch, updated both functions to use `/usr/local/bin/entrypoint.sh` instead of `runner.sh`.

PR #5192, opened 2026-06-24
ci: add test lane for alma linux

codingben:

Add a test lane to run tests on Alma Linux.

**Release note**:

```release-note
None
```

## Summary by CodeRabbit

* **New Features**
  * Added two new presubmit CI jobs for AlmaLinux 9 and 10 in the common-instancetypes test suite.
  * These jobs run only when relevant preference or functest files change, helping target validation more precisely.
  * Both jobs use the existing Kubernetes CI environment and dedicated test focus settings for each AlmaLinux version.


dhiller (commented):
/approve

Looks good to me, however @dollierp do we still use the nodeSelector for bare-metal-external ?

dollierp:
> do we still use the nodeSelector for bare-metal-external ?

It is still present for historical reason on the nodes and the jobs but it can be safely removed.

0xFelix (approved):

PR #5214, opened 2026-07-01
[v1.9] End of Code Freeze 

fossedihelm:
[v1.9] End of Code Freeze

### DIFF
--- a/github/ci/prow-deploy/kustom/base/configs/current/config/config.yaml
+++ b/github/ci/prow-deploy/kustom/base/configs/current/config/config.yaml
@@ -338,8 +338,9 @@ tide:
     - "dco-signoff: no"
   - repos:
     - kubevirt/kubevirt
-    excludedBranches:
-    - "main"
+    # Handling CODE FREEZE
+    # excludedBranches:
+    # - "main"
     labels:
     - lgtm
     - approved


fossedihelm:
/hold
This will be removed once the release branch has been created.

github/ci/prow-deploy/kustom/base/configs/current/config/config.yaml
```diff
@@ -338,8 +338,9 @@ tide:
     - "dco-signoff: no"
   - repos:
     - kubevirt/kubevirt
-    excludedBranches:
```
xpivarc, line 341: you need to also drop what was added

dhiller: At first I thought we must also remove the two queries spanning ll 304-338, since otherwise the approved-vep or kind/bug will be the only ones - now I think this is the right way, since those will be two separate queries that will give PRs matching those a separate tide query, effectively fast-laning those. So great idea, @fossedihelm  !

However we should consider to do that for the other combinations or at least for `kind/flake` + `priority/critical-urgent` as well - probably as a follow up.

@xpivarc @fossedihelm WDYT?

github/ci/prow-deploy/kustom/base/configs/current/config/config.yaml
```diff
@@ -338,8 +338,9 @@ tide:
     - "dco-signoff: no"
```
dhiller, line 338: This will effectively fast lane the approved-vep and kind/bug on main branch ❤️

github/ci/prow-deploy/kustom/base/configs/current/config/config.yaml
fossedihelm: Yes, the main idea was that. I mean, we don't need to drop the others since they will work in addition.
@dhiller Agree, for a follow-up

dhiller (approved):
/approve

/hold to be cancelled for code thaw

fossedihelm:
release-1.9 has been cut.
I'm in favor of merging this.
@dhiller @xpivarc ?

PR #5242, opened 2026-07-06
coverage: add coverage delta and configurable threshold

Whitedyl:
Adds coverage delta computation and a configurable per-repo coverage threshold to the coverage external plugin. When a PR with Go file changes is opened, the coverage job now compares PR coverage against the base branch and posts the result as a GitHub commit status (see below`). If coverage drops below the configured threshold, the job fails.

Previously the plugin only ran coverage on the PR code with no baseline comparison and no minimum enforcement.

The coverage threshold is configurable per repo in the ConfigMap (`coverageThreshold`, default 70%). Repos can override it to match their current baseline.

/cc @dhiller @dollierp 
I've created a rehearsal in the git status to show what it will look like. 

## Summary by CodeRabbit

* **New Features**
  * Added configurable coverage acceptance thresholds with sensible defaults.
  * Added support for selecting the GitHub token secret used by coverage reporting.
  * Coverage reporting now produces richer artifacts (including a coverage summary) and automatically posts GitHub commit statuses (pending/success/failure/error).
* **Bug Fixes**
  * Improved validation for coverage threshold values (including per-repo overrides) to prevent invalid settings.
  * Fixed repo-level configuration inheritance/override so defaults and overrides behave consistently.
  * Updated the coverage job to mount and use the configured GitHub token secret.

### DIFF
--- a/external-plugins/coverage/plugin/handler/handler.go
+++ b/external-plugins/coverage/plugin/handler/handler.go
@@ -45,6 +45,8 @@ type JobConfig struct {
 	GracePeriodSeconds int                 `yaml:"gracePeriodSeconds"`
 	UtilityImages      UtilityImagesConfig `yaml:"utilityImages"`
 	GCS                GCSConfig           `yaml:"gcs"`
+	CoverageThreshold  int                 `yaml:"coverageThreshold"`
+	GitHubTokenSecret  string              `yaml:"githubTokenSecret"`
 }
 
 type Config struct {
@@ -84,6 +86,12 @@ func (c *Config) RepoConfig(repo string) (*JobConfig, bool) {
 	}
 	if repoCfg.GCS != (GCSConfig{}) {
 		merged.GCS = repoCfg.GCS
+	}
+	if repoCfg.CoverageThreshold != 0 {
+		merged.CoverageThreshold = repoCfg.CoverageThreshold
+	}
+	if repoCfg.GitHubTokenSecret != "" {
+		merged.GitHubTokenSecret = repoCfg.GitHubTokenSecret
 	}
 	return &merged, true
 }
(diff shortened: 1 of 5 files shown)


images/covreport/Containerfile
```diff
@@ -1,3 +1,5 @@
-FROM quay.io/kubevirtci/golang:v20260703-61a4923
+FROM quay.io/kubevirtci/golang:v20260612-73bc71e
 RUN /usr/local/bin/entrypoint.sh /bin/sh -ce "GOBIN=/usr/local/bin go install github.com/cancue/covreport@v0.5.0"
+COPY coverage-report.sh /usr/local/bin/coverage-report.sh
+RUN chmod +x /usr/local/bin/coverage-report.sh
```
dollierp: You can use `COPY --chmod=`.

external-plugins/coverage/plugin/handler/handler.go
```diff
@@ -141,6 +149,15 @@ func LoadConfig(path string) (*Config, error) {
 	if cfg.Defaults.GracePeriodSeconds == 0 {
 		cfg.Defaults.GracePeriodSeconds = 15
 	}
+	if cfg.Defaults.CoverageThreshold < 0 || cfg.Defaults.CoverageThreshold > 100 {
+		return nil, fmt.Errorf("config: defaults.coverageThreshold must be between 0 and 100")
+	}
+	if cfg.Defaults.CoverageThreshold == 0 {
+		cfg.Defaults.CoverageThreshold = 70
```
dollierp: It might be interesting to define constants for fallback values.

images/covreport/coverage-report.sh
```diff
@@ -0,0 +1,137 @@
...
+    echo "${SPYGLASS_BASE}/pr-logs/pull/${REPO_OWNER}_${REPO_NAME}/${PULL_NUMBER}/${JOB_NAME}/${BUILD_ID}/"
+}
+
+post_github_status() {
+    local state="$1"
+    local description="$2"
+    local target_url="${3:-}"
+
+    if [[ ! -f "${GITHUB_TOKEN_FILE}" ]]; then
+        echo "WARNING: GitHub token not found, skipping status update"
```
dollierp: It would be great to send log/error messages to stderr.

external-plugins/coverage/plugin/handler/handler.go
Whitedyl: Updated now.

images/covreport/coverage-report.sh
Whitedyl: Updated now.

images/covreport/Containerfile
Whitedyl: Updated now.

images/covreport/coverage-report.sh
dhiller, line 1: nit: it would be great if this would be a go binary, so we could unit test it. But maybe we take that in and do that as a follow up. WDYT @Whitedyl ?

dhiller (approved):
/approve

One comment inside.

/hold in case you want to handle the nit right away.

images/covreport/coverage-report.sh
Whitedyl: Yeah I agree and I think it should be done as a follow up PR. Will revisit when I return! 

/unhold

PR #5227, opened 2026-07-02
kubevirtci: update jobs to use GO_MOD_PATH to allow gocli bump

brianmcarey:
**What this PR does / why we need it**:

In order to properly test updating gocli to go 1.26[1] - the CI jobs need to be updated to use GO_MOD_PATH

[1] https://github.com/kubevirt/kubevirtci/pull/1754

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

/cc @dhiller @Whitedyl 

**Checklist**

This checklist is not enforcing, but it's a reminder of items that could be relevant to every PR.
Approvers are expected to review this list.

- [ ] Design: A [design document](https://github.com/kubevirt/community/tree/main/design-proposals) was considered and is present (link) or not required
- [ ] PR: The PR description is expressive enough and will help future contributors
- [ ] Code: [Write code that humans can understand](https://en.wikiquote.org/wiki/Martin_Fowler#code-for-humans) and [Keep it simple](https://en.wikipedia.org/wiki/KISS_principle)
- [ ] Refactor: You have [left the code cleaner than you found it (Boy Scout Rule)](https://learning.oreilly.com/library/view/97-things-every/9780596809515/ch08.html)
- [ ] Upgrade: Impact of this change on upgrade flows was considered and addressed if required
- [ ] Testing: New code requires [new unit tests](https://github.com/kubevirt/kubevirt/blob/main/docs/reviewer-guide.md#when-is-a-pr-good-enough). New features and bug fixes require at least on e2e test
- [ ] Documentation: A [user-guide update](https://github.com/kubevirt/user-guide/) was considered and is present (link) or not required. You want a user-guide update if it's a user facing feature / API change.
- [ ] Community: Announcement to [kubevirt-dev](https://groups.google.com/g/kubevirt-dev/) was considered

**Release note**:

```release-note

```

## Summary by CodeRabbit

* **Chores**
  * Updated several CI jobs to use module-based Go configuration for build and test execution.
  * Applied the same environment setting changes across presubmit, postsubmit, and periodic workflows for consistent behavior.

PR #5248, opened 2026-07-07
kubevirtci: fix GO_MOD_PATH by using entrypoint.sh and correcting path

Whitedyl:
Fixes two bugs introduced by #5227 

- runner.sh - entrypoint.sh: All kubevirtci golang-image jobs were still using runner.sh, which does not read GO_MOD_PATH. Only entrypoint.sh (provided by the golang image) reads it, extracts the Go version,
sets GIMME_GO_VERSION, and then delegates to runner.sh.
- go.mod - cluster-provision/gocli/go.mod: GO_MOD_PATH was set to go.mod but kubevirtci has no go.mod at the repo root.

Jobs were passing because the golang image bakes in GIMME_GO_VERSION=1.26.4.

/cc @dhiller @dollierp @brianmcarey 

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated CI jobs to use the correct startup behavior and module location, helping presubmit, periodic, and publish runs execute more reliably.
  * Applied the same fix across multiple job types so checks and release-related runs stay consistent.

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/kubevirtci/kubevirtci-periodics.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/kubevirtci/kubevirtci-periodics.yaml
@@ -218,7 +218,7 @@ periodics:
     containers:
     - image: quay.io/kubevirtci/golang:v20260703-61a4923
       command:
-      - "/usr/local/bin/runner.sh"
+      - "/usr/local/bin/entrypoint.sh"
       - "/bin/bash"
       - "-ce"
       - |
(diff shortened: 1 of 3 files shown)


dollierp (approved):
Oops, it's a bit shameful '-_-.

Great catch @Whitedyl!

/approve

dollierp:
Actually, it seems that none of the jobs that define `GO_MOD_PATH` use `entrypoint.sh`; they all use `runner.sh`...

It might require a follow up PR to fix them all '-_-.

Whitedyl:
> Oops, it's a bit shameful '-_-.
> 
> Great catch @Whitedyl!
> 
> /approve

Never shameful, we are humans after all!

Whitedyl:
> Actually, it seems that none of the jobs that define `GO_MOD_PATH` use `entrypoint.sh`; they all use `runner.sh`...
> 
> It might require a follow up PR to fix them all '-_-.

I didn't look into this, but I will this morning and make sure all the jobs that use it are updated.

PR #5292, opened 2026-07-16
don't trigger kubevirtci required lanes on docs-only PRs

enp0s3:
Assisted-by: Claude Opus 4.6 <noreply@anthropic.com>

Save CI resources by skipping the required lanes when the PR is a doc-change only PR.
Example: https://github.com/kubevirt/kubevirtci/pull/1765

**Release Note**
```release-note
NONE
```

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/kubevirtci/kubevirtci-presubmits.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/kubevirtci/kubevirtci-presubmits.yaml
@@ -435,7 +435,8 @@ presubmits:
           privileged: true
       nodeSelector:
         type: bare-metal-external
-  - always_run: true
+  - always_run: false
+    skip_if_only_changed: '^[^/]+\.md$'
     cluster: prow-workloads
     decorate: true
     decoration_config:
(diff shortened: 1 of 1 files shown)


enp0s3:
FYI @dhiller @dollierp

dhiller (commented):
@enp0s3 I fear it doesn't work like that - if you are switching from `always_run: true` to `skip_if_only_changed` the lanes are becoming conditional and thus are not perceived as being required by Tide. The latter is only true for `always_run: true` lanes.

See https://docs.prow.k8s.io/docs/jobs/#requiring-job-statuses for more information.

enp0s3:
> @enp0s3 I fear it doesn't work like that - if you are switching from `always_run: true` to `skip_if_only_changed` the lanes are becoming conditional and thus are not perceived as being required by Tide. The latter is only true for `always_run: true` lanes.
> 
> See https://docs.prow.k8s.io/docs/jobs/#requiring-job-statuses for more information.

@dhiller Thank you for the review! By looking at the docs I see the following bullet:
```
conditionally run jobs with required status contexts are required to have passed on a pull request to merge if the job currently matches the pull request.
```
AFAIK even when the job is conditional but with `optional:false` it is still required, and it will run when the PR matches the job.
Are you referring to the case of batch merge?

PR #5296, opened 2026-07-16
fix(external-plugins): pin images to newest tag and disable bumper

dollierp:

Two tags have been built today:
- v20260716-ff8cd14 (older)
- v20260716-a2b9efe (newer which includes github app support)

Unfortunately the bumper use the older one due to alphabetical order.

Pin to the newest tag and disable the bumper temporarly so that the change is not reverted.

/cc @dhiller @Whitedyl

## Summary by CodeRabbit

* **Chores**
  * Updated several CI service components to newer container image versions.
  * Image tags now include the build time down to the minute, improving version traceability.
* **Bug Fixes**
  * Temporarily enabled dry-run mode for periodic image bump pull requests to prevent incorrect tag ordering from triggering changes.
  * Added a follow-up note to remove this safeguard once tag sorting is corrected.

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/project-infra/project-infra-periodics.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/project-infra/project-infra-periodics.yaml
@@ -1060,7 +1060,8 @@ periodics:
       - -c
       args:
       - |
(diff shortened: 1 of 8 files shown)


dhiller (approved):
/approve

PR #5330, opened 2026-07-28
chore(kubevirtci): disable auto-bump of the QEMU runner

dollierp:

The QEMU runner should be updated very sparsely to limit the drift between the user space if the containerized QEMU and the kernel space of the nodes of the Prow workloads cluster (RHCOS 9).

/cc @dhiller @Whitedyl

## Summary by CodeRabbit

* **Chores**
  * Removed the scheduled automation that periodically created updates for Fedora provisioning image versions.
  * Added a new presubmit job to bump Fedora provisioning base image versions on pull requests targeting the main branch.
  * Other periodic and presubmit maintenance jobs continue to run as configured.

### DIFF
--- a/github/ci/prow-deploy/files/jobs/kubevirt/kubevirtci/kubevirtci-periodics.yaml
+++ b/github/ci/prow-deploy/files/jobs/kubevirt/kubevirtci/kubevirtci-periodics.yaml
@@ -72,43 +72,6 @@ periodics:
       resources:
         requests:
           memory: "8Gi"
-- name: periodic-kubevirtci-bump-fedora-provision-image
-  cron: "5 3 * * 4"
-  annotations:
-    testgrid-create-test-group: "false"
-  decoration_config:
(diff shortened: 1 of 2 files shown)


dhiller (approved):
/approve

While I agree that we should do that very seldom, I think about changing this job from periodic to always_run: false presubmit, which we can then trigger if need be.

WDYT?

/hold to be removed if you disagree

dollierp:
> While I agree that we should do that very seldom, I think about changing this job from periodic to always_run: false presubmit, which we can then trigger if need be.
> 

PR updated to move the periodic job to a presubmit job triggered manually, PTAL.

dhiller (approved):
Thank you @dollierp !

dollierp:
Switch to Fedora 43 has been merged in KubeVirt CI.

Merging this one to avoid a bump to Fedora 44.

/hold cancel

PR #5303, opened 2026-07-17
chore(jobs): consolidate GitHub credentials preset

dollierp:
**What this PR does / why we need it**:

Update `preset-github-credentials` to use the secret created during GitHub App migration of Prow core components.

It is a unified secret which contains both GitHub App credentials (keys: `appid`, `cert`) and kubevirt-bit user token (key: `token`).

This is a pre-requisite to switch some commands to use GitHub App credentials in Prow jobs.

**Special notes for your reviewer**:

/cc @dhiller @Whitedyl

## Summary by CodeRabbit

- **Bug Fixes**
  - Standardized GitHub credential handling across automated builds, releases, reporting, and maintenance tasks.
  - Improved reliability of workflows that publish reports, create pull requests, synchronize labels, and manage releases.
  - Updated selected validation jobs to run as optional checks where appropriate.

- **Chores**
  - Consolidated credential configuration and removed obsolete token setup.
  - Enhanced configuration validation to report a broader range of warnings.

- **Documentation**
  - Updated command usage examples and default credential-path documentation.

PR #5334, opened 2026-07-29
fix(jobs): drop stray quote breaking knmstate release job

qinqon:

The `release-kubernetes-nmstate` postsubmit cannot publish releases, it dies before running anything:

```
/bin/bash: -c: line 3: unexpected EOF while looking for matching `"'
```

5a548bb7a (#5303) moved the command from a quoted `command:` string into a `|` block scalar under `args:` and dropped the leading `"` but not the trailing one. In a block scalar that quote is literal shell input, hence the parse error. This PR removes it.

Failing build: https://storage.googleapis.com/kubevirt-prow/logs/release-kubernetes-nmstate/2082127492130279424/build-log.txt

`publish-kubernetes-nmstate` got the same conversion without the typo, and the release job only triggers on `^v\d+\.\d+\.\d+$` branches, so this stayed dormant until the next release attempt. I grepped the other files touched by 5a548bb7a and this was the only occurrence. Verified by extracting both jobs' `args[0]` and running `bash -n`.

/cc @dollierp


dollierp (approved):
/approve

PR #5432, opened 2026-09-03
fix(golang): entrypoint.sh reads toolchain directive for Go compiler version

Whitedyl:
The `entrypoint.sh` script reads `GO_MOD_PATH` to auto-detect the Go version from `go.mod`.
**Current behavior:**
- Reads `go 1.26.0` 
- Tries to install Go 1.26.0 via gimme
- Fails because 1.26.0 isn't cached in the image or downloadable

**Example failure:** [rehearsal-pull-project-infra-test-releng](https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_project-infra/5423/rehearsal-pull-project-infra-test-releng/2095488938968879104)

After this PR it reads the `toolchain` directive instead of the `go` directive by:
- First checks for `toolchain go1.x.y` 
- Falls back to `go 1.x.y` for older repos without toolchain
- Strips the `go` prefix from toolchain 

/cc @dhiller @dollierp 

## Summary by CodeRabbit

* **Bug Fixes**
  * Go version selection now prioritizes the toolchain version specified in the project configuration.
  * Added fallback behavior to use the configured Go version when no toolchain version is available.

### DIFF
--- a/images/golang/entrypoint.sh
+++ b/images/golang/entrypoint.sh
@@ -21,7 +21,13 @@ set -o pipefail
 GO_MOD_PATH=${GO_MOD_PATH:-}
 
 if [[ -n ${GO_MOD_PATH} && -f ${GO_MOD_PATH} ]]; then
-  export GIMME_GO_VERSION="$(grep -E '^go' "${GO_MOD_PATH}" | cut -d' ' -f 2)"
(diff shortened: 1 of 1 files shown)


dollierp (approved):
I don't think this will fix the occasional issue with gimme, which I feel is more likely related to transient network errors.

However, it's a nice optimization to download the right toolchain version from the start.

/lgtm

dhiller (commented):
/approve

Thank you @Whitedyl !

PR #5437, opened 2026-09-06
fix(golang,entrypoint): don't fail in no toolchain is set in go.mod

dollierp:

The `entrypoint.sh` script of the golang image sets `errexit` and `pipefail` options. Since the `grep` command exits with an error when there are no matches, the script fails if no `toolchain` directive is present in the `go.mod` file.

This change replaces `grep` commands by `awk` equivalent commands that do not fail if there is no match.

Fixes regression from https://github.com/kubevirt/project-infra/pull/5432.

/cc @dhiller @Whitedyl

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Go version detection from project configuration files.
  * Ensured toolchain versions are parsed consistently and validated against the expected format.
  * Improved handling of configured project paths and version values, reducing potential setup or build issues when Go versions are specified in different supported formats.

### DIFF
--- a/images/golang/entrypoint.sh
+++ b/images/golang/entrypoint.sh
@@ -20,13 +20,13 @@ set -o pipefail
 
 GO_MOD_PATH=${GO_MOD_PATH:-}
 
-if [[ -n ${GO_MOD_PATH} && -f ${GO_MOD_PATH} ]]; then
-  toolchain_version="$(grep -E '^toolchain go' "${GO_MOD_PATH}" | cut -d' ' -f2 | sed 's/^go//')"
-  if [[ -n ${toolchain_version} ]]; then
+if [[ -n "${GO_MOD_PATH}" && -f "${GO_MOD_PATH}" ]]; then
+  toolchain_version="$(awk '/^toolchain go[0-9]+(.[0-9]+){1,2}/ { sub("^go", "", $2); print($2) }' "${GO_MOD_PATH}")"
+  if [[ -n "${toolchain_version}" ]]; then
     export GIMME_GO_VERSION="${toolchain_version}"
   else
     # Fallback to go directive for repos without toolchain line
-    export GIMME_GO_VERSION="$(grep -E '^go ' "${GO_MOD_PATH}" | cut -d' ' -f2)"
+    export GIMME_GO_VERSION="$(awk '/^go [0-9]+(.[0-9]+){1,2}/ { print($2) }' "${GO_MOD_PATH}")"
   fi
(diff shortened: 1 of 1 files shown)


images/golang/entrypoint.sh
```diff
@@ -21,12 +21,12 @@ set -o pipefail
 GO_MOD_PATH=${GO_MOD_PATH:-}
 
 if [[ -n ${GO_MOD_PATH} && -f ${GO_MOD_PATH} ]]; then
-  toolchain_version="$(grep -E '^toolchain go' "${GO_MOD_PATH}" | cut -d' ' -f2 | sed 's/^go//')"
+  toolchain_version="$(awk '/^toolchain go[0-9]+(.[0-9]+){1,2}/ { sub("^go", "", $2); print($2) }' "${GO_MOD_PATH}")"
   if [[ -n ${toolchain_version} ]]; then
```
nunnatsa: @dollierp thanks for fixing this.

I don't think the grep is the issue, but the following `if`:

Running inside this container:
```
$podman run -it --rm -v$(pwd):/home/test:z --entrypoint sh quay.io/kubevirtci/gocli:2608271042-dbec3e2e
/workdir # cd /home/test/
/home/test # [[ -n "${toolchain_version}" ]]
/home/test # echo $?
1
/home/test # [[ -n ${toolchain_version} ]]
/home/test # echo $?
0
```

You can leave the grep expression as is. It's working properly.

Please fix line 23 to:
```
if [[ -n "${GO_MOD_PATH}" && -f "${GO_MOD_PATH}" ]]; then
```

and line 25 to
```
  if [[ -n "${toolchain_version}" ]]; then
```

i.e.
```
if [[ -n "${GO_MOD_PATH}" && -f "${GO_MOD_PATH}" ]]; then
  toolchain_version="$(grep -E '^toolchain go' "${GO_MOD_PATH}" | cut -d' ' -f2 | sed 's/^go//')"
  if [[ -n "${toolchain_version}" ]]; then
```

nunnatsa (approved):
/lgtm
/approve

dollierp (commented):
/approve

images/golang/entrypoint.sh
dollierp: As discussed, with `set -o errexit` and `set -o pipefail` the script fails due to grep at the toolchain_version assignement.

In bash (not sh), quoting shouldn't have any effect with `[[`, only with a single `[`.

It's still good practice, so I added the quotes.

